PR #416, opened 2022-01-18
Simplify MD5 hash generation and avoid warnings

thebluepotato:
Using CryptoKit (part of all Apple platforms) allows us to avoid multiple warnings about how MD5 is deprecated in a cryptographic context and avoids the very un-Swifty C operations related to CommonCrypto. Getting Data from the URL is also simplified.

### DIFF
--- a/Zotero/Extensions/MD5+Url.swift
+++ b/Zotero/Extensions/MD5+Url.swift
@@ -12,13 +12,31 @@ import Foundation
 import CocoaLumberjackSwift
 
 func md5(from url: URL) -> String? {
+    let bufferSize = 1024 * 1024
 
     do {
-        // Read data from file URL
-        let fileData = try Data(contentsOf: url)
+        // Open file for reading:
+        let file = try FileHandle(forReadingFrom: url)
+        defer {
+            file.closeFile()
+        }
+
+        // Create and initialize MD5 hasher:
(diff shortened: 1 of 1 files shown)


Zotero/Extensions/MD5+Url.swift
```diff
@@ -6,41 +6,19 @@
...
-                data.withUnsafeBytes {
-                    _ = CC_MD5_Update(&context, $0.baseAddress, numericCast(data.count))
-                }
-                return true // Continue
-            } else {
-                return false // End of file
-            }
-        }) { }
+        // Read data from file URL
+        let fileData = try Data(contentsOf: url)
```
michalrentka, line 18: You simplified it a bit too much. In this case you load the whole file to memory. Since we're calculating MD5 hash from files which can potentially be very big and not fit into memory, you need to use another approach here: https://developer.apple.com/documentation/cryptokit/insecure/md5/3237554-update.

thebluepotato: I did use the update method as well but as far as I understand the orignal code, the file was read into a buffer in 1 MB chunks and the hash was updated accordingly until it reached EOF. In the end we read the same amount of bytes.

If the issue is a potential blocking of the thread, we can probably dispatch an async thread or indeed do the update and finalise. What would you prefer?

michalrentka (approved):

PR #643, opened 2023-03-21
Proposed revision to the Zotero iOS AppIcon

babbage:
In spending a little time with Zotero for iOS lately I'd been struck by the possibility of evolving the iOS app icon to one that still represented the Zotero brand, but which might look more native on the iOS platform. 

The reflective metallic/glass box of the current icon is distinctive at 1024px but less so at app icon size, and has the disadvantage that it ends up surrounded by a white border in the iOS springboard and other places it appears in the interface. It seemed there could be an opportunity to present it in a more natural format for iOS by taking the reflective box full bleed and allowing the app icon itself to provide the border, rather than the current apparent hexagon frame of the three-sided box.

Screenshot of original Zotero iOS app icon, surrounded by a hexagonal box cut-out, and a revised app icon where the box edges extend to the edges of the iOS app icon

I fully appreciate it is entirely unsolicited. It may not fit with Zotero's brand direction. I have no problems of course if this is discarded. It didn't take long to put together and the only way to sensibly propose this was to show exactly what I was proposing in situ. The Z in this replacement is drawn directly from the previous icon, but the gradients, shadows and reflection are new. You're of course welcome to use this, discard it, or re-create another version with the original artwork if you like the idea but want to start over.

Apple no longer requires sizes other than a 1024px app icon, and the OS will resize as required automatically unless you wish to customise the look at smaller sizes. This pull request currently eliminates the smaller sizes and only includes a replacement 1024px version. I can easily supply the smaller sizes if Zotero wants to proceed with this change, and the smaller sizes are wanted.

Close up of original Zotero iOS app icon, surrounded by a hexagonal box cut-out, and a revised app icon where the box edges extend to the edges of the iOS app icon


fwcd:
The official icon has been [updated upstream](https://github.com/zotero/zotero/commit/1d8baa635a72d699217ed3011bac030b76e980f9):



...so it would be nice to have a matching iOS icon

dstillman:
Yes, we'll be updating the iOS app with the new official icon once Zotero 7 is out.

Thanks @babbage for the contribution here!

Issue #656, opened 2023-04-08
Remember last pen size


dstillman:
Currently reset on each PDF open

PR #661, opened 2023-04-13
Fix/pdf conditional

angiemugo:
- Added the pdf conditional so that app compiles 
- Changes system colors to color assets so that dark mode is consistent 
- Fixed the pen size resetting as in this [issue](https://github.com/zotero/zotero-ios/issues/656)

### DIFF
--- a/Zotero.xcodeproj/project.pbxproj
+++ b/Zotero.xcodeproj/project.pbxproj
@@ -5467,8 +5467,8 @@
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
 				CODE_SIGN_ENTITLEMENTS = Zotero/Zotero.entitlements;
-				CODE_SIGN_IDENTITY = "iPhone Developer";
-				CODE_SIGN_STYLE = Manual;
+				CODE_SIGN_IDENTITY = "Apple Development";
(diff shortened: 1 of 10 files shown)


Zotero.xcodeproj/project.pbxproj
```diff
@@ -5467,8 +5467,8 @@
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
 				CODE_SIGN_ENTITLEMENTS = Zotero/Zotero.entitlements;
-				CODE_SIGN_IDENTITY = "iPhone Developer";
-				CODE_SIGN_STYLE = Manual;
+				CODE_SIGN_IDENTITY = "Apple Development";
+				CODE_SIGN_STYLE = Automatic;
```
dstillman, line 5471: We can switch to automatic if that's better, but it should be in a separate commit.

Zotero/Scenes/Detail/Annotation Popover/Views/ColorPickerCell.swift
```diff
@@ -10,6 +10,7 @@ import UIKit
 
 import RxSwift
 
+#if PDFENABLED
```
dstillman, line 13: This is still in the wrong commit.

PR #672, opened 2023-04-24
fix error that was preventing changing author type

### LINKED ISSUE #651: Can't change creator type
https://forums.zotero.org/discussion/103411/cannot-change-creator-type-on-zotero-for-ipad

Appears to change but then shows the old type when going back to item details

angiemugo:
Fixes #651

### DIFF
--- a/Zotero/Controllers/Database/Requests/EditItemFromDetailDbRequest.swift
+++ b/Zotero/Controllers/Database/Requests/EditItemFromDetailDbRequest.swift
@@ -45,8 +45,6 @@ struct EditItemFromDetailDbRequest: DbRequest {
     }
 
     private func updateCreators(with data: ItemDetailState.Data, snapshot: ItemDetailState.Data, item: RItem, changes: inout RItemChanges, database: Realm) {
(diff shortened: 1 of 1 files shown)


Zotero/Controllers/Database/Requests/EditItemFromDetailDbRequest.swift
```diff
@@ -45,10 +45,8 @@ struct EditItemFromDetailDbRequest: DbRequest {
     }
 
     private func updateCreators(with data: ItemDetailState.Data, snapshot: ItemDetailState.Data, item: RItem, changes: inout RItemChanges, database: Realm) {
-        guard data.creatorIds != snapshot.creatorIds else { return }
-
+        guard data.creators != snapshot.creators else { return }
```
angiemugo, line 48: I  removed the older condition and added this one because I noticed when creators are changed we remove all creators and create new ones. This one covers all cases and so comparing the IDs may not be necessary.

michalrentka: My bad, I didn't notice you added another condition. All good.

PR #673, opened 2023-04-25
fix cancel create item manually

### LINKED ISSUE #665: Click Cancel in new item screen should skip creation of empty item
Currently it takes you back to the item details pane with an empty item, but we shouldn't create an item at all in this case

angiemugo:
Fixes #665

### DIFF
--- a/Zotero/Scenes/Detail/ItemDetail/ViewModels/ItemDetailActionHandler.swift
+++ b/Zotero/Scenes/Detail/ItemDetail/ViewModels/ItemDetailActionHandler.swift
@@ -760,12 +760,13 @@ struct ItemDetailActionHandler: ViewModelActionHandler, BackgroundDbProcessingAc
     }
 
     private func cancelChanges(in viewModel: ViewModel<ItemDetailActionHandler>) {
-        if case .duplication = viewModel.state.type  {
+        switch viewModel.state.type {
+        case .duplication, .creation:
             self.perform(request: MarkObjectsAsDeletedDbRequest<RItem>(keys: [viewModel.state.key], libraryId: viewModel.state.library.identifier)) { [weak viewModel] error in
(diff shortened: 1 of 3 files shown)


Zotero/Scenes/Detail/ItemDetail/ViewModels/ItemDetailActionHandler.swift
```diff
@@ -780,6 +780,25 @@ struct ItemDetailActionHandler: ViewModelActionHandler, BackgroundDbProcessingAc
             return
         }
 
+        if case .creation = viewModel.state.type {
```
michalrentka, line 783: No need to duplicate the whole block, you can just change the `if` to `switch`:
```
switch viewModel.state.type {
case .duplication, .creation:
... just one deletion block here ...
default: break
}
```

Zotero/Scenes/Detail/ItemDetail/ViewModels/ItemDetailActionHandler.swift
```diff
@@ -777,7 +778,8 @@ struct ItemDetailActionHandler: ViewModelActionHandler, BackgroundDbProcessingAc
                     state.hideController = true
                 }
             }
-            return
```
michalrentka, line 780: This `return` shouldn't be removed, there's no need to update the viewModel again below, since we're leaving the screen (`state.hideController = true`).

angiemugo: In that case, then I can put this piece of code in the last case of `viewModel.state.type`

michalrentka: Sure, that works too

Issue #646, opened 2023-03-24
Adjust highlight comment icon


dstillman:
https://forums.zotero.org/discussion/comment/424317/#Comment_424317

Based on the screenshots we should probably adjust the icon to more closely match the desktop in size and position.

We can also play with opacity.

nataliachodelski:
@dstillman - Hi Dan, I'm a new contribute to this project. I investigated this issue and made a PR with some changes that I think might address this issue. Could you check it out and let me know what you think?

PR #674, opened 2023-04-26
changes to address #646 - Adjust highlight comment icon

nataliachodelski:
Hi, I'm a research scientist who recently started using Zotero to collaborate with a friend on several reference collections. 
I'm also an iOS developer, and have been looking for open source projects to contribute to and came across Zotero. I've spent some time exploring the repository and forum discussions to see if I could address any open, unassigned issues.
<br>

I looked into issue #646 and wanted to propose a few small changes to improve to the size and opacity of the note annotations shown on PDFs, re the discussion on this post: https://forums.zotero.org/discussion/comment/424317.   

**I made the following changes:**
- reduced note icon size from 22 to 16 (initially tried 12, but the icons looked too small when I tested it) 
- added 50% opacity to fill color used for comment icons
- added 50% opacity to fill color used for note icon that is added highlight

<br>

**Sample screenshots of note and highlight annotations with my these changes**
        
<br>

**Sample screenshots of the same PDF before these changes (in the current App Store app)**

### DIFF
--- a/Zotero/Controllers/CommentIconDrawingController.swift
+++ b/Zotero/Controllers/CommentIconDrawingController.swift
@@ -24,7 +24,7 @@ struct CommentIconDrawingController {
         let newBoundingBox = CGRect(origin: origin, size: size)
 
         context.clip(to: newBoundingBox, mask: colorizedCgImage)
-        color.setFill()
+        color.withAlphaComponent(0.5).setFill()
         context.fill(newBoundingBox)
 
         context.resetClip()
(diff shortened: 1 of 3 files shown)


warwickmm:
I believe issue #646 is more concerned with the size of the highlight comment icon (the yellow one in the screenshots), and not the note annotation icon.

mathjiajia:
> @nataliachodelski, I believe issue #646 is more concerned with the size of the highlight comment icon (the yellow one in the screenshots) and not the note annotation icon, although perhaps those should be made consistent with the desktop app as well if they aren't already.

I think both the highlight and note icons are too huge

PR #676, opened 2023-04-26
Pin annotation bar

### LINKED ISSUE #662: Annotation toolbar position in place of navigation bar
Implement the ability to dock the annotation toolbar to top in place of navigation bar, similar to PDF Viewer app.

IMG_0361

angiemugo:
- Fixes #662 
@michalrentka , I decided to take a different route with this other than the previous branch I had.


Zotero/Scenes/Detail/PDF/Views/PDFReaderViewController.swift
```diff
@@ -487,12 +493,17 @@ class PDFReaderViewController: UIViewController {
             // Move to closest point. Use different threshold for vertical/horizontal rotation.
             let threshold: CGFloat
             if frame.height > frame.width {
-                threshold = self.isCompactWidth ? 150 : 100
+                threshold = self.isCompactWidth ? 400 : 350
```
angiemugo, line 496: I chose these numbers by testing and seeing what worked better for both the iPad screen and iPhone screen.

michalrentka: @angiemugo I don't think these numbers work better than before. Here you can see it immediately wants to push the toolbar up, which doesn't feel right:
https://www.dropbox.com/s/0fghix5v7dnc1tr/RPReplay_Final1682584450.MP4?dl=0

You can maybe make it proportional to current screen height, something like `UIScreen.main.frame.height / 4` to make the threshold the top quarter of screen.

@dstillman what do you think?

michalrentka:
@angiemugo one issue I mentioned previously in branch comments is still not resolved. When you start dragging, the navigation bar is immediately hidden while previews are still hidden too. Previews show up when position actually changes, so navigation bar should hide at the same time. Current approach has a few problems:

1. When you start dragging, navigation bar hides and previews don't show up, which seems a bit confusing:
IMG_CD44F436BAA7-1

2. When you just swipe from side to side, the navbar jumps up and down:
https://www.dropbox.com/s/6pe49b5hdznnh50/issue.mp4?dl=0

Zotero/Scenes/Detail/PDF/Views/PDFReaderViewController.swift
dstillman: Are we basing this on the frame borders or the finger position? I think it should be the latter. I wouldn't expect it to pin to the top until my finger was over the top toolbar zone.

michalrentka: @dstillman it should be center of the toolbar. Initially it was set up so that when the toolbar center is above some threshold it moves up, moving the finger seemed too much, but it can be changed easily.

dstillman: It was fine before when there were only three positions, but with the top position now, I think it's going to feel weird to have it jump to the top for no apparent reason. I think PDF Viewer is doing finger.

michalrentka: @dstillman that's why I don't like the threshold change to around 400 px from top. When it was 100 px you'd have to move the toolbar a bit higher before it wants to move up. I'd add some distance between `pinned` and `top` and keep the center point instead of finger point. PDF Viewer does finger that's true, but to me it seems a bit too far. I'm not against either though.

angiemugo:
> And one more issue, when navigation bar is hidden, it should probably not allow both `pinned` and `top` positions. PDF Viewer app does this well, try to match their behaviour: https://www.dropbox.com/s/1uxykiy1z7zcmih/RPReplay_Final1682585454.MP4?dl=0
> 
> It's looks broken when you have both `pinned` and `top` available when navbar is hidden: https://www.dropbox.com/s/wce5g7rueey7ykq/issue2.mp4?dl=0

True. I will do it like PDFKit

Zotero/Scenes/Detail/PDF/Views/PDFReaderViewController.swift
angiemugo: I revoked the changes on the threshold and instead implemented some changes to the condition for pinned/top position. While earlier I was checking where the minY of the annotation bar was to determine the position, I am now working with centerY. 
The reason I had changed the threshold was that while working with minY, by the time we met the threshold conditions, the pinned condition is what would always be met.

dstillman:
I really think this needs to be based on the touch position. It's totally unpredictable now, and it means that, for example, dropping a toolbar exactly where it started moves it to a new location:

https://user-images.githubusercontent.com/783082/235881836-1b596174-9065-4729-948d-0159eb257f03.mov

michalrentka:
> I really think this needs to be based on the touch position. It's totally unpredictable now, and it means that, for example, dropping a toolbar exactly where it started moves it to a new location:
> 
>  toolbar-positioning.mov

Even using center of toolbar shouldn't do that, so there's definitely something wrong.

dstillman:
Screenshot 2023-05-09 at 6 09 33 AM

This still seems wrong.

michalrentka:
@angiemugo it's better, but there are still some issues:

1. we talked about using the finger point instead of center of the toolbar
2. This still seems wrong: 
IMG_0380
3. Previews on sides are too low in this case:
IMG_C7B3DAF4FAE0-1
4. The top/pinned behaviour still doesn't match the PDF Viewer app, try to play with pinned/top location, show/hide the UI on tap and combinations of those, you'll see theirs makes a bit more sense.

mvasilak (changes requested):
Minor changes/questions. Looks good.

Zotero/Scenes/Detail/PDF/Views/PDFDocumentViewController.swift
```diff
@@ -352,6 +366,10 @@ final class PDFDocumentViewController: UIViewController {
         stateManager.lineWidth = lineWidth
     }
 
+    func setInterface(hidden: Bool) {
+        self.pdfController?.userInterfaceView.alpha = hidden ? 0 : 1
```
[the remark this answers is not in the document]
michalrentka: PSPDFKit uses `alpha` instead of `isHidden` internally, so if I restore hidden ui state by setting `isHidden` the UI just never shows up after that. It surprised me too.

Zotero/Scenes/Detail/PDF/Views/DashedView.swift
```diff
@@ -20,14 +47,21 @@ class DashedView: UIView {
...
-        self.dashColor = dashColor
+    init(type: Kind) {
+        self.dashColor = .black
+        self.type = type
         super.init(frame: CGRect())
-        self.addDashedBorder(color: dashColor)
+        switch type {
+        case .rounded(let cornerRadius):
+            self.layer.cornerRadius = cornerRadius
+        case .partialStraight: break
```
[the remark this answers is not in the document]
michalrentka: Fixed in 408e3c9b3c2ea5d15addca475b5dfd0f7e5fa595

Zotero/Scenes/Detail/PDF/Views/PDFReaderViewController.swift
```diff
@@ -543,126 +598,253 @@ class PDFReaderViewController: UIViewController {
...
         case .ended, .failed:
             let velocity = recognizer.velocity(in: self.view)
-            let position = self.position(fromCenter: self.annotationToolbarController.view.center, frame: self.annotationToolbarController.view.frame,
-                                         containerFrame: self.documentController.view.frame, velocity: velocity)
+            let location = recognizer.location(in: self.view)
+            let position = self.position(
+                fromTouch: location,
+                frame: self.annotationToolbarController.view.frame,
+                containerFrame: self.documentController.view.frame,
+                velocity:velocity,
```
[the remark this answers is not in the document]
michalrentka: Fixed in 408e3c9b3c2ea5d15addca475b5dfd0f7e5fa595

mvasilak (approved):

PR #679, opened 2023-05-02
Toggle collapsed when subcollection is added

### LINKED ISSUE #678: Open parent collection when subcollection is created on closed collection
In collections screen you can long press any collection and create subcollection, when subcollection is created and parent collection is closed, we should expand it so that the new collection is visible:
IMG_E72AF2B81116-1

angiemugo:
Fixes #678

### DIFF
--- a/Zotero/Controllers/Database/Requests/CreateCollectionDbRequest.swift
+++ b/Zotero/Controllers/Database/Requests/CreateCollectionDbRequest.swift
@@ -27,8 +27,11 @@ struct CreateCollectionDbRequest: DbRequest {
 
         var changes: RCollectionChanges = .name
 
-        if let key = self.parentKey {
+        if let key = self.parentKey, let parent = database.objects(RCollection.self).filter(.key(key, in: self.libraryId)).first {
             collection.parentKey = key
+            if parent.collapsed {
+                parent.collapsed = false
+            }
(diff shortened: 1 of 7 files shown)


michalrentka:
@angiemugo the parent collection should open when it's edited, not when you start editing it.

michalrentka:
@angiemugo there's quite big delay between when the collection is created and when the parent is opened:
https://github.com/zotero/zotero-ios/assets/1022956/cdb6b765-12aa-4fe7-ac81-bb9ef659b19f

That delay happens because you set the `collapsed` flag separately from collection creation. So you create a collection, results in `CollectionViewController` are updated, then you set the collapsed flag, but this doesn't cause results to update. So then a sync happens in background and objects are updated again, which causes the UI to reload with proper collapsed state.

You're overcomplicating the whole thing a bit. You can just set the `collapsed` flag in `CreateCollectionDbRequest` in here https://github.com/zotero/zotero-ios/blob/6100d56c20f010a3b04fa69a9063502e89be33c8/Zotero/Scenes/Master/Collection%20Editing/ViewModels/CollectionEditActionHandler.swift#L72

You don't even need the `shouldCollapse` flag, you can check in database whether the parent is collapsed and if it is, then change the flag to `false`.

PR #684, opened 2023-05-04
Fix save prefs

### LINKED ISSUE #660: "Save Files" pref not followed for Add by Identifier
1) Disable "Save Files"
2) Enter `arXiv:2004.13001` in Add by Identifier.

A file will be added.

angiemugo:
Fixes #660


Zotero/Scenes/Detail/Lookup/Views/LookupViewController.swift
```diff
@@ -179,11 +179,7 @@ class LookupViewController: UIViewController {
             guard let `self` = self, let value = change.newValue, value.height != self.tableViewHeight.constant else { return }
 
             self.tableViewHeight.constant = value.height
-
```
angiemugo, line 182: @michalrentka with the table view height being updated here, I don't believe this condition would be met hence why I removed it.

PR #686, opened 2023-05-09
Share image annotation

### LINKED ISSUE #677: Add Share button to image annotation popup
https://forums.zotero.org/discussion/104684/suggest-add-area-selection-and-copy-image-function-in-ios-app

angiemugo:
- Fixes #677

### DIFF
--- a/Zotero/Assets/en.lproj/Localizable.strings
+++ b/Zotero/Assets/en.lproj/Localizable.strings
@@ -221,6 +221,7 @@
 "pdf.search.one_match" = "Found 1 match";
 "pdf.search.multiple_matches" = "Found %d matches";
 "pdf.annotation_popover.title" = "Edit Annotation";
+"pdf.annotation_popover.share" = "Share Annotation";
 "pdf.annotation_popover.delete" = "Delete Annotation";
 "pdf.annotation_popover.update_subsequent_pages" = "Update subsequent pages";
 "pdf.annotation_popover.page_label_title" = "Edit Page Number";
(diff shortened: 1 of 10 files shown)


Zotero/Assets/en.lproj/Localizable.strings
```diff
@@ -221,6 +221,7 @@
 "pdf.search.one_match" = "Found 1 match";
 "pdf.search.multiple_matches" = "Found %d matches";
 "pdf.annotation_popover.title" = "Edit Annotation";
+"pdf.annotation_popover.share" = "Share Annotation";
```
dstillman, line 224: I think we'll want this to be "Share Image". Sharing an entire annotation implies sharing the comment and citation, as we do for annotation Quick Copy on the desktop, and we'll support that on iOS later (though doing it for image annotations is complicated).

michalrentka:
@angiemugo in #677 we said that it would be better to share higher resolution images. You're low-res sharing images from memory cache. Also if we render higher resolution images, I'm not sure we should keep it in memory and share from memory, I'd probably store it as a file and share from a file. But you can test how much memory those images take and whether one or another make sense. So most of the PR needs to be changed unfortunately.

angiemugo:
> @angiemugo in #677 we said that it would be better to share higher resolution images. You're low-res sharing images from memory cache. Also if we render higher resolution images, I'm not sure we should keep it in memory and share from memory, I'd probably store it as a file and share from a file. But you can test how much memory those images take and whether one or another make sense. So most of the PR needs to be changed unfortunately.

Gotcha, I'd missed that conversation. My bad!

PR #692, opened 2023-05-30
Add SwiftLint

mvasilak:
Adds a SwiftLint phase, with some proposed rules and fixes the errors found.

As there are several [rules](https://realm.github.io/SwiftLint/rule-directory.html), 
I suggest that we pick only those that we think will improve code clarity and eliminate ambiguity,
without causing a lot of fuss in the current codebase.
To that effect, I added a few rules explicitly, rather than use the default ones.
These are additionally configured to return an error, 
so that we don't end up with a bunch of ignored warnings in the end.
Most of those can be easily fixed by running `swiftlint --fix`, which we could either run manually, 
so that we are aware of the changes, or we could make part of the script.

Will keep it as a draft PR, until we decide which rules, if any, should be included.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -65,6 +65,7 @@ only_rules:
   - direct_return
   - discarded_notification_center_observer
   - empty_collection_literal
+  - empty_count
 
 anyobject_protocol: error
 
@@ -110,6 +111,10 @@ discouraged_direct_init:
   severity: error
 
 empty_collection_literal: error
+
+empty_count:
+  severity: error
+  only_after_dot: true
 
 empty_enum_arguments: error
 
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -29,6 +29,7 @@ only_rules:
   - trailing_semicolon
   - trailing_whitespace
   - unavailable_condition
+  - unneeded_break_in_switch
 
 closing_brace: error
 
@@ -102,4 +103,6 @@ trailing_whitespace:
   ignores_comments: true
 
 unavailable_condition: error
+
+unneeded_break_in_switch: error
 
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -72,6 +72,9 @@ only_rules:
   - identical_operands
   - joined_default_parameter
   - operator_usage_whitespace
+  - overridden_super_call
+  - period_spacing
+  - prefer_zero_over_explicit_init
 
 anyobject_protocol: error
 
@@ -175,6 +178,15 @@ operator_usage_whitespace:
 operator_whitespace: error
 
 orphaned_doc_comment: error
+
+overridden_super_call:
+  severity: error
+  excluded: []
+  included: ["*"]
+
+period_spacing: error
+
+prefer_zero_over_explicit_init: error
 
 private_over_fileprivate:
   severity: error
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -30,6 +30,7 @@ only_rules:
   - trailing_whitespace
   - unavailable_condition
   - unneeded_break_in_switch
+  - unused_closure_parameter
 
 closing_brace: error
 
@@ -106,3 +107,4 @@ unavailable_condition: error
 
 unneeded_break_in_switch: error
 
+unused_closure_parameter: error
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -13,7 +13,10 @@ only_rules:
   - leading_whitespace
   - mark
   - multiple_closures_with_trailing_closure
+  - no_space_in_method_call
   - opening_brace
+  - operator_whitespace
+  - orphaned_doc_comment
   - trailing_newline
   - trailing_semicolon
   - trailing_whitespace
@@ -50,9 +53,15 @@ mark: error
 
 multiple_closures_with_trailing_closure: error
 
+no_space_in_method_call: error
+
 opening_brace:
   severity: error
   allowMultilineFunc: true
+
+operator_whitespace: error
+
+orphaned_doc_comment: error
 
 trailing_newline: error
 
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -76,6 +76,11 @@ only_rules:
   - period_spacing
   - prefer_zero_over_explicit_init
   - self_binding
+  - unused_capture_list
+  - vertical_parameter_alignment_on_call
+  - vertical_whitespace_closing_braces
+  - vertical_whitespace_opening_braces
+  - yoda_condition
 
 anyobject_protocol: error
 
@@ -243,6 +248,8 @@ unavailable_condition: error
 
 unneeded_break_in_switch: error
 
+unused_capture_list: error
+
 unused_closure_parameter: error
 
 unused_enumerated: error
@@ -255,8 +262,18 @@ unused_setter_value: error
 
 vertical_parameter_alignment: error
 
+vertical_parameter_alignment_on_call: error
+
 vertical_whitespace:
   severity: error
   max_empty_lines: 1
 
+vertical_whitespace_closing_braces:
+  severity: error
+  only_enforce_before_trivial_lines: false
+
+vertical_whitespace_opening_braces: error
+
 void_return: error
+
+yoda_condition: error
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -5,8 +5,10 @@ only_rules:
   - comma
   - computed_accessors_order
   - control_statement
+  - deployment_target
   - empty_enum_arguments
   - empty_parameters
+  - empty_parentheses_with_trailing_closure
   - mark
   - opening_brace
   - trailing_newline
@@ -26,9 +28,16 @@ comma: error
 
 control_statement: error
 
+deployment_target:
+    severity: error
+    iOSApplicationExtension_deployment_target: 14.7
+    iOS_deployment_target: 14.7
+
 empty_enum_arguments: error
 
 empty_parameters: error
+
+empty_parentheses_with_trailing_closure: error
 
 mark: error
 
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -75,6 +75,7 @@ only_rules:
   - overridden_super_call
   - period_spacing
   - prefer_zero_over_explicit_init
+  - self_binding
 
 anyobject_protocol: error
 
@@ -207,6 +208,10 @@ redundant_string_enum_value: error
 redundant_void_return: error
 
 return_arrow_whitespace: error
+
+self_binding:
+  severity: error
+  bindIdentifier: self
 
 self_in_property_initialization: error
 
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -36,7 +36,11 @@ only_rules:
   - redundant_string_enum_value
   - redundant_void_return
   - return_arrow_whitespace
+  - self_in_property_initialization
+  - shorthand_operator
+  - statement_position
   - switch_case_alignment
+  - syntactic_sugar
   - trailing_comma
   - trailing_newline
   - trailing_semicolon
@@ -48,6 +52,7 @@ only_rules:
   - unused_optional_binding
   - unused_setter_value
   - vertical_parameter_alignment
+  - vertical_whitespace
   - void_return
   # opt-in rules
 
@@ -132,8 +137,18 @@ redundant_void_return: error
 
 return_arrow_whitespace: error
 
+self_in_property_initialization: error
+
+shorthand_operator: error
+
+statement_position: 
+  statement_mode: default
+  severity: error
+
 switch_case_alignment:
   severity: error
+
+syntactic_sugar: error
 
 trailing_comma:
   severity: error
@@ -163,5 +178,9 @@ unused_optional_binding:
 unused_setter_value: error
 
 vertical_parameter_alignment: error
+
+vertical_whitespace:
+  severity: error
+  max_empty_lines: 1
 
 void_return: error
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -61,6 +61,9 @@ only_rules:
   - closure_spacing
   - collection_alignment
   - contains_over_first_not_nil
+  - contains_over_range_nil_comparison
+  - direct_return
+  - discarded_notification_center_observer
 
 anyobject_protocol: error
 
@@ -89,12 +92,18 @@ comma: error
 
 contains_over_first_not_nil: error
 
+contains_over_range_nil_comparison: error
+
 control_statement: error
 
 deployment_target:
     severity: error
     iOSApplicationExtension_deployment_target: 14.7
     iOS_deployment_target: 14.7
+
+direct_return: error
+
+discarded_notification_center_observer: error
 
 discouraged_direct_init:
   severity: error
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,6 +1,7 @@
 only_rules:
   - closing_brace
   - closure_parameter_position
+  - colon
   - comma
   - computed_accessors_order
   - control_statement
@@ -15,6 +16,11 @@ only_rules:
 closing_brace: error
 
 closure_parameter_position: error
+
+colon:
+  severity: error
+  # flexible_right_spacing: true
+  # apply_to_dictionaries: false
 
 comma: error
 
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -55,12 +55,29 @@ only_rules:
   - vertical_whitespace
   - void_return
   # opt-in rules
+  - anyobject_protocol
+  - array_init
+  - closure_end_indentation
+  - closure_spacing
+  - collection_alignment
+
+anyobject_protocol: error
+
+array_init: error
 
 block_based_kvo: error
 
 closing_brace: error
 
+closure_end_indentation: error
+
 closure_parameter_position: error
+
+closure_spacing: error
+
+collection_alignment:
+  severity: error
+  align_colons: false
 
 colon:
   severity: error
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -71,6 +71,7 @@ only_rules:
   - flatmap_over_map_reduce
   - identical_operands
   - joined_default_parameter
+  - operator_usage_whitespace
 
 anyobject_protocol: error
 
@@ -164,6 +165,12 @@ ns_number_init_as_function_reference: error
 opening_brace:
   severity: error
   allowMultilineFunc: true
+
+operator_usage_whitespace:
+  severity: error
+  lines_look_around: 2
+  skip_aligned_constants: true
+  allowed_no_space_operators: ["...", "..<"]
 
 operator_whitespace: error
 
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,0 +1,29 @@
+only_rules:
+  - comma
+  - computed_accessors_order
+  - control_statement
+  - empty_enum_arguments
+  - empty_parameters
+  - mark
+  - trailing_newline
+  - trailing_semicolon
+  - trailing_whitespace
+
+comma: error
+
+control_statement: error
+
+empty_enum_arguments: error
+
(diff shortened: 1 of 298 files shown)


michalrentka:
Rules added seem fine. 

I think we could add these default rules too, since the codebase mostly adheres to them already:
- `opening_brace` and `closing_brace`
- `closure_parameter_position`
- `colon`
- `deployment_target`
- `empty_parentheses_with_trailing_closure`
- `implicit_getter`
- `leading_whitespace`
- `multiple_closures_with_trailing_closure`
- `no_space_in_method_call`
- `opening_brace`
- `operator_whitespace`
- `orphaned_doc_comment`
- `private_over_fileprivate`
- `redundant_discardable_let`
- `redundant_set_access_control`
- `redundant_string_enum_value`
- `redundant_void_return`
- `return_arrow_whitespace`
- `switch_case_alignment`
- `trailing_comma`
- `unavailable_condition`
- `unneeded_break_in_switch`
- `unused_closure_parameter`
- `unused_enumerated`
- `unused_optional_binding`
- `unused_setter_value`
- `vertical_parameter_alignment`
- `void_return`

And maybe some more for discussion. I don't have strong feelings for (not) having them, but they make sense to me:
- `block_based_kvo` I think we use the new API everywhere, no reason to use delegate-based approach
- `discouraged_direct_init` we shouldn't have these anywhere and I guess it's good to keep an eye on it
- `large_tuple`
- `legacy_cggeometry_functions`, `legacy_constant`, `legacy_constructor`, `legacy_nsgeometry_functions`, `legacy_random`
- `legacy_hashing` there may be some instances of old hashing values, but I guess we can update them while we're at it
- `line_length` do we want to set some max line length? I'm not a fan of too short lines (70-100), but I think there should be some limit. I think we could try a limit of around 200?
-  `ns_number_init_as_function_reference`
- `private_unit_test`
- `protocol_property_accessors_order`
- `redundant_optional_initialization` I think I use both options from time to time, so good time to clean this up
- `self_in_property_initialization`
- `shorthand_operator`
- `statement_position`
- `syntactic_sugar`
- `type_name` do we want any limits here? Tbh I can have some long-ish names since I tend to use post-fixes and descriptive names like `ReadAllDownloadedAndForUploadItemsDbRequest`, which is not great, but also not unreasonable imo. But maybe having a bigger upper limit could be beneficial so that if such name comes up, we can at least discuss whether it can be shortened.
- `vertical_whitespace`

And some opt-in rules for discussion. I like some more than others, but I'm listing those that make sense to me:
- `accessibility_trait_for_button` might be good to improve our accessibility
- `anyobject_protocol` I think this is Xcode warning too now
- `array_init`
- `closure_end_indentation`
- `closure_spacing`
- `collection_alignment`
- `contains_over_filter_count`
- `contains_over_filter_is_empty`
- `contains_over_first_not_nil` not sure whether this won't conflict with some realm APIs, but I can't find any instances quickly
- `contains_over_range_nil_comparison`
- `convenience_type`
- `direct_return`
- `discarded_notification_center_observer`
- `empty_collection_literal`
- `empty_count`
- `empty_string`
- `fallthrough`
- `flatmap_over_map_reduce`
- `function_default_parameter_at_end` there might be some conflicts, but I like the idea
- `identical_operands`
- `indentation_width` my Xcode is currently set up to 4 spaces, so most codebase is like that, we should probably match that
- `joined_default_parameter`
- `no_magic_numbers`
- `operator_usage_whitespace`
- `overridden_super_call`
- `period_spacing`
- `prefer_zero_over_explicit_init`
- `self_binding` I currently use `guard let 'self' = self else { return }` mostly out of habit, but this might be a good time to switch to probably clearer `guard let self = self else { return }`
- `sorted_enum_cases` I like the idea, do you have any example where this might be counter intuitive?
- `sorted_imports` I usually separate imports into native/ios imports and 3rd party imports and sort those 2 groups separately, but we could just sort them as a whole
- `superfluous_else`
- `type_contents_order` do we want to have a specified order? I adhere to some rules out of habit, but there will most likely be conflicts
- `unowned_variable_capture` I like the rule, but might be slightly annoying when you actually want to use `unowned`. Though you can just disable it for that given case, so I'm leaning towards adding this?
- `unused_capture_list`
- `vertical_parameter_alignment_on_call`
- `vertical_whitespace_between_cases`
- `vertical_whitespace_closing_braces`
- `vertical_whitespace_opening_braces`
- `weak_delegate`
- `yoda_condition`

If you feel like something is too restrictive or you'd like to add some more, I'm open for suggestions.

I agree with having them all return errors. You can always ignore them in file if you want to break some rule for whatever reason.

We should definitely keep `swiftlint --fix` a manual option. We should probably add `swiftlint lint` to build phases so that we are aware of issues while building and testing our code. If I remember correctly there was also some Xcode extension which would show linting errors directly in code. So we can enable it individually. 
Ideally when CI works we could have a message/badge on PRs to see whether there are any missed errors.

mvasilak:
I've gone through the default rules (opt-in rules will be added in a next pass). 
Here are some notable cases that may warrant further discussion:
* `unneeded_break_in_switch`: Please validate changes in this commit https://github.com/zotero/zotero-ios/pull/692/commits/a5a7fd288b04286467f44d2e9857479f9292b35d
* `discouraged_direct_init`: Please validate changes in this commit https://github.com/zotero/zotero-ios/pull/692/commits/ce70695e2d32c1f66e2edf9e1251c7b9d9d3756e
* `large_tuple`: Skipped, as it would require significant changes where there are tuples passed with even 5 parameters or more. We could relax the rule parameters, but if we want to adhere to it, it might be better to tackle it properly in a separate PR.
* `line_length`: Skipped for further discussion. I agree that we shouldn't use a very limiting size, however even 200 characters faces many cases, especially of long function definitions or calls. I'm personally fun of the code style suggestion that breaks such lines by moving each parameter defined/called, in each own line, and the closing paranthesis also in each own line, to balance vertically with the function definition/call start. Glad to discuss any other suggestion.
* `type_name`: Skipped for further discussion. I agree that we shouldn't restrict type names too much, especially if it makes sense to have large names, so we can be descriptive and convery meaning. We could go over each case individually, to see if there is sensible way to shorten them.

michalrentka:
> unneeded_break_in_switch: Please validate changes in this commit https://github.com/zotero/zotero-ios/commit/a5a7fd288b04286467f44d2e9857479f9292b35d

Looks good

> discouraged_direct_init: Please validate changes in this commit https://github.com/zotero/zotero-ios/commit/ce70695e2d32c1f66e2edf9e1251c7b9d9d3756e

Left comment in https://github.com/zotero/zotero-ios/commit/ce70695e2d32c1f66e2edf9e1251c7b9d9d3756e#r117040369

> large_tuple: Skipped, as it would require significant changes where there are tuples passed with even 5 parameters or more. We could relax the rule parameters, but if we want to adhere to it, it might be better to tackle it properly in a separate PR.

Agreed, I forgot there are so big tuples used

> line_length: Skipped for further discussion. I agree that we shouldn't use a very limiting size, however even 200 characters faces many cases, especially of long function definitions or calls. I'm personally fun of the code style suggestion that breaks such lines by moving each parameter defined/called, in each own line, and the closing paranthesis also in each own line, to balance vertically with the function definition/call start. Glad to discuss any other suggestion.

The only thing I don't like about it is that it is very horizontally limiting. One function call takes a lot of lines even if it has a few parameters and there's a lot of empty horizontal space. I usually just add a new line before argument that surpasses some line length and align the new line to first argument on previous line. So that it uses the most of horizontal space before taking more vertical space. Example of what I use would be:
```
func veryLongFunctionName(parameter1: String, parameter2: Int,
                          parameter3: Bool, parameter4: String) {
self.otherVeryLongFunction(withParametert1: argument1, parameter2: argument2,
                           parameter3: argument3, parameter4: argument4)
self.smallFunction(withParameter1: argument1, parameter2: argument2)
...
}
```

Compared to your style:
```
func veryLongFunctionName(
    argument1: String, 
    argument2: Int,
    argument3: Bool, 
    argument4: String
) {
self.otherVeryLongFunction(
    withParametert1: argument1, 
    parameter2: argument2,
    parameter3: argument3, 
    parameter4: argument4
)
self.smallSmallFunction(
    withParameter1: argument1, 
    parameter2: argument2
)
...
}
```

With your style I like that you can quickly see all parameters of functions because they are all aligned with the same offset. But it also looks like I wanted to fit the code to a phone screen size. Do you maybe have a bigger chunk of code written in your style which I could check out? Maybe if I see more of it it'll look better and make more sense on bigger screen too.

Another thing to consider is real-life limitations. Do you for example work on a small 13 inch macbook so that you'd welcome smaller line length? I have a 27" monitor and 15" macbook, so I'd welcome longer lines. But if you have more limited screen space, it'll be easier for me to adjust.

> type_name: Skipped for further discussion. I agree that we shouldn't restrict type names too much, especially if it makes sense to have large names, so we can be descriptive and convery meaning. We could go over each case individually, to see if there is sensible way to shorten them.

Ok we can skip this for now and just deal with it over time if needed.

mvasilak:
> > line_length: Skipped for further discussion. I agree that we shouldn't use a very limiting size, however even 200 characters faces many cases, especially of long function definitions or calls. I'm personally fan of the code style suggestion that breaks such lines by moving each parameter defined/called, in its own line, and the closing paranthesis also in its own line, to balance vertically with the function definition/call start. Glad to discuss any other suggestion.
> 
> The only thing I don't like about it is that it is very horizontally limiting. One function call takes a lot of lines even if it has a few parameters and there's a lot of empty horizontal space. I usually just add a new line before argument that surpasses some line length and align the new line to first argument on previous line. So that it uses the most of horizontal space before taking more vertical space. Example of what I use would be:
> 
> ```
> func veryLongFunctionName(parameter1: String, parameter2: Int,
>                           parameter3: Bool, parameter4: String) {
> self.otherVeryLongFunction(withParametert1: argument1, parameter2: argument2,
>                            parameter3: argument3, parameter4: argument4)
> self.smallFunction(withParameter1: argument1, parameter2: argument2)
> ...
> }
> ```
> 
> Compared to your style:
> 
> ```
> func veryLongFunctionName(
>     argument1: String, 
>     argument2: Int,
>     argument3: Bool, 
>     argument4: String
> ) {
> self.otherVeryLongFunction(
>     withParametert1: argument1, 
>     parameter2: argument2,
>     parameter3: argument3, 
>     parameter4: argument4
> )
> self.smallSmallFunction(
>     withParameter1: argument1, 
>     parameter2: argument2
> )
> ...
> }
> ```
> 
> With your style I like that you can quickly see all parameters of functions because they are all aligned with the same offset. But it also looks like I wanted to fit the code to a phone screen size. Do you maybe have a bigger chunk of code written in your style which I could check out? Maybe if I see more of it it'll look better and make more sense on bigger screen too.
> 
> Another thing to consider is real-life limitations. Do you for example work on a small 13" macbook so that you'd welcome smaller line length? I have a 27" monitor and 15" macbook, so I'd welcome longer lines. But if you have more limited screen space, it'll be easier for me to adjust.

I do work in a 14" macbook, but I am ok with large lines. The goal of the code style is to have both code legibility and a consistent way of writing, so we should be able to come up with one that accomplishes that, without being to restrictive.

The argument for one parameter per line is also that it's easier to edit, and keep respecting consistently the linting rules.
However, if I may suggest a style in-between those two:

```
self.smallSmallFunction(withParameter1: argument1,  parameter2: argument2) // line limited not exceeded, keep call/definition in one line

func veryLongFunctionName(
    parameter1: String, parameter2: Int, parameter3: Bool, // exceeds line limit, continue in a new line
    parameter4: String // final parameter, continue in new line, to have all parameters nested in the function parenthesis
) {
    // code
}
```
Small functions that don't exceed the line limit we pick, can be left as one liners.
Larger ones, will wrap around parameter in groups, but keep the closing parenthesis in a new line, to balance it with the function definition/call start, as it is easier (I think) to visually understand where the call/definition ends.

michalrentka:
> > > line_length: Skipped for further discussion. I agree that we shouldn't use a very limiting size, however even 200 characters faces many cases, especially of long function definitions or calls. I'm personally fan of the code style suggestion that breaks such lines by moving each parameter defined/called, in its own line, and the closing paranthesis also in its own line, to balance vertically with the function definition/call start. Glad to discuss any other suggestion.
> > 
> > 
> > The only thing I don't like about it is that it is very horizontally limiting. One function call takes a lot of lines even if it has a few parameters and there's a lot of empty horizontal space. I usually just add a new line before argument that surpasses some line length and align the new line to first argument on previous line. So that it uses the most of horizontal space before taking more vertical space. Example of what I use would be:
> > ```
> > func veryLongFunctionName(parameter1: String, parameter2: Int,
> >                           parameter3: Bool, parameter4: String) {
> > self.otherVeryLongFunction(withParametert1: argument1, parameter2: argument2,
> >                            parameter3: argument3, parameter4: argument4)
> > self.smallFunction(withParameter1: argument1, parameter2: argument2)
> > ...
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Compared to your style:
> > ```
> > func veryLongFunctionName(
> >     argument1: String, 
> >     argument2: Int,
> >     argument3: Bool, 
> >     argument4: String
> > ) {
> > self.otherVeryLongFunction(
> >     withParametert1: argument1, 
> >     parameter2: argument2,
> >     parameter3: argument3, 
> >     parameter4: argument4
> > )
> > self.smallSmallFunction(
> >     withParameter1: argument1, 
> >     parameter2: argument2
> > )
> > ...
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > With your style I like that you can quickly see all parameters of functions because they are all aligned with the same offset. But it also looks like I wanted to fit the code to a phone screen size. Do you maybe have a bigger chunk of code written in your style which I could check out? Maybe if I see more of it it'll look better and make more sense on bigger screen too.
> > Another thing to consider is real-life limitations. Do you for example work on a small 13" macbook so that you'd welcome smaller line length? I have a 27" monitor and 15" macbook, so I'd welcome longer lines. But if you have more limited screen space, it'll be easier for me to adjust.
> 
> I do work in a 14" macbook, but I am ok with large lines. The goal of the code style is to have both code legibility and a consistent way of writing, so we should be able to come up with one that accomplishes that, without being to restrictive.
> 
> The argument for one parameter per line is also that it's easier to edit, and keep respecting consistently the linting rules. However, if I may suggest a style in-between those two:
> 
> ```
> self.smallSmallFunction(withParameter1: argument1,  parameter2: argument2) // line limited not exceeded, keep call/definition in one line
> 
> func veryLongFunctionName(
>     parameter1: String, parameter2: Int, parameter3: Bool, // exceeds line limit, continue in a new line
>     parameter4: String // final parameter, continue in new line, to have all parameters nested in the function parenthesis
> ) {
>     // code
> }
> ```
> 
> Small functions that don't exceed the line limit we pick, can be left as one liners. Larger ones, will wrap around parameter in groups, but keep the closing parenthesis in a new line, to balance it with the function definition/call start, as it is easier (I think) to visually understand where the call/definition ends.

Ok that sounds reasonable. Now the question is what line length to choose? What is a comfortable line length for you that you can see normally when working? I'd suggest somewhere in range 150-200.

Then do you want to convert the whole codebase in one go? Can linter do this automatically with good results or do you have to do this manually? If manual, do we maybe want to edit gradually?

mvasilak:
The default `line_length.error: 200` produces `143` errors. Increasing to `250`, still produces `136` errors, so I'd say the defaults are fine. Furthermore configuring the linter to ignore this rule for URLs, function declarations, comments, interpolated strings, will result in `106` errors.

Since for this rule there is no autocorrection, I suggest that we group all rules requiring a manual approach (I'm going over the opt-in rules now), in a separate issue, and we tackle them gradually. 

For the style we discussed above, we can start following it now, and we gradually fix existing code if we do some work upon it. No need to go over everything at once.

mvasilak:
Here are comments for the opt-in rules:

- `accessibility_trait_for_button`, `convenience_type`, `no_magic_numbers`, `weak_delegate`: Skipped for future issue. Require significant changes, will be part of a separate issue, to be resolved later. Along default rules `line_length`, `large_tuple`, `type_name`
- `contains_over_filter_count`, `contains_over_filter_is_empty`: Skipped for future issue. They seem to trigger also `filter` calls that use a predicate, which is not exactly the same case.
- `function_default_parameter_at_end`: Skipped for future issue. Not only it requires significant changes, but we should consider if changing the parameter order might affect the call's "meaning". An alternative for those cases might be to add default values to existing parameters in the end, not having one.
-`indentation_width`: Skipped for future issue. Requires significant changes.  I do agree on using the default 4 spaces indent, and as a first step we could use in all files `Editor > Structure > Re-Indent`, before enforcing this rule.
- `self_binding`: Applied. We could even use the newer syntax `guard let self else { return }` whenever the same name is used for the local scope.
- `sorted_enum_cases`: Skipped for further discussion. If the cases are presented in a semantic order, that doesn’t correspond to lexicographic order, then it might not be preferred. E.g. `enum sizes { case small, medium, large }`
- `sorted_imports`: Skipped for further discussion.  Maybe imports should also move to the top of the file, at least if not in an compile directive block.
- `superfluous_else`: Skipped for further discussion. Seems to be triggered by wrong examples? Also, the result might not be more clear to understand.
- `type_contents_order`: Skipped for further discussion. I wouldn’t mind a specific order, I find it useful to know where to look in any class, especially coupled with `MARK` comments so you can see this in file navigation/minimap.
- `unowned_variable_capture`: Skipped for further discussion. Why not use `weak` in these cases, combined with a guard? The result isn’t the same most of the times?
- `vertical_whitespace_between_cases`: Skipped for further discussion. Not sure about this one. I find that case nesting is enough to achieve code legibility.

michalrentka:
> * `self_binding`: Applied. We could even use the newer syntax `guard let self else { return }` whenever the same name is used for the local scope.

Yes, I started using that from now on too.

> * `sorted_enum_cases`: Skipped for further discussion. If the cases are presented in a semantic order, that doesn’t correspond to lexicographic order, then it might not be preferred. E.g. `enum sizes { case small, medium, large }`

I guess those special cases can be just ignored by swiftlint, but we can skip this completely. We can sort manually if it makes sense.

> * `superfluous_else`: Skipped for further discussion. Seems to be triggered by wrong examples? Also, the result might not be more clear to understand.

Ok let's skip this and we'll hopefully take care of it manually.

> * `unowned_variable_capture`: Skipped for further discussion. Why not use `weak` in these cases, combined with a guard? The result isn’t the same most of the times?

Not sure I understand you here. This rule states that you should prefer `weak` (+ `guard` if needed) instead of `unowned`, so that's what you're suggesting?

> * `vertical_whitespace_between_cases`: Skipped for further discussion. Not sure about this one. I find that case nesting is enough to achieve code legibility.

I think I mostly already add white space here, but we can alter to one or another style. But I think it would be good to have coherent style here, with or without white space.

mvasilak:
> > * `unowned_variable_capture`: Skipped for further discussion. Why not use `weak` in these cases, combined with a guard? The result isn’t the same most of the times?
> 
> Not sure I understand you here. This rule states that you should prefer `weak` (+ `guard` if needed) instead of `unowned`, so that's what you're suggesting?

I'm referring to your comment on this:
> I like the rule, but might be slightly annoying when you actually want to use unowned. Though you can just disable it for that given case, so I'm leaning towards adding this?

I was wondering why not use always `weak` for captured variables.

> 
> > * `vertical_whitespace_between_cases`: Skipped for further discussion. Not sure about this one. I find that case nesting is enough to achieve code legibility.
> 
> I think I mostly already add white space here, but we can alter to one or another style. But I think it would be good to have coherent style here, with or without white space.

The indentation in the case block is indeed the default, and that's fine. This rule adds an empty line between cases, that's why it seems a bit too much to me.

mvasilak:
> Got it, I understand now. I sometimes use `unowned` when I'm 100% sure the object won't be deallocated. But we can opt to always use `weak` and `guard`.

I'll add the rule it in the separate issue, along with the other rules to be applied later.

> Yes I understand. I just remember that I used to add an empty line between bigger `case`s to separate them better. But I agree that we can skip this and just remove those extra lines when we see them.

Since it's already under partial use, let's add this rule then. It supports autocorrection, so it's easy to enforce it.

PR #696, opened 2023-06-07
Fix file sync issue when background upload finishes but doesn't update state in main app

michalrentka:
Background uploads can sometime pass and file is successfully uploaded to backend, but main app is not notified and local state is not updated, so when user opens main app it tries to upload the same file during sync again and receives 412 `If-None-Match: * set but file exists`. In this case we re-download the remote file version to be sure both are the same and mark file as uploaded.

### DIFF
--- a/Zotero/Controllers/Sync/SyncActions/RevertLibraryFilesSyncAction.swift
+++ b/Zotero/Controllers/Sync/SyncActions/RevertLibraryFilesSyncAction.swift
@@ -95,6 +95,7 @@ struct RevertLibraryFilesSyncAction: SyncAction {
                 DDLogWarn("RevertLibraryFilesSyncAction: can't rename file - \(error)")
                 // If it can't be moved, at least delete the old one. It'll have to be re-downloaded anyway.
                 try? self.fileStorage.remove(oldFile)
+                try? self.fileStorage.remove(newFile)
             }
         }
     }
--- a/Zotero/Controllers/Sync/SyncActions/RevertLibraryFilesSyncAction.swift
+++ b/Zotero/Controllers/Sync/SyncActions/RevertLibraryFilesSyncAction.swift
@@ -1,0 +1,101 @@
+//
+//  RevertLibraryFilesSyncAction.swift
+//  Zotero
+//
+//  Created by Michal Rentka on 13.06.2023.
+//  Copyright © 2023 Corporation for Digital Scholarship. All rights reserved.
+//
+
+import Foundation
+
+import CocoaLumberjackSwift
+import RxSwift
+
+struct RevertLibraryFilesSyncAction: SyncAction {
+    typealias Result = [String]
+
+    let libraryId: LibraryIdentifier
+
+    unowned let dbStorage: DbStorage
+    unowned let fileStorage: FileStorage
+    unowned let schemaController: SchemaController
+    unowned let dateParser: DateParser
+    let queue: DispatchQueue
+
+    var result: Single<[String]> {
+        return Single.create { subscriber -> Disposable in
+            DDLogInfo("RevertLibraryFilesSyncAction: revert files to upload")
+
+            do {
+                let toUpload = try self.dbStorage.perform(request: ReadAllAttachmentUploadsDbRequest(libraryId: self.libraryId), on: self.queue)
+                var cachedResponses: [ItemResponse] = []
+                var failedKeys: [String] = []
+
+                for item in toUpload {
+                    do {
+                        let file = Files.jsonCacheFile(for: .item, libraryId: self.libraryId, key: item.key)
+                        let data = try self.fileStorage.read(file)
+                        let jsonObject = try JSONSerialization.jsonObject(with: data, options: .allowFragments)
+
+                        if let jsonData = jsonObject as? [String: Any] {
+                            let response = try ItemResponse(response: jsonData, schemaController: self.schemaController)
+                            cachedResponses.append(response)
+                        } else {
+                            failedKeys.append(item.key)
+                        }
+                    } catch let error {
+                        DDLogError("RevertLibraryFilesSyncAction: can't load cached file - \(error)")
+                        failedKeys.append(item.key)
+                    }
+                }
+
+                DDLogInfo("RevertLibraryFilesSyncAction: loaded \(cachedResponses.count) cached items, missing \(failedKeys.count)")
+                
+                DDLogInfo("RevertLibraryFilesSyncAction: delete files which were not uploaded yet")
+                for key in failedKeys {
+                    let file = Files.attachmentDirectory(in: self.libraryId, key: key)
+                    try? self.fileStorage.remove(file)
+                }
+
+                var changedFilenames: [StoreItemsResponse.FilenameChange] = []
+                try self.dbStorage.perform(on: self.queue) { coordinator in
+                    // Delete items that didn't sync before and weren't submitted to backend
+                    DDLogError("RevertLibraryFilesSyncAction: delete failed keys")
+                    try coordinator.perform(request: DeleteObjectsDbRequest<RItem>(keys: failedKeys, libraryId: self.libraryId))
+                    // Store cached objects from backend to local database to get rid of local changes.
+                    DDLogError("RevertLibraryFilesSyncAction: restore cached objects")
+                    let request = StoreItemsDbResponseRequest(responses: cachedResponses, schemaController: self.schemaController, dateParser: self.dateParser, preferResponseData: true)
+                    changedFilenames = try coordinator.perform(request: request).changedFilenames
+                    coordinator.invalidate()
+                }
+
+                DDLogError("RevertLibraryFilesSyncAction: rename local files to match file names")
+                self.renameExistingFiles(changes: changedFilenames, libraryId: self.libraryId)
+
+                subscriber(.success(failedKeys))
+            } catch let error {
+                subscriber(.failure(error))
+            }
+
+            return Disposables.create()
+        }
+    }
+
+    private func renameExistingFiles(changes: [StoreItemsResponse.FilenameChange], libraryId: LibraryIdentifier) {
+        for change in changes {
+            let oldFile = Files.attachmentFile(in: libraryId, key: change.key, filename: change.oldName, contentType: change.contentType)
+
+            guard self.fileStorage.has(oldFile) else { continue }
+
+            let newFile = Files.attachmentFile(in: libraryId, key: change.key, filename: change.newName, contentType: change.contentType)
+
+            do {
+                try self.fileStorage.move(from: oldFile, to: newFile)
+            } catch let error {
+                DDLogWarn("RevertLibraryFilesSyncAction: can't rename file - \(error)")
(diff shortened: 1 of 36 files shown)


Zotero/Controllers/Sync/SyncController.swift
```diff
@@ -1747,12 +1771,23 @@ final class SyncController: SynchronizationController {
...
             self.enqueue(actions: actions, at: 0, delay: delay)
 
-        case .libraryConflict:
+        case .libraryConflict(let response):
             guard self.conflictRetries < self.conflictDelays.count else {
                 self.abort(error: .cantResolveConflict(data: .from(syncObject: object, keys: keys, libraryId: libraryId)))
                 return true
             }
 
+            if response == "If-None-Match: * set but file exists" {
```
dstillman, line 1780: Don't want to test against this — regular English strings (i.e., not JSON with a specific structure) in the response body are just for debugging and aren't guaranteed to be stable. The relevant conditions here would be 1) a `412` for a file upload authorization request and 2) `Last-Modified-Version` of the response > the local attachment version. And then there are some other situations that the desktop app handles if (2) isn't true:

https://github.com/zotero/zotero/blob/6106e379c96a26897ead32a300190e2327a11f3e/chrome/content/zotero/xpcom/storage/zfs.js#L552-L620

Also doesn't seem right that this is under something called `libraryConflict`. This would specifically be for a file upload authorization request. You can see the other 4xx response codes handled by the desktop for a file upload authorization request in the function linked above. I would expect this to be handled in the same place.

michalrentka: @dstillman I'm not sure I understand fully. In case of 412 on authorization it first checks response header version, compares to local item version and if remote version is newer, it re-downloads the file and updates item. 

But if remote `version <= item.version` then it fetches the same item from backend and checks versions again? Wouldn't `Last-Modified-Version` from authorization response be the same as `version` of item fetched directly from backend?

Then if remote `version <= item.version` it checks md5 hashes of remote and local files, if they match it just updates the local file as synced. If they don't match, it marks a file conflict, which we probably don't support on iOS yet. So I guess I'll just abort sync in that case?

Also for 413 I guess I'll just show an error with usage and quota values?

dstillman: Yeah, sorry, that's really confusing ("downloading" is referring to the attachment item, not the file), possibly evolved incorrectly over the years, and probably isn't relevant in any case as long as there's no local file editing.

It looks like there are three possible causes of 412 for a file upload authorization request currently:

1) If-Match being set to a hash but the file not existing
2) The ETag not matching the current version of the file
3) If-None-Match: * being set but the file existing

(1) won't ever happen from iOS, I don't think.

(2) and (3) can probably be simplified to force-downloading the file and marking as uploaded, yes, at least as long as we don't have local file editing.

michalrentka:
@dstillman I updated `SyncController` to handle upload authorization failures separately as requested: https://github.com/zotero/zotero-ios/blob/dfa41cb6de595a5808a9a90bb4aa7142203d915e/Zotero/Controllers/Sync/SyncController.swift#L1646-L1700. 

And I implemented the logic to revert file uploads in case of 403 on file authorization: https://github.com/zotero/zotero-ios/blob/dfa41cb6de595a5808a9a90bb4aa7142203d915e/Zotero/Controllers/Sync/SyncActions/RevertLibraryFilesSyncAction.swift.

There's also a bit of refactoring and simplification of `SyncScheduler` and cleanups due to us adding SwiftLint to code base.

Zotero/Controllers/Sync/SyncActions/UploadFixSyncAction.swift
```diff
@@ -0,0 +1,168 @@
...
+                            subscriber(.failure(Error.attachmentMissingRemotely))
+                            return Disposables.create()
+
+                        case .local, .localAndChangedRemotely, .remote: break
+                        }
+
+                        // Create new attachment model with updated location so that `AttachmentDownloader` doesn't ignore it
+                        let newAttachment = Attachment(type: .file(filename: filename, contentType: contentType, location: .remote, linkType: linkType),
+                                                       title: attachment.title, key: attachment.key, libraryId: attachment.libraryId)
+                        subscriber(.success(newAttachment))
```
mvasilak, line 158: This seems a bit more clear, since switch is exhaustive.

```suggestion

                        case .local, .localAndChangedRemotely, .remote:
                            // Create new attachment model with updated location so that `AttachmentDownloader` doesn't ignore it
                            let newAttachment = Attachment(type: .file(filename: filename, contentType: contentType, location: .remote, linkType: linkType),
                                                           title: attachment.title, key: attachment.key, libraryId: attachment.libraryId)
                            subscriber(.success(newAttachment))
                        }
```

Zotero/Controllers/Sync/SyncActions/UploadFixSyncAction.swift
```diff
@@ -0,0 +1,168 @@
...
+
+                switch update.kind {
+                case .failed(let error):
+                    self.finishDownload?(.failure(error))
+                    self.finishDownload = nil
+
+                case .ready:
+                    self.finishDownload?(.success(()))
+                    self.finishDownload = nil
+                case .progress, .cancelled: break
```
mvasilak, line 68:
```suggestion
                    self.finishDownload = nil
                    
                case .progress, .cancelled: 
                    break
```

Zotero/Controllers/Sync/SyncScheduler.swift
```diff
@@ -8,180 +8,219 @@
...
 }
 
 protocol WebSocketScheduler: AnyObject {
     func webSocketUpdate(libraryId: LibraryIdentifier)
 }
 
-private typealias SchedulerAction = (syncType: SyncController.SyncType, librarySyncType: SyncController.LibrarySyncType)
-
+/// Controller that schedules synchronisation of local and remote data. All syncs are requested through this controller and it decides if/when next sync should be started.
+/// If a sync fails this controller can also retry it with to resolve issues.
```
mvasilak, line 28: Something like:
```suggestion
/// If a sync fails this controller can also retry it with increasing intervals to resolve issues.
```

Zotero/Controllers/Sync/SyncScheduler.swift
```diff
@@ -8,180 +8,219 @@
...
-        }
+    /// Requests a new sync of given type with specified libraries.
+    /// - parameter type: Sync type
+    /// - parameter libraries: Libraries to sync
+    func request(sync type: SyncController.Kind, libraries: SyncController.Libraries) {
+        DDLogInfo("SyncScheduler: requested \(type) sync for \(libraries)")
+        self.enqueueAndStart(sync: Sync(type: type, libraries: libraries))
     }
 
+    /// Requests a sync based on websocked reported changes in given library.
```
mvasilak, line 114:
```suggestion
    /// Requests a sync based on websocket reported changes in given library.
```

Zotero/Controllers/Sync/SyncScheduler.swift
```diff
@@ -8,180 +8,219 @@
...
-        guard let (nextSyncType, nextLibrarySyncType) = self.nextAction else {
-            self.nextAction = action
+    /// Start next sync in queue. In case of normal/external syncs wait at least `syncTimeout` between syncs. In case of retry syncs wait for specified delay based on current retry attempt count.
+    private func startNextSync() {
+        // Ignore this if there is a sync in progress. Otherwise pick next sync in queue.
+        guard self.syncInProgress == nil, let nextSync = self.syncQueue.first else {
+            if self.syncInProgress == nil && self.syncQueue.isEmpty {
+                // Report finished queue
+                self.inProgress.accept(false)
+            }
```
mvasilak, line 177: This seems a bit more clear:
```suggestion
        // Ignore this if there is a sync in progress.
        guard self.syncInProgress == nil else { return }
        // Otherwise pick next sync in queue.
        guard let nextSync = self.syncQueue.first else {
            // Report finished queue
            self.inProgress.accept(false)
```

Zotero/Controllers/Sync/SyncActions/RevertLibraryFilesSyncAction.swift
```diff
@@ -0,0 +1,101 @@
...
+            guard self.fileStorage.has(oldFile) else { continue }
+
+            let newFile = Files.attachmentFile(in: libraryId, key: change.key, filename: change.newName, contentType: change.contentType)
+
+            do {
+                try self.fileStorage.move(from: oldFile, to: newFile)
+            } catch let error {
+                DDLogWarn("RevertLibraryFilesSyncAction: can't rename file - \(error)")
+                // If it can't be moved, at least delete the old one. It'll have to be re-downloaded anyway.
+                try? self.fileStorage.remove(oldFile)
```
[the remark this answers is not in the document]
michalrentka: I mean, there's always a chance :). If something broke previously during sync and the name didn't change properly... So removing the new file too won't hurt.

dstillman:
Do we have tests for these things?

michalrentka:
> Do we have tests for these things?

They are in line after it's approved. I fixed existing tests in `testfixes` branch. I'll add new test for this file reverting mechanism (do we have any on desktop app?) and then we can gradually add more as needed. I'll create a new ticket for me and @mvasilak with a (mostly) prioritized list of tests so that we can increase our test coverage together.

Zotero/Controllers/Sync/SyncActions/RevertLibraryFilesSyncAction.swift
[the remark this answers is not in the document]
michalrentka: Check 49628051723bedbfb92ec2c57f4cc9f7282ec56b

Zotero/Controllers/Sync/SyncActions/UploadFixSyncAction.swift
michalrentka: Check abfcf1ab5443c69ed6e054242a368773a2bdecc0

Zotero/Controllers/Sync/SyncScheduler.swift
michalrentka: Check abfcf1ab5443c69ed6e054242a368773a2bdecc0

Zotero/Controllers/Sync/SyncScheduler.swift
michalrentka: Check abfcf1ab5443c69ed6e054242a368773a2bdecc0

Zotero/Controllers/Sync/SyncScheduler.swift
michalrentka: Check abfcf1ab5443c69ed6e054242a368773a2bdecc0

Zotero/Controllers/Sync/SyncActions/UploadFixSyncAction.swift
michalrentka: Check abfcf1ab5443c69ed6e054242a368773a2bdecc0

mvasilak (approved):

PR #697, opened 2023-06-07
Merge zotero schemes

mvasilak:
* Merges schemes `Zotero` & `Zotero + PSPDFKIT` to a common `Zotero` scheme, that has PDF enabled throughout.
* Removes unnecessary conditional compilation directives
* Updates README


README.md
```diff
@@ -5,5 +5,6 @@
 1. `git clone --recursive https://github.com/zotero/zotero-ios.git`
 1. `cd zotero-ios`
 1. `./scripts/setup.sh`
-1. If you have your own PSPDFKit license, add it to `pdf-keys/license.plist` and `pdf-keys/pspdfkit-pod.txt`.
+1. If you have your own PSPDFKit license, add it to `licenses/licenses.plist` and `licenses/license_id.txt`.
```
michalrentka, line 8: `licenses/license_id.txt` should not exist for other users, that's just in our own repo for our use. The mention of `pdf-keys/pspdfkit-pod.txt` was from times when we still used CocoaPods, I forgot to update that myself 👎. So remove the ` and licenses/license_id.txt` part.

michalrentka (changes requested):

Issue #683, opened 2023-05-03
Identifier lookup download aborted when popup is closed


angiemugo:
When downloading an item with an attachment, a pop up shows up that indicates the progress of the download. When this pop is closed, the download is also canceled and never completes.

dstillman:
There's a Save Files toggle in the app settings that controls that.

PR #707, opened 2023-06-22
Handle lookup item & attachment creation in IdentifierLookupController

mvasilak:
closes https://github.com/zotero/zotero-ios/issues/683

Moves item & attachment creation from identifier lookup to a separate user controller, that is guarranted to fulfill ongoing lookups, if lookup window is closed.

With this change, perhaps we should label the button as `Close`, instead of `Cancel`. Furthermore, the items added for this identifier, perhaps should indicate that they might have pending downloads.

-- 

TODO:
- [x] Keep track of separate lookup window initializations, if selected library & collections are different.
- [x] Include downloaded attachments from translated identifiers, in items toolbar label `Downloaded x/y`.
- [x] Show progress of identifier lookup in items toolbar label as `Saved x/y`.
- [x] Add button in items toolbar, that will restore lookup window, when there are identifier lookups in the background.
- [x] Add `Cancel all` button in lookup window, that will cancel all downloads and trash items/attachments already fetched.
- [x] Rename current `Cancel` button in lookup window to `Close`, that dismisses window and continues lookup in the background.
- [ ] Add a cancel button in each lookup item row, for individual cancellation. Cancelling an item, will also remove its attachments. Use an X or trash icon.
- [ ] Add a cancel button in each attachment row, that cancels/removes this attachment.
- [x] Cleanup of finished lookup items from controller, when lookup window is closed, and all operations are finished/cancelled.
- [x] When new identifiers are added for lookup, with previous ones queued, show a list of all items.

### DIFF
--- a/Zotero/Controllers/Attachment Downloader/RemoteAttachmentDownloader.swift
+++ b/Zotero/Controllers/Attachment Downloader/RemoteAttachmentDownloader.swift
@@ -114,6 +114,17 @@ final class RemoteAttachmentDownloader {
             self.operationQueue.addOperations(operations, waitUntilFinished: false)
         }
     }
+    
+    private func resetBatchDataIfNeeded() {
+        guard operations.isEmpty else { return }
+        totalBatchCount = 0
+        batchProgress = nil
+    }
+
+    func stop() {
+        DDLogInfo("RemoteAttachmentDownloader: stop")
+        self.operationQueue.cancelAllOperations()
+    }
 
     private func createDownload(url: URL, attachment: Attachment, parentKey: String) -> (Download, RemoteAttachmentDownloadOperation)? {
         let download = Download(key: attachment.key, parentKey: parentKey, libraryId: attachment.libraryId)
@@ -170,10 +181,7 @@ final class RemoteAttachmentDownloader {
     private func finish(download: Download, file: File, attachment: Attachment, parentKey: String, result: Result<(), Swift.Error>) {
         self.operations[download] = nil
         progressObservers[download] = nil
-        if operations.isEmpty {
-            totalBatchCount = 0
-            batchProgress = nil
-        }
+        resetBatchDataIfNeeded()
 
         switch result {
         case .success:
@@ -184,7 +192,7 @@ final class RemoteAttachmentDownloader {
         case .failure(let error):
             DDLogError("RemoteAttachmentDownloader: failed to download attachment \(download.key), \(download.libraryId) - \(error)")
 
-            let isCancelError = (error as? AttachmentDownloadOperation.Error) == .cancelled
+            let isCancelError = (error as? RemoteAttachmentDownloadOperation.Error) == .cancelled
             self.errors[download] = isCancelError ? nil : error
 
             if isCancelError {
--- a/Zotero/Controllers/Attachment Downloader/RemoteAttachmentDownloader.swift
+++ b/Zotero/Controllers/Attachment Downloader/RemoteAttachmentDownloader.swift
@@ -104,14 +104,59 @@ final class RemoteAttachmentDownloader {
 
             DDLogInfo("RemoteAttachmentDownloader: enqueue \(data.count) attachments")
 
-            let downloadAndOperations = data.compactMap({ self.createDownload(url: $0.1, attachment: $0.0, parentKey: $0.2) })
+            let downloadAndOperations = data.compactMap({ createDownload(url: $0.1, attachment: $0.0, parentKey: $0.2) })
             let downloads = downloadAndOperations.map({ $0.0 })
             let operations = downloadAndOperations.map({ $0.1 })
             downloads.forEach {
                 // Send first update to immediately reflect new state
                 self.observable.on(.next(Update(download: $0, kind: .progress(0))))
             }
-            self.operationQueue.addOperations(operations, waitUntilFinished: false)
+            operationQueue.addOperations(operations, waitUntilFinished: false)
+        }
+        
+        func createDownload(url: URL, attachment: Attachment, parentKey: String) -> (Download, RemoteAttachmentDownloadOperation)? {
+            let download = Download(key: attachment.key, parentKey: parentKey, libraryId: attachment.libraryId)
+
+            guard operations[download] == nil, let file = file(for: attachment) else { return nil }
+
+            let progress = Progress(totalUnitCount: 100)
+            let operation = RemoteAttachmentDownloadOperation(url: url, file: file, progress: progress, apiClient: apiClient, fileStorage: fileStorage, queue: processingQueue)
+            operation.finishedDownload = { [weak self] result in
+                self?.processingQueue.async(flags: .barrier) {
+                    guard let self else { return }
+                    self.finish(download: download, file: file, attachment: attachment, parentKey: parentKey, result: result)
+                }
+            }
+            operation.progressHandler = { [weak self] progress in
+                self?.processingQueue.async(flags: .barrier) {
+                    guard let self else { return }
+                    self.observe(progress: progress, attachment: attachment, download: download)
+                }
+            }
+
+            operations[download] = operation
+            totalBatchCount += 1
+            
+            if let batchProgress {
+                batchProgress.addChild(progress, withPendingUnitCount: 100)
+                batchProgress.totalUnitCount += 100
+            } else {
+                let batchProgress = Progress(totalUnitCount: 100)
+                batchProgress.addChild(progress, withPendingUnitCount: 100)
+                self.batchProgress = batchProgress
+            }
+
+            return (download, operation)
+            
+            func file(for attachment: Attachment) -> File? {
+                switch attachment.type {
+                case .file(let filename, let contentType, _, _):
+                    return Files.attachmentFile(in: attachment.libraryId, key: attachment.key, filename: filename, contentType: contentType)
+
+                case .url:
+                    return nil
+                }
+            }
         }
     }
     
@@ -126,56 +171,11 @@ final class RemoteAttachmentDownloader {
         self.operationQueue.cancelAllOperations()
     }
 
-    private func createDownload(url: URL, attachment: Attachment, parentKey: String) -> (Download, RemoteAttachmentDownloadOperation)? {
-        let download = Download(key: attachment.key, parentKey: parentKey, libraryId: attachment.libraryId)
-
-        guard self.operations[download] == nil, let file = self.file(for: attachment) else { return nil }
-
-        let progress = Progress(totalUnitCount: 100)
-        let operation = RemoteAttachmentDownloadOperation(url: url, file: file, progress: progress, apiClient: self.apiClient, fileStorage: self.fileStorage, queue: self.processingQueue)
-        operation.finishedDownload = { [weak self] result in
-            self?.processingQueue.async(flags: .barrier) {
-                guard let self else { return }
-                self.finish(download: download, file: file, attachment: attachment, parentKey: parentKey, result: result)
-            }
-        }
-        operation.progressHandler = { [weak self] progress in
-            self?.processingQueue.async(flags: .barrier) {
-                guard let self else { return }
-                self.observe(progress: progress, attachment: attachment, download: download)
-            }
-        }
-
-        self.operations[download] = operation
-        totalBatchCount += 1
-        
-        if let batchProgress {
-            batchProgress.addChild(progress, withPendingUnitCount: 100)
-            batchProgress.totalUnitCount += 100
-        } else {
-            let batchProgress = Progress(totalUnitCount: 100)
-            batchProgress.addChild(progress, withPendingUnitCount: 100)
-            self.batchProgress = batchProgress
-        }
-
-        return (download, operation)
-    }
-
     private func observe(progress: Progress, attachment: Attachment, download: Download) {
         let observer = progress.observe(\.fractionCompleted) { [weak self] progress, _ in
             self?.observable.on(.next(Update(download: download, kind: .progress(CGFloat(progress.fractionCompleted)))))
         }
         self.progressObservers[download] = observer
-    }
-
-    private func file(for attachment: Attachment) -> File? {
-        switch attachment.type {
-        case .file(let filename, let contentType, _, _):
-            return Files.attachmentFile(in: attachment.libraryId, key: attachment.key, filename: filename, contentType: contentType)
-
-        case .url:
-            return nil
-        }
     }
 
     private func finish(download: Download, file: File, attachment: Attachment, parentKey: String, result: Result<(), Swift.Error>) {
--- a/Zotero/Controllers/Attachment Downloader/RemoteAttachmentDownloader.swift
+++ b/Zotero/Controllers/Attachment Downloader/RemoteAttachmentDownloader.swift
@@ -30,7 +30,8 @@ final class RemoteAttachmentDownloader {
         let kind: Kind
     }
 
-    private let queue: DispatchQueue
+    private let accessQueue: DispatchQueue
+    private let processingQueue: DispatchQueue
     private let operationQueue: OperationQueue
     private let disposeBag: DisposeBag
     private unowned let apiClient: ApiClient
@@ -41,18 +42,36 @@ final class RemoteAttachmentDownloader {
     private var operations: [Download: RemoteAttachmentDownloadOperation]
     private var errors: [Download: Swift.Error]
     private var progressObservers: [Download: NSKeyValueObservation]
+    private var batchProgress: Progress?
+    private var totalBatchCount: Int = 0
+    
+    var batchData: (Progress?, Int, Int) {
+        var progress: Progress?
+        var totalBatchCount = 0
+        var remainingBatchCount = 0
+
+        self.accessQueue.sync { [weak self] in
+            guard let self else { return }
+            progress = self.batchProgress
+            remainingBatchCount = self.operations.count
+            totalBatchCount = self.totalBatchCount
+        }
+
+        return (progress, remainingBatchCount, totalBatchCount)
+    }
 
     init(apiClient: ApiClient, fileStorage: FileStorage) {
-        let queue = DispatchQueue(label: "org.zotero.RemoteAttachmentDownloader.ProcessingQueue", qos: .userInteractive, attributes: .concurrent)
+        let processingQueue = DispatchQueue(label: "org.zotero.RemoteAttachmentDownloader.ProcessingQueue", qos: .userInteractive, attributes: .concurrent)
         let operationQueue = OperationQueue()
         operationQueue.maxConcurrentOperationCount = 2
         operationQueue.qualityOfService = .userInitiated
-        operationQueue.underlyingQueue = queue
+        operationQueue.underlyingQueue = processingQueue
 
         self.apiClient = apiClient
         self.fileStorage = fileStorage
         self.operationQueue = operationQueue
-        self.queue = queue
+        self.accessQueue = DispatchQueue(label: "org.zotero.RemoteAttachmentDownloader.AccessQueue", qos: .userInteractive, attributes: .concurrent)
+        self.processingQueue = processingQueue
         self.observable = PublishSubject()
         self.operations = [:]
         self.progressObservers = [:]
@@ -65,7 +84,7 @@ final class RemoteAttachmentDownloader {
         var progress: CGFloat?
         var error: Swift.Error?
 
-        self.queue.sync { [weak self] in
+        self.accessQueue.sync { [weak self] in
             if let operation = self?.operations[download] {
                 if let _progress = operation.progress {
                     progress = CGFloat(_progress.fractionCompleted)
@@ -80,8 +99,8 @@ final class RemoteAttachmentDownloader {
     }
 
     func download(data: [(Attachment, URL, String)]) {
-        self.queue.async(flags: .barrier) { [weak self] in
-            guard let self = self else { return }
+        self.accessQueue.async(flags: .barrier) { [weak self] in
+            guard let self else { return }
 
             DDLogInfo("RemoteAttachmentDownloader: enqueue \(data.count) attachments")
 
@@ -95,21 +114,32 @@ final class RemoteAttachmentDownloader {
 
         guard self.operations[download] == nil, let file = self.file(for: attachment) else { return nil }
 
-        let operation = RemoteAttachmentDownloadOperation(url: url, file: file, apiClient: self.apiClient, fileStorage: self.fileStorage, queue: self.queue)
+        let progress = Progress(totalUnitCount: 100)
+        let operation = RemoteAttachmentDownloadOperation(url: url, file: file, progress: progress, apiClient: self.apiClient, fileStorage: self.fileStorage, queue: self.processingQueue)
         operation.finishedDownload = { [weak self] result in
-            self?.queue.async(flags: .barrier) {
-                guard let self = self else { return }
+            self?.processingQueue.async(flags: .barrier) {
+                guard let self else { return }
                 self.finish(download: download, file: file, attachment: attachment, parentKey: parentKey, result: result)
             }
         }
         operation.progressHandler = { [weak self] progress in
-            self?.queue.async(flags: .barrier) {
-                guard let self = self else { return }
+            self?.processingQueue.async(flags: .barrier) {
+                guard let self else { return }
                 self.observe(progress: progress, attachment: attachment, download: download)
             }
         }
 
         self.operations[download] = operation
+        totalBatchCount += 1
+        
+        if let batchProgress {
+            batchProgress.addChild(progress, withPendingUnitCount: 100)
+            batchProgress.totalUnitCount += 100
+        } else {
+            let batchProgress = Progress(totalUnitCount: 100)
+            batchProgress.addChild(progress, withPendingUnitCount: 100)
+            self.batchProgress = batchProgress
+        }
 
         return operation
     }
@@ -133,6 +163,11 @@ final class RemoteAttachmentDownloader {
 
     private func finish(download: Download, file: File, attachment: Attachment, parentKey: String, result: Result<(), Swift.Error>) {
         self.operations[download] = nil
+        progressObservers[download] = nil
+        if operations.isEmpty {
+            totalBatchCount = 0
+            batchProgress = nil
+        }
 
         switch result {
         case .success:
--- a/Zotero/Controllers/Attachment Downloader/RemoteAttachmentDownloader.swift
+++ b/Zotero/Controllers/Attachment Downloader/RemoteAttachmentDownloader.swift
@@ -104,12 +104,18 @@ final class RemoteAttachmentDownloader {
 
             DDLogInfo("RemoteAttachmentDownloader: enqueue \(data.count) attachments")
 
-            let operations = data.compactMap({ self.createDownload(url: $0.1, attachment: $0.0, parentKey: $0.2) })
+            let downloadAndOperations = data.compactMap({ self.createDownload(url: $0.1, attachment: $0.0, parentKey: $0.2) })
+            let downloads = downloadAndOperations.map({ $0.0 })
+            let operations = downloadAndOperations.map({ $0.1 })
+            downloads.forEach {
+                // Send first update to immediately reflect new state
+                self.observable.on(.next(Update(download: $0, kind: .progress(0))))
+            }
             self.operationQueue.addOperations(operations, waitUntilFinished: false)
         }
     }
 
-    private func createDownload(url: URL, attachment: Attachment, parentKey: String) -> RemoteAttachmentDownloadOperation? {
+    private func createDownload(url: URL, attachment: Attachment, parentKey: String) -> (Download, RemoteAttachmentDownloadOperation)? {
         let download = Download(key: attachment.key, parentKey: parentKey, libraryId: attachment.libraryId)
 
         guard self.operations[download] == nil, let file = self.file(for: attachment) else { return nil }
@@ -141,7 +147,7 @@ final class RemoteAttachmentDownloader {
             self.batchProgress = batchProgress
         }
 
-        return operation
+        return (download, operation)
     }
 
     private func observe(progress: Progress, attachment: Attachment, download: Download) {
--- a/Zotero/Controllers/Attachment Downloader/RemoteAttachmentDownloader.swift
+++ b/Zotero/Controllers/Attachment Downloader/RemoteAttachmentDownloader.swift
@@ -122,13 +122,13 @@ final class RemoteAttachmentDownloader {
             let progress = Progress(totalUnitCount: 100)
             let operation = RemoteAttachmentDownloadOperation(url: url, file: file, progress: progress, apiClient: apiClient, fileStorage: fileStorage, queue: processingQueue)
             operation.finishedDownload = { [weak self] result in
-                self?.processingQueue.async(flags: .barrier) {
+                self?.accessQueue.async(flags: .barrier) {
                     guard let self else { return }
                     self.finish(download: download, file: file, attachment: attachment, parentKey: parentKey, result: result)
                 }
             }
             operation.progressHandler = { [weak self] progress in
-                self?.processingQueue.async(flags: .barrier) {
+                self?.accessQueue.async(flags: .barrier) {
                     guard let self else { return }
                     self.observe(progress: progress, attachment: attachment, download: download)
                 }
--- a/Zotero/Controllers/Attachment Downloader/RemoteAttachmentDownloader.swift
+++ b/Zotero/Controllers/Attachment Downloader/RemoteAttachmentDownloader.swift
@@ -123,14 +123,12 @@ final class RemoteAttachmentDownloader {
             let operation = RemoteAttachmentDownloadOperation(url: url, file: file, progress: progress, apiClient: apiClient, fileStorage: fileStorage, queue: processingQueue)
             operation.finishedDownload = { [weak self] result in
                 self?.accessQueue.async(flags: .barrier) {
-                    guard let self else { return }
-                    self.finish(download: download, file: file, attachment: attachment, parentKey: parentKey, result: result)
+                    finish(download: download, file: file, attachment: attachment, parentKey: parentKey, result: result)
                 }
             }
             operation.progressHandler = { [weak self] progress in
                 self?.accessQueue.async(flags: .barrier) {
-                    guard let self else { return }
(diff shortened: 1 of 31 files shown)


michalrentka (changes requested):

Zotero/Scenes/Detail/Lookup/LookupCoordinator.swift
```diff
@@ -50,8 +50,12 @@ final class LookupCoordinator: NSObject, Coordinator {
     private func lookupController(multiLookupEnabled: Bool, hasDarkBackground: Bool, userControllers: UserControllers) -> LookupViewController {
         let collectionKeys = Defaults.shared.selectedCollectionId.key.flatMap({ Set([$0]) }) ?? []
         let state = LookupState(multiLookupEnabled: multiLookupEnabled, hasDarkBackground: hasDarkBackground, collectionKeys: collectionKeys, libraryId: Defaults.shared.selectedLibrary)
-        let handler = LookupActionHandler(dbStorage: userControllers.dbStorage, fileStorage: self.controllers.fileStorage, translatorsController: self.controllers.translatorsAndStylesController,
-                                          schemaController: self.controllers.schemaController, dateParser: self.controllers.dateParser, remoteFileDownloader: userControllers.remoteFileDownloader)
+        let handler = LookupActionHandler(
+            identifierLookupController: controllers.userControllers!.identifierLookupController,
```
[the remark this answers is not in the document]
mvasilak: `userControllers` are actually passed as a non-optional parameter, somehow I missed that 😅
Check https://github.com/zotero/zotero-ios/pull/707/commits/eaeb31a0dac23c575ed08c779687364e6ac1219d

michalrentka:
Code-wise it's fine, it moved existing functionality to background so that things can finish even if the UI is closed. 

> With this change, perhaps we should label the button as Close, instead of Cancel

@mvasilak I think that `Cancel` currently still cancels the translation? The web view is still inside the popup, so if the popup closes, translation is cancelled, right? I'd keep it as it is - you can cancel the translation if something goes wrong, but if items are translated successfully, you can close the popup so that you don't have to watch items download one by one.

> Furthermore, the items added for this identifier, perhaps should indicate that they might have pending downloads.

@dstillman a little explanation and question: currently translation happens, item(s) are created for the main (parent) items, then files are downloaded and if individual downloads succeed then their attachment items are created. Which means if you're on slow internet you'll first see parent item(s) without attachments and only when those files are downloaded attachments start showing up.

As @mvasilak hinted, would it be better to create all items immediately and then start all downloads (and show their progress the same way as if downloading them from our server)? This approach would be more complicated just because you'd have to handle interruptions and retry downloads (if users just close the app, downloads are interrupted and will have to be restarted after the app starts again). Also if people don't really download that many big papers at once, you most likely won't even notice the difference, because smaller papers are downloaded quickly and they just show up after few seconds. So I'm not sure whether it's worth it.

dstillman:
Is the question about showing pending downloads in the lookup progress dialog or in the items list itself?

For the former, we know immediately from the translator whether there's an attachment, and we know the name to use. That's how we show the attachment line right away in the Zotero Connector save popup (and maybe the share extension?). So at least for that, we can show attachment progress without creating the attachment item first. Maybe we're already doing that?

If the question is about the items list, and we're talking about switching to a Close button once items are created but before downloads are done, could we hook into the lookup controller for the items list progress indicator, and also queue an 'open' action so that it opens the file once it exists? Seems very complicated, but I don't think we want to create an attachment item first, because that seems very likely to lead to attachment items with missing files, attachment items being synced before files are available, etc.

For a simpler option, could we just add an indicator/button somewhere that would reopen the lookup dialog to show progress? We do something similar for multi-item PDF metadata retrieval in the desktop app.

michalrentka:
> Is the question about showing pending downloads in the lookup progress dialog or in the items list itself?

The former is already implemented, it was the latter. I agree it's probably unnecessarily overcomplicated and we can create attachments only after they download and just ignore them if they don't.

> For a simpler option, could we just add an indicator/button somewhere that would reopen the lookup dialog to show progress? We do something similar for multi-item PDF metadata retrieval in the desktop app.

That's a good idea. We could have a progress bar on bottom, same as when we're downloading multiple files (screenshot below) with a button to re-open the lookup popup. That way you'd know you're currently downloading and can view state as well.

IMG_139D1C7CA62F-1

It could say "Lookup x/y" with progress bar below and "Open" to the right.

dstillman:
> We could have a progress bar on bottom, same as when we're downloading multiple files (screenshot below) with a button to re-open the lookup popup. That way you'd know you're currently downloading and can view state as well.

👍

dstillman:
"Lookup" is an internal term for the feature. Can it not just say "Downloaded x/y"?

michalrentka:
> "Lookup" is an internal term for the feature. Can it not just say "Downloaded x/y"?

Sure, I thought we might want to distinguish attachment downloads from lookup downloads. But I guess we can even combine them if both are in progress then.

michalrentka:
@dstillman there's one more usability issue probably. When you add multiple identifiers, downloads are now in background, but translation happens sequentially. Translation could be moved to background too (by moving the underlying webview to parent controller). Do we want to do that? Or is it fine like this?

dstillman:
What would the usability issue be? That translation stops if you close the app?

michalrentka:
> What would the usability issue be? That translation stops if you close the app?

Currently yes. The popup says "Cancel" while translations are in progress, so when you close the popup currently running translations are stopped. Started downloads finish in background.

dstillman:
That seems fine.

dstillman:
Ah, I see. So the question is if we want to change it to process them in the background with a Close button instead of (or in addition to) Cancel? And we added something to reopen the progress dialog, right? Does that button stick around after completion so you can view results?

dstillman:
> I guess we could add cancel to individual translations instead of general cancel and we'd keep Close for the whole window.

Why? Because some network request might be hanging? Maybe we want that, but seems weird not to also give the option to cancel the whole process if we have any cancel options.

> `translated z/w`

"Saved z/w"

michalrentka:
@mvasilak got it.

Some other issues:
- When single translation/download is in progress, the download progress in toolbar doesn't show up. 
https://github.com/zotero/zotero-ios/assets/1022956/41938fbd-dd2c-4518-b700-c0fe15b36fa9

- When I start translation on bad internet (I turned on network limiter), item and attachment are translated and I can see their titles, but then download progress is stuck at 0 and I don't see "downloading" in the toolbar. I see it only after it downloads a bit, do you know what it's waiting for?  
https://github.com/zotero/zotero-ios/assets/1022956/e8effc3b-35a2-4e83-b310-f63f65f95b98

michalrentka:
> > I guess we could add cancel to individual translations instead of general cancel and we'd keep Close for the whole window.
> 
> Why? Because some network request might be hanging? Maybe we want that, but seems weird not to also give the option to cancel the whole process if we have any cancel options.

I thought cancel would make sense when the translation finishes and you see the resulted item is not what you wanted. But we can add cancel (all) and individual cancels or just cancel option.

mvasilak:
@michalrentka I'd notice that as well. Probably need an initial event w/ 0 progress, similar to what `AttachmentDownloader` does. I will investigate it.

dstillman:
> I thought cancel would make sense when the translation finishes and you see the resulted item is not what you wanted.

The latter could be useful, but it should probably be an (x) delete button. And I guess that'd be fine for cancelling while in progress too.

mvasilak:
New specs, from the discussion, please verify:
* Downloaded attachments from translated identifiers, are included in items toolbar label `Downloaded x/y`.
* New label `Saved x/y` will be added also in items toolbar, to show translation progress in the background.
* Button in items toolbar, will allow to re-open queue of translated identifiers.
* Lookup window will have `Cancel all` button, that will cancel all downloads and remove items/attachments already fetched. Probably it's good to show a confirmation alert.
* Lookup window will also have `Close` button, that just continues the process in the background.
* Each lookup item row will have a cancel button, for individual cancellation. Cancelling an item, will also remove its attachments. If item already downloaded, in a sense it becomes a delete/remove button, but could be the same, e.g. an X or trash icon.
* Each attachment row will also have a cancel button, just to cancel/remove this attachment.

A few things not commented already:
* Looking up new identifiers, could open the lookup input window, but when lookup is started, add them in the list with existing lookup items, if any.
* Lookup items should be removed at some point from the list. I guess when all their attachments have finished (successfully or otherwise), and if the lookup list window is not visible at the time.

michalrentka:
> @michalrentka regarding the sole download not appearing, it also happens for normal downloads of attachments, due to this guard
> 
> https://github.com/zotero/zotero-ios/blob/f9bfbf88afe8b2630d135fbe04cabb1287543782/Zotero/Scenes/Detail/Items/Models/ItemsState.swift#L35
> 
> What is the reasoning behind this? Should this edge case be maintained? If so I'll make this an option for the constructor. Or could we remove this check for all cases?

Yeah it made sense when it was used just for downloads. When you tap a single attachment which needs to download, you already see the download progress in the icon, so I ignored it. I guess just remove the `guard`.

michalrentka:
> @michalrentka the "backgrounding" of the identifiers lookup has been implemented, and while the PR is not complete yet, I wouldn't mind some first comments.

I like the general idea, I think it's going in the right direction. There are some mostly minor things I'd comment on a finished PR, but I don't think it's worth doing actual code review until it's finished? Or do you want to fully review the `IdentifierLookupController` for example?

> * Uses web view from protocol implemented by LibrariesViewController (currently).

This is quite clean in the end so it's fine in there.

> * Uses nested function style for those private functions it makes sense, so please also provide some feedback on this.

I can't say I hate them, it's certainly better in Xcode, but it's quite hard to read when doing code review. You can't just fold them here, so when you have something like:
```
func first(...) {
    func second(...) {
        ...
    }
    func third(...) {
        ...
    }

    actual body of first
}
```
it's easy to loose track of what's what and where the actual initial (`first` in this case) function starts. Is it possible to define them on bottom instead of on top? I think that would be easier to navigate here on github at least.

mvasilak:
> I like the general idea, I think it's going in the right direction. There are some mostly minor things I'd comment on a finished PR, but I don't think it's worth doing actual code review until it's finished? Or do you want to fully review the `IdentifierLookupController` for example?

If they are only minor 😅, better review in total when it's finished.

> it's easy to loose track of what's what and where the actual initial (`first` in this case) function starts. Is it possible to define them on bottom instead of on top? I think that would be easier to navigate here on github at least.

That's a good suggestion, check https://github.com/zotero/zotero-ios/pull/707/commits/2d56f4964abcb1bbaa985189165c3684ff6e9674, to see if it makes it easier to read in github.

michalrentka:
> > I like the general idea, I think it's going in the right direction. There are some mostly minor things I'd comment on a finished PR, but I don't think it's worth doing actual code review until it's finished? Or do you want to fully review the `IdentifierLookupController` for example?
> 
> If they are only minor 😅, better review in total when it's finished.

Yes I'd leave that for now, nothing worth a bigger rewrite :).

> > it's easy to loose track of what's what and where the actual initial (`first` in this case) function starts. Is it possible to define them on bottom instead of on top? I think that would be easier to navigate here on github at least.
> 
> That's a good suggestion, check [2d56f49](https://github.com/zotero/zotero-ios/commit/2d56f4964abcb1bbaa985189165c3684ff6e9674), to see if it makes it easier to read in github.

Yup I like this more. If you're ok with it we can stick with that.

mvasilak:
@dstillman @michalrentka only the individual cancellation actions are not yet implemented. As this is getting quite big and complex, it might be could to review the functionality & code so far, and even consider moving the final tasks to a separate PR. 

Some notes:
* Looking up identifiers, while there are ongoing lookups that have been dismissed before, will show a list of all lookups, with the new ones added on top.
* Cancelling all lookups will stop unfinished operations, and trash items/attachments already stored.
* Cancelling all lookups will restore the lookup window to the prior textview input.
* To restore ongoing lookups, the progress view in the items toolbar has an indicative UI and uses a tap gesture recognizer.

Zotero/Controllers/IdentifierLookupController.swift
```diff
@@ -0,0 +1,445 @@
...
+            failedCount = self.lookupFailedCount
+            totalCount = self.lookupTotalCount
+            data = self.lookupData
+        }
+    }
+
+    // MARK: Setups
+    private func setupObservers() {
+        remoteFileDownloader.observable
+            .subscribe { update in
```
michalrentka, line 208: you can create scheduler with `backgroundQueue` and just call `observeOn(.backgroundScheduler)` instead of calling `backgroundQueue.async` down in `finish(download:attachment:)`.

Zotero/Controllers/IdentifierLookupController.swift
```diff
@@ -0,0 +1,445 @@
...
+                return
+            }
+            self.setupObserver(for: lookupWebViewHandler)
+            lookupData = self.lookupData
+        }
+    }
+    
+    func lookUp(libraryId: LibraryIdentifier, collectionKeys: Set<String>, identifier: String) {
+        let lookupSettings = LookupWebViewHandler.LookupSettings(libraryIdentifier: libraryId, collectionKeys: collectionKeys)
+        lookupWebViewHandlersByLookupSettings[lookupSettings]?.lookUp(identifier: identifier)
```
michalrentka, line 147: Does this always exist? Maybe add a `guard` and a `DDLogError` at least in case something breaks in future and users report that this does nothing.

Zotero/Scenes/Master/Libraries/Views/LibrariesViewController.swift
```diff
@@ -207,3 +212,18 @@ extension LibrariesViewController: UITableViewDataSource, UITableViewDelegate {
...
+extension LibrariesViewController: IdentifierLookupWebViewProvider {
+    func addWebView() -> WKWebView {
+        let webView = WKWebView()
+        webView.isHidden = true
+        view.insertSubview(webView, at: 0)
+        return webView
+    }
+    
+    func removeWebView(_ webView: WKWebView) {
+        if view.subviews.contains(where: { $0 == webView }) {
```
michalrentka, line 225: Do you need to check this? Does this even need to be called here? Can't you just call `webView.removeFromSuperview()` directly?

Zotero/Controllers/IdentifierLookupController.swift
```diff
@@ -0,0 +1,445 @@
...
+    
+    func cancelAllLookups() {
+        accessQueue.async(flags: .barrier) { [weak self] in
+            guard let self else { return }
+            DDLogInfo("IdentifierLookupController: cancel all lookups")
+            let keys = self.lookupWebViewHandlersByLookupSettings.keys
+            for key in keys {
+                if let webView = self.lookupWebViewHandlersByLookupSettings.removeValue(forKey: key)?.webViewHandler.webView {
+                    inMainThread {
+                        self.webViewProvider?.removeWebView(webView)
```
michalrentka, line 158: Can't we call `webView.removeFromSuperview()` directly?

Zotero/Controllers/IdentifierLookupController.swift
```diff
@@ -0,0 +1,445 @@
...
+        lookupWebViewHandlersByLookupSettings[lookupSettings]?.lookUp(identifier: identifier)
+    }
+    
+    func cancelAllLookups() {
+        accessQueue.async(flags: .barrier) { [weak self] in
+            guard let self else { return }
+            DDLogInfo("IdentifierLookupController: cancel all lookups")
+            let keys = self.lookupWebViewHandlersByLookupSettings.keys
+            for key in keys {
+                if let webView = self.lookupWebViewHandlersByLookupSettings.removeValue(forKey: key)?.webViewHandler.webView {
```
michalrentka, line 156: I'd change this to `guard`, at least we'll get rid of some depth below.

Zotero/Controllers/IdentifierLookupController.swift
```diff
@@ -0,0 +1,445 @@
...
+            let storedItemResponses: [(ItemResponse, LibraryIdentifier)] = lookupData.compactMap {
+                switch $0.state {
+                case .translated(let translatedLookupData):
+                    return (translatedLookupData.response, translatedLookupData.libraryId)
+                    
+                default:
+                    return nil
+                }
+            }
+            for (response, libraryId) in storedItemResponses {
```
michalrentka, line 174: Call `self.backgroundQueue.async` before the `for` loop, unnecessary overhead. You can also create all dbrequests in the `for` loop (or `map`) and then call single `dbStorage.perform` with all requests for single db write.

Zotero/Controllers/IdentifierLookupController.swift
```diff
@@ -0,0 +1,445 @@
...
+                    } catch let error {
+                        DDLogError("IdentifierLookupController: can't trash item(s) - \(error)")
+                    }
+                }
+            }
+        }
+    }
+    
+    func getIdentifiersLookupCount(callback: @escaping (Int, Int, Int, [LookupData]) -> Void) {
+        accessQueue.async { [weak self] in
```
michalrentka, line 189: Does this need to be asynchronous? If yes, this looks overly complicated, I'd just do
```
accessQueue.async { [weak self] in
    callback(self.lookupSavedCount, self.lookupFailedCount, self.lookupTotalCount, self.lookupData)
}
```

Ideally it would be synchronous and then you'd need all those variables.

Zotero/Controllers/IdentifierLookupController.swift
```diff
@@ -0,0 +1,445 @@
...
+                        remoteFileDownloader.download(data: downloadData)
+                        observable.on(.next(Update(kind: .pendingAttachments(identifier: identifier, response: response, attachments: attachments))))
+                    }
+                }
+            }
+        }
+    }
+    
+    // MARK: Lookup Data
+    func cleanupLookup(force: Bool, alreadyInQueue: Bool, update: Update?) {
```
michalrentka, line 384: I don't love the `alreadyInQueue` parameter. You specify `_cleanupLookup` function anyway, so I'd just have separate functions, where `cleanupLookup` calls `_cleanupLookup` on proper queue and if you know you're on proper queue you can call `_cleanupLookup` directly.

I'd also change the name to `cleanupLookupIfNeeded` since you check for finished tasks in it.

Zotero/Controllers/IdentifierLookupController.swift
```diff
@@ -0,0 +1,445 @@
...
+            // If forced,
+            // or all lookups have been saved or failed, all attachments have finished, and either no presenter is assigned, or it doesn't present currently,
+            // then cleanup.
+            lookupData = []
+            lookupSavedCount = 0
+            lookupFailedCount = 0
+            DDLogInfo("IdentifierLookupController: cleaned up lookup data")
+            guard let update else { return }
+            observable.on(.next(update))
+        }
```
michalrentka, line 416:
```suggestion
            if force {
                cleanup(update: update)
            }
            if lookupRemainingCount == 0 && remoteFileDownloader.batchData.2 == 0 {
                // If all lookups have been saved or failed, all attachments have finished
                guard let presenter else {
                    // and no presenter is assigned, then cleanup
                    cleanup(update: update)
                    return
                }
                // and presenter is assigned
                var isPresenting = false
                inMainThread(sync: true) {
                    isPresenting = presenter.isPresenting()
                }
                if !isPresenting {
                    // but is not presenting, then cleanup
                    cleanup(update: update)
                }
            }
            
            func cleanup(update: Update?) {
                lookupData = []
                lookupSavedCount = 0
                lookupFailedCount = 0
                DDLogInfo("IdentifierLookupController: cleaned up lookup data")
                guard let update else { return }
                observable.on(.next(update))
            }
        }
```

michalrentka (changes requested):

Zotero/Controllers/Attachment Downloader/RemoteAttachmentDownloader.swift
```diff
@@ -80,38 +99,66 @@ final class RemoteAttachmentDownloader {
     }
 
     func download(data: [(Attachment, URL, String)]) {
-        self.queue.async(flags: .barrier) { [weak self] in
-            guard let self = self else { return }
+        self.accessQueue.async(flags: .barrier) { [weak self] in
```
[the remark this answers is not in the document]
mvasilak: `createDownload` that is called within the block modifies local properties. Moved it as a nested function to make it more clear. Check https://github.com/zotero/zotero-ios/pull/707/commits/018260333c910c4086b3674f8b53b23d041e1796

Zotero/Scenes/Detail/Lookup/ViewModels/LookupActionHandler.swift
```diff
@@ -11,90 +11,94 @@ import Foundation
...
+                    .observe(on: MainScheduler.instance)
+                    .subscribe(with: viewModel) { [weak self] viewModel, update in
+                        guard let self else { return }
+                        switch update.kind {
+                        case .lookupError(let error):
+                            self.update(viewModel: viewModel) { state in
+                                state.lookupState = .failed(error)
+                            }
+
+                        case .noIdentifiersDetected:
```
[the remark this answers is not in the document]
mvasilak: `.lookupError` is used for errors observed from `LookupWebViewHandler`, while `.noIdentifiersDetected` is not considered an error in the `IdentifierLookupController` context. I modified it in `LookupActionHandler`, because it is handled as an error, to display the failure message. Do you think we should merge them both in one case?

Zotero/Controllers/IdentifierLookupController.swift
```diff
@@ -0,0 +1,445 @@
...
+            guard let self = self else { return }
+            let lookupSettings = LookupWebViewHandler.LookupSettings(libraryIdentifier: libraryId, collectionKeys: collectionKeys)
+            if self.lookupWebViewHandlersByLookupSettings[lookupSettings] != nil {
+                lookupData = self.lookupData
+                return
+            }
+            inMainThread(sync: true) {
+                if let webView = self.webViewProvider?.addWebView() {
+                    let lookupWebViewHandler = LookupWebViewHandler(lookupSettings: lookupSettings, webView: webView, translatorsController: self.translatorsController)
+                    self.lookupWebViewHandlersByLookupSettings[lookupSettings] = lookupWebViewHandler
```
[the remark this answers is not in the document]
mvasilak: `WKWebView` construction needs to happen in main thread, otherwise it crashes.

Zotero/Controllers/Attachment Downloader/RemoteAttachmentDownloader.swift
```diff
@@ -80,38 +99,66 @@ final class RemoteAttachmentDownloader {
...
 
-        let operation = RemoteAttachmentDownloadOperation(url: url, file: file, apiClient: self.apiClient, fileStorage: self.fileStorage, queue: self.queue)
+        let progress = Progress(totalUnitCount: 100)
+        let operation = RemoteAttachmentDownloadOperation(url: url, file: file, progress: progress, apiClient: self.apiClient, fileStorage: self.fileStorage, queue: self.processingQueue)
         operation.finishedDownload = { [weak self] result in
-            self?.queue.async(flags: .barrier) {
-                guard let self = self else { return }
+            self?.processingQueue.async(flags: .barrier) {
+                guard let self else { return }
                 self.finish(download: download, file: file, attachment: attachment, parentKey: parentKey, result: result)
```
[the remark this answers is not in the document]
mvasilak: Correct, check https://github.com/zotero/zotero-ios/pull/707/commits/cb8af619324e87b09b192b0a566960bdc2f325f7

Zotero/Scenes/Detail/Lookup/ViewModels/LookupActionHandler.swift
```diff
@@ -11,90 +11,94 @@ import Foundation
...
-
-            self.remoteFileDownloader.observable
-                .observe(on: MainScheduler.instance)
-                .subscribe(with: viewModel, onNext: { [weak self] viewModel, update in
-                    switch update.kind {
-                    case .ready(let attachment):
-                        self?.finish(download: update.download, attachment: attachment, in: viewModel)
-                    case .cancelled, .failed, .progress: break
+        case .initialize:
+            identifierLookupController.initialize(libraryId: libraryId, collectionKeys: collectionKeys) { [weak self] lookupData in
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/707/commits/87174e01219837a21b0e7d7d9239c66dd0308be7

Zotero/Controllers/IdentifierLookupController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/707/commits/cac7e7339967ac0f968550762198127a4132c069

Zotero/Controllers/IdentifierLookupController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/707/commits/97438bb929d4e0d770d16970d92326c45fac536b

Zotero/Scenes/Master/Libraries/Views/LibrariesViewController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/707/commits/c79834da3fe9b9acead80374f8016ff0327d007b

Zotero/Controllers/IdentifierLookupController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/707/commits/c79834da3fe9b9acead80374f8016ff0327d007b

Zotero/Controllers/IdentifierLookupController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/707/commits/1903e2ba890cb62198179c812fb54e43b0112a4a

Zotero/Controllers/IdentifierLookupController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/707/commits/f73cdf75e9b42a27d48d096deeff37488fd0989d

Zotero/Controllers/IdentifierLookupController.swift
```diff
@@ -0,0 +1,445 @@
...
+                    if identifiers.isEmpty {
+                        cleanupLookup(force: true, alreadyInQueue: false, update: Update(kind: .noIdentifiersDetected))
+                    } else {
+                        let enqueuedIdentifiers = identifiers.map({ identifier(from: $0) })
+                        enqueueLookup(for: enqueuedIdentifiers)
+                        observable.on(.next(Update(kind: .identifiersDetected(identifiers: enqueuedIdentifiers))))
+                    }
+                    
+                case .item(let data):
+                    guard let lookupId = data["identifier"] as? [String: String] else { return }
```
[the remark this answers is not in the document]
mvasilak: I made it a warning, should it be an error instead?
Check https://github.com/zotero/zotero-ios/pull/707/commits/83db6f69d8dd797de76a743c13c55c03f9b9c64d

Zotero/Controllers/IdentifierLookupController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/707/commits/1196ff95e7aa78d9f47d4d5b321dd5dd96f0c6a4

Zotero/Controllers/IdentifierLookupController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/707/commits/46ae8d700e03a23b5cdc03ac6026b9f5dee80234

Zotero/Scenes/Detail/Lookup/ViewModels/LookupActionHandler.swift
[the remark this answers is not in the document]
michalrentka: I'd maybe consider no identifiers detected as error, because the input was most likely wrong then. But if there are valid reasons why it's not detected while not being an error this is fine then.

Zotero/Controllers/IdentifierLookupController.swift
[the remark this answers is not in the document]
michalrentka: Sure, but `lookupWebViewHandlersByLookupSettings` should be accessed/written from `accessQueue`. I guess in this case it doesn't matter since it's synchronous and inside `accessQueue` barrier block. I'd probably still keep web view initialization on main queue and assign it on `accessQueue`, but this shouldn't create any race conditions.

Zotero/Controllers/IdentifierLookupController.swift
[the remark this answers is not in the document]
michalrentka: Warning is fine

Zotero/Controllers/IdentifierLookupController.swift
michalrentka: Didn't know you can check queue like that, nice :)

Zotero/Controllers/IdentifierLookupController.swift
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/707/commits/60b3c43d1a29b3e89e2967aad066e81f71b02bb0

Zotero/Controllers/IdentifierLookupController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/707/commits/7b9aa5de1ea91ae6227dbbbd0bd94261c2e0745d

Zotero/Scenes/Detail/Lookup/ViewModels/LookupActionHandler.swift
```diff
@@ -106,82 +110,22 @@ final class LookupActionHandler: ViewModelActionHandler, BackgroundDbProcessingA
...
-                    }
-                } catch let error {
-                    DDLogError("LookupActionHandler: can't create item(s) - \(error)")
-
-                    inMainThread { [weak self] in
-                        guard let self = self, let viewModel = viewModel else { return }
-                        let failedData = LookupState.LookupData(identifier: identifier, state: .failed)
-                        self.update(lookupData: failedData, in: viewModel)
-                    }
+                    state.lookupState = .lookup(data)
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/707/commits/a29a864c86ff0eabc79981d05550e39492fdba52

To make things hopefullly simpler:
* A separate `LookupState` for `waitingInput` was added, and `loadingIdentifiers` is used as before just for waiting until the first updates arrive. 
* Since `IdentifierLookupController` needs to hold the lookup state anyway, it passes it to every update it publishes, so `LookupActionHandler` just needs to copy it when needed.
* Also fixed some potential edge case bugs in IdentifierLookupController, by making the workflow for publishing updates more strict.

Zotero/Scenes/Detail/Lookup/ViewModels/LookupActionHandler.swift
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/707/commits/a29a864c86ff0eabc79981d05550e39492fdba52

It merges the `identifiersDetected` and `noIdentifiersDetected` states, and lets `LookupActionHandler` to determine how the view model should be affected when the identifiers array passed with the update is empty.

michalrentka:
Code looks good to me now, there are a few issues I noticed when testing:

1. Looks a bit weird that one part is bigger than the other, I'd probably make both small:
IMG_0408

2. When I add multiple identifiers, the popup correctly scales and takes maximum possible height. But when I close the popup and re-open with multiple items, it doesn't scale up properly:
IMG_0409

3. When downloads are in progress, the table view reloads for progress updates are creating this weird blinking:
https://github.com/zotero/zotero-ios/assets/1022956/3d88f21f-54d9-4ee3-a8cf-dbc85412c01f

mvasilak:
@michalrentka 

>     1. Looks a bit weird that one part is bigger than the other, I'd probably make both small:

I differentiated them in purpose, to make one more button like :) 
Probably the color change suffices, I'll make them the same size.

>     2. When I add multiple identifiers, the popup correctly scales and takes maximum possible height. But when I close the popup and re-open with multiple items, it doesn't scale up properly:

I will investigate this one.

>     3. When downloads are in progress, the table view reloads for progress updates are creating this weird blinking:
>        https://github.com/zotero/zotero-ios/assets/1022956/3d88f21f-54d9-4ee3-a8cf-dbc85412c01f

If I'm not mistaken, that is a prexisting problem. I suggest we open a separate issue.

michalrentka:
> > ```
> > 3. When downloads are in progress, the table view reloads for progress updates are creating this weird blinking:
> >    https://github.com/zotero/zotero-ios/assets/1022956/3d88f21f-54d9-4ee3-a8cf-dbc85412c01f
> > ```
> 
> If I'm not mistaken, that is a prexisting problem. I suggest we open a separate issue.

That's possible :) feel free to open new issue

mvasilak:
@michalrentka ready for another look, regarding the UI issues you noticed.

michalrentka:
@mvasilak I've encountered some more weirdness, but it works better now.

1. First I accidentally pasted the same 4 identifiers twice and translation got stuck at some point:
IMG_0411

2. Then I tried again with original 4 identifiers and the first one never translated as well. If there is an error it should show it:
IMG_0412

3. And then when I tried `Cancel All` button I was shown the initial popup with input, so I tried adding the same identifiers and starting translation, but then it just got stuck at the beginning:
IMG_0413

I was testing with these identifiers: "arXiv:2307.06949, arXiv:2307.06925, arXiv:2307.06526, arXiv:2307.06335"

mvasilak:
@michalrentka latest issues have been addressed and PR is ready for another review. Changes of note:
* When canceling all the lookup window is dismissed. If we want to keep it open, some minor changes are needed, to re-initialize the web views, as they are now more aggresively cleaned up when all identifiers have been processed.
* Error messages have been improved to accommodate the case where there are ongoing lookups in the background.

mvasilak:
Rebased to `master`

mvasilak:
Rebased to `master`. @michalrentka @dstillman I wouldn't mind another quick look (nothing changed, just in case you find an issue due to other changes since last time). I think this is ready for release, and I propose to move the remaining taks to another issue, with low priority, as they are just covering edge use cases.

michalrentka (approved):
I think it's good to merge this, it definitely improves and fixes some things and we released necessary bug fixes that we needed out quickly. @dstillman any thoughts? Otherwise I'll merge this next week.

PR #709, opened 2023-06-27
Added github actions support for unit tests

michalrentka:
Includes additional unit test fixes which my Xcode didn't show before. All tests are passing now: https://github.com/michalrentka/zotero-ios/actions/runs/5387866540/jobs/9779768175


Zotero/Scenes/Detail/PDF/PDFCoordinator.swift
```diff
@@ -357,7 +357,7 @@ extension PDFCoordinator: PdfAnnotationsCoordinatorDelegate {
                 DDLogError("PDFCoordinator: can't render annotation image - \(error)")
                 elementProvider([])
             }
-            .disposed(by: disposeBag)
+            .disposed(by: self.disposeBag)
```
mvasilak, line 360: Since using `self` for instance properties is unnecessary, unless there is one with the same name in the local scope, we should decide if we want to adhere to always prefixing w/ `self`, as part of the project style.

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,38 @@
...
+
+on: push
+
+jobs:
+  tests:
+    name: Run Xcode Tests
+    runs-on: macos-latest
+    strategy:
+      matrix:
+        xcode: ["14.2.0"]
```
mvasilak, line 11: Should we use `latest` or `latest-stable` instead?

mvasilak (approved):

.github/workflows/run-tests.yml
michalrentka: `14.2.0` is their latest currently. I'd keep it explicitly specified just because in future we might not want to update to latest Xcode as fast as Github does. For example some third party library is broken on new Xcode and we're waiting for an update or we just don't have time to update issues related to new Xcode, so tests will be broken in the meanwhile because actions will run on new Xcode already.

Zotero/Scenes/Detail/PDF/PDFCoordinator.swift
michalrentka: I actually got an error here when using `xcbuild` during testing. I can check whether it'll still show an error here.

Anyway in the whole codebase I added `self` everywhere explicitly. I personally find it nicer because you immediately know which `disposeBag` (in this case) is being used, without having to look whether there is a local `disposeBag` instance somewhere. I know Xcode highlights them differently, but Xcode highlights are also broken most of the time. But I know most people find it annoying and even Apple in their code remove `self` everywhere. If you think that explicit `self`s are mostly unnecessary we can start reverting it. I don't feel that strongly about one or another approach.

mvasilak: Oh, because github actions use Xcode 14.2, and implicit use of `self` for this case was added in 14.3!

PR #713, opened 2023-06-28
Show proper hierarchy when there is restored state

### LINKED ISSUE #703: iPhone state restoration
When restoring PDF state on iPhone, the collection screen appears at first, but then when you select some collection the PDF just pops up without animation. The PDF should appear immediately, I think we can select "All Items" collection in background, so that if PDF is closed the user sees all items in given library.

mvasilak:
closes #703 

Restores proper hierarchy for both iPhone & iPad.
Updates `Defaults.shared.selectedLibrary` & `Defaults.shared.selectedCollectionId` accordingly.

Tested cases, for both devices:
* open pdf, close app, launch app with restored state, pdf is shown, and selected collection should be the same
* open pdf, close app, share with extension to a different collection, launch app with restored state, pdf is shown, and selected collection should be the same

Easiest way to be certain that you get restoration of your previous state, is to close normally the app, restart simulator/device, launch the app again.

### DIFF
--- a/Zotero/Extensions/NSUserActivity+Activities.swift
+++ b/Zotero/Extensions/NSUserActivity+Activities.swift
@@ -11,6 +11,7 @@ import Foundation
 struct RestoredStateData {
     let key: String
     let libraryId: LibraryIdentifier
+    let collectionId: CollectionIdentifier?
 }
 
 extension NSUserActivity {
@@ -21,9 +22,13 @@ extension NSUserActivity {
         return NSUserActivity(activityType: self.mainId)
     }
 
-    static func pdfActivity(for key: String, libraryId: LibraryIdentifier) -> NSUserActivity {
+    static func pdfActivity(for key: String, libraryId: LibraryIdentifier, collectionId: CollectionIdentifier) -> NSUserActivity {
         let activity = NSUserActivity(activityType: self.pdfId)
-        activity.addUserInfoEntries(from: ["key": key, "libraryId": self.libraryIdToString(libraryId)])
+        var pdfUserInfo: [AnyHashable: Any] = ["key": key, "libraryId": libraryIdToString(libraryId)]
+        if let collectionIdData = try? JSONEncoder().encode(collectionId) {
+            pdfUserInfo["collectionId"] = collectionIdData
+        }
+        activity.addUserInfoEntries(from: pdfUserInfo)
         return activity
     }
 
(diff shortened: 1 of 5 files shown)


Zotero/Extensions/NSUserActivity+Activities.swift
```diff
@@ -11,6 +11,7 @@ import Foundation
 struct RestoredStateData {
     let key: String
     let libraryId: LibraryIdentifier
+    let collectionId: CollectionIdentifier?
```
michalrentka, line 14: Should `CollectionIdentifier` here be optional? I know it's not stored yet, but we could default to all items or `Defaults.shared.selectedCollectionId` for first-time use. Otherwise library and collection should always be available?

Zotero/Scenes/AppCoordinator.swift
```diff
@@ -121,6 +122,16 @@ final class AppCoordinator: NSObject {
...
+        guard let userActivity = connectionOptions?.userActivities.first ?? session?.stateRestorationActivity, let data = userActivity.restoredStateData else { return }
+        // If scene had state stored, check if defaults need to be updated first
+        DDLogInfo("AppCoordinator: Preprocessing restored state - \(data)")
+        Defaults.shared.selectedLibrary = data.libraryId
+        if let collectionId = data.collectionId {
+            Defaults.shared.selectedCollectionId = collectionId
+        }
+    }
+    
     private func process(connectionOptions: UIScene.ConnectionOptions, session: UISceneSession) {
```
michalrentka, line 135: We load the `RestoredStateData` in new `preprocess(connectionOptions:session:)` function and then load them again a few lines later here. Maybe the `preprocess` function could return `RestoredStateData?` and they could just be passed to `process(restoredState data: RestoredStateData, urlContext: UIOpenURLContext)` after `self.show(viewController: viewController, in: window, animated: animated)` happens.

Zotero/Scenes/Main/Views/MainViewController.swift
```diff
@@ -177,6 +132,9 @@ extension MainViewController: UISplitViewControllerDelegate {
 
 extension MainViewController: MainCoordinatorDelegate {
     func showItems(for collection: Collection, in library: Library, isInitial: Bool) {
+        if !isInitial {
```
michalrentka, line 135: Why was it moved from `MasterTopCoordinator` to here? Did it omit some cases?

michalrentka (changes requested):

Zotero/Extensions/NSUserActivity+Activities.swift
mvasilak: If we have a default option as fallback, we could do that. I'll check existance in order,
user info value > `Defaults.shared.selectedCollectionId` > all items

Zotero/Scenes/Main/Views/MainViewController.swift
mvasilak: Yes, it omitted the new code paths for restoration, because it is called  directly in the AppCoordinator. 
However, now that preprocessing saves the defaults from restored state, perhaps it is not needed to be moved,
but it seems more correct that way, as previously the deeper `showItems` didn't use at all the `isInitial` argument.

Zotero/Scenes/AppCoordinator.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/713/commits/76c09809ca57bccf4a83c63cbde1c739161141d2

michalrentka (dismissed):

michalrentka:
@mvasilak actually I just re-tested the most recent changes and I can still reproduce the initial issue on iPhone. When PDF is restored, I first see collections screen and when I select a collection the PDF opens up. The PDF should open immediately.

mvasilak:
@michalrentka can you provide exact steps to what you are doing to reproduce the isssue?

michalrentka:
@mvasilak the second time I open Zotero (after killing the app) I have to tap some collection before PDF appears: https://www.dropbox.com/s/ciccq1s4017rqrn/RPReplay_Final1688563410.MP4?dl=0

PR #714, opened 2023-06-30
Fixed storing of more complicated structures inside annotationPosition field

michalrentka:


Zotero/Models/API/ItemResponse.swift
```diff
@@ -331,15 +331,22 @@ struct ItemResponse {
...
             } else if let _value = object.value as? Int {
                 value = "\(_value)"
             } else if let _value = object.value as? Double {
                 value = "\(_value.rounded(to: 3))"
+            } else if let _value = object.value as? Bool {
+                value = "\(_value)"
             } else {
-                value = "\(object.value)"
+                // If `object.value` is not a basic type (string or number) convert it to JSON and store JSON string
+                if let data = try? JSONSerialization.dataWithRoundedDecimals(withJSONObject: object.value), let _value = String(data: data, encoding: .utf8) {
```
mvasilak, line 345: This nested if-else could be part of the outer one.

mvasilak (approved):
Couple of minor comments.

PR #716, opened 2023-07-04
Refactored Coordinator.navigationController to be weak instead of unowned

michalrentka:
I couldn't reproduce the crash myself, but obviously it happened that `navigationController` was deinitialized while `PDFCoordinator` instance was presented and when going back from PDF to previous screen the app would crash. Making `navigationController` in `Coordinator` protocol weak will solve this and potentially there may be user reports of UI issues which will point us in right direction.

### DIFF
--- a/Zotero/Scenes/Detail/DetailCoordinator.swift
+++ b/Zotero/Scenes/Detail/DetailCoordinator.swift
@@ -253,10 +253,11 @@ final class DetailCoordinator: Coordinator {
     }
 
     private func showWebView(for url: URL) {
+        guard let currentNavigationController = self.navigationController else { return }
         let controller = WebViewController(url: url)
         let navigationController = UINavigationController(rootViewController: controller)
         navigationController.modalPresentationStyle = .fullScreen
-        self.navigationController?.present(navigationController, animated: true, completion: nil)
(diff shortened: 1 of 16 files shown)


Zotero/Scenes/Detail/DetailCoordinator.swift
```diff
@@ -241,14 +249,14 @@ final class DetailCoordinator: Coordinator {
...
-        self.navigationController.present(navigationController, animated: true, completion: nil)
+        self.navigationController?.present(navigationController, animated: true, completion: nil)
     }
 
     private func showWebView(for url: URL) {
         let controller = WebViewController(url: url)
         let navigationController = UINavigationController(rootViewController: controller)
         navigationController.modalPresentationStyle = .fullScreen
-        self.navigationController.present(navigationController, animated: true, completion: nil)
+        self.navigationController?.present(navigationController, animated: true, completion: nil)
```
mvasilak, line 259: It will be more clear if the method starts with (or something like it)
```
        guard let currentNavigationController = self.navigationController else { return }
```
and then the local `navigationController` is initialized.

mvasilak (approved):

PR #718, opened 2023-07-06
Show proper hierarchy when there is restored state

mvasilak:
closes https://github.com/zotero/zotero-ios/issues/703

Restores proper hierarchy for both iPhone & iPad.
Updates Defaults.shared.selectedLibrary & Defaults.shared.selectedCollectionId accordingly.

Tested cases, for both devices:
* open pdf, close app, launch app with restored state, pdf is shown, and selected collection should be the same
* open pdf, close app, share with extension to a different collection, launch app with restored state, pdf is shown, and selected collection should be the same

Easiest way to be certain that you get restoration of your previous state, is to close normally the app, restart simulator/device, launch the app again.

### DIFF
--- a/Zotero/Scenes/AppCoordinator.swift
+++ b/Zotero/Scenes/AppCoordinator.swift
@@ -107,6 +107,7 @@ final class AppCoordinator: NSObject {
             self.conflictAlertQueueController = nil
             self.controllers.userControllers?.syncScheduler.syncController.set(coordinator: nil)
         } else {
+            preprocess(connectionOptions: connectionOptions, session: session)
             let controller = MainViewController(controllers: self.controllers)
             viewController = controller
 
@@ -121,6 +122,16 @@ final class AppCoordinator: NSObject {
         self.process(connectionOptions: options, session: session)
     }
 
+    private func preprocess(connectionOptions: UIScene.ConnectionOptions?, session: UISceneSession?) {
+        guard let userActivity = connectionOptions?.userActivities.first ?? session?.stateRestorationActivity, let data = userActivity.restoredStateData else { return }
+        // If scene had state stored, check if defaults need to be updated first
+        DDLogInfo("AppCoordinator: Preprocessing restored state - \(data)")
+        Defaults.shared.selectedLibrary = data.libraryId
+        if let collectionId = data.collectionId {
+            Defaults.shared.selectedCollectionId = collectionId
+        }
+    }
+    
     private func process(connectionOptions: UIScene.ConnectionOptions, session: UISceneSession) {
         if let urlContext = connectionOptions.urlContexts.first, let urlController = self.controllers.userControllers?.customUrlController {
             // If scene was started from custom URL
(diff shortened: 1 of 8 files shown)


Zotero/Scenes/AppCoordinator.swift
```diff
@@ -258,7 +282,9 @@ final class AppCoordinator: NSObject {
                     case .local, .localAndChangedRemotely:
                         let file = Files.attachmentFile(in: libraryId, key: key, filename: filename, contentType: contentType)
                         url = file.createUrl()
-                        library = try coordinator.perform(request: ReadLibraryDbRequest(libraryId: libraryId))
+                        let (_collection, _library) = try coordinator.perform(request: ReadCollectionAndLibraryDbRequest(collectionId: collectionId, libraryId: libraryId))
```
michalrentka, line 285: So `ReadCollectionAndLibraryDbRequest` returns `nil` for custom collections. I've checked and I think we can just change `ReadCollectionAndLibraryDbRequest` to return `Collection` for custom `CollectionIdentifier`s. So the result value can also be `(Collection, Library)` there too.

michalrentka (changes requested):

Zotero/Scenes/AppCoordinator.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/718/commits/32026a7d276daa10e4c9408cc2411c435753002c

Zotero/Scenes/AppCoordinator.swift
```diff
@@ -218,9 +233,17 @@ final class AppCoordinator: NSObject {
...
-
+              let (url, library, collection) = self.loadRestoredStateData(forKey: data.key, libraryId: data.libraryId, collectionId: data.collectionId) else { return }
+        if let collection {
+            DDLogInfo("AppCoordinator: show restored state - \(data.key); \(data.libraryId); \(data.collectionId); \(url.relativePath)")
+            mainController.showItems(for: collection, in: library, isInitial: false)
+        } else {
+            DDLogInfo("AppCoordinator: show restored state - \(data.key); \(data.libraryId); \(url.relativePath)")
+            // Collection is missing, show all items instead
+            let collection = Collection(custom: .all)
+            mainController.showItems(for: collection, in: library, isInitial: true)
```
michalrentka, line 244: I'd probably even show a warning/error here, because the collection should be available.

Also last question, why is `isInital: true` here and `isInitial: false` above on 239?

mvasilak: `true` doesn't save the selected collection to defaults.

mvasilak: Check https://github.com/zotero/zotero-ios/pull/718/commits/5e9013080fe4609379b740eacb93f7b89d94dd40

michalrentka: Might be worth renaming `isInitial` to something else then, doesn't make sense when `isInitial` is `true` and `false` at the same time based on whether the collection is available.

michalrentka (approved):

Zotero/Scenes/AppCoordinator.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/718/commits/2a1d6715294a69792e65ad6012d7c6a273dcb875

michalrentka (approved):

PR #723, opened 2023-07-10
Improved search filters

### LINKED ISSUE #720: Improve item search

michalrentka:
Added html free content field (notes and comments) to database, improved searc filters to contain new field and other fields too. Closes #720.


Zotero/Extensions/String+Extensions.swift
```diff
@@ -41,4 +41,9 @@ extension String {
         else { return nil }
         return ext.takeRetainedValue() as String
     }
+
+    var strippedHtmlTags: String {
+        guard !self.isEmpty else { return self }
+        return self.replacingOccurrences(of: "<[^>]*>", with: "", options: .regularExpression, range: nil)
```
dstillman, line 47: This is probably OK for notes, which are technically HTML and theoretically shouldn't have raw or unclosed angled brackets, but annotation comments aren't HTML — they just borrow HTML syntax for rich text. The first four HTML tags [here](https://www.zotero.org/support/kb/rich_text_bibliography) (with no attributes) are valid. `He had <10 dogs but definitely >2.` would be a valid comment, and you'd want to be able to find it with a search for `dogs`.

(The same applies to item fields, though in that case all of the tags on the linked page are supported.)

dstillman: > so for annotation comments I'll just remove i/b/sub/sup tags? 

Yes.

> Should I improve this for notes to only include tags we use too?

No, notes are actually HTML, so stripping all tags is appropriate.

> Do all item fields support those linked html tags?

Er, no, sorry, I guess just titles currently. It's possible HTML tags could be placed in Abstract by a translator, but they aren't technically HTML.

michalrentka: @dstillman I don't support those tags in titles on iOS too yet. So do I just strip them or do I render them?

dstillman: If you have the ability to do rich text, it would be great to render those with the respective formatting, as we [just did for the web library](https://github.com/zotero/web-library/issues/515). See link there to the desktop code for this.

If you can't do rich text currently…I don't know. The desktop app always just showed the raw value before it started rendering the rich text.

For searching, strip.

Zotero/Models/Database/Database.swift
```diff
@@ -44,6 +44,32 @@ struct Database {
...
+            let content: String
+
+            switch rawType {
+            case ItemTypes.note:
+                guard let _content = fields.first(where: { $0["key"] as? String == FieldKeys.Item.note })?["value"] as? String, !_content.isEmpty else { return }
+                content = _content.strippedHtmlTags
+
+            case ItemTypes.annotation:
+                guard let _content = fields.first(where: { $0["key"] as? String == FieldKeys.Item.Annotation.comment })?["value"] as? String, !_content.isEmpty else { return }
+                content = _content.strippedHtmlTags
```
mvasilak, line 66: Is the annotation comment in the previous scheme html?

mvasilak (approved):

Zotero/Extensions/String+Extensions.swift
dstillman: > Items list now renders those rich text tags

I'm not seeing anything that specifically handles the two supported `span` attributes. We shouldn't be rendering any other HTML.

> Do I maybe render rich text in item detail while not editing?

Yeah, you totally can. We just never got around to implementing that on desktop.

Zotero/Models/Database/Database.swift
michalrentka: No, I forgot to change it there, good catch 👍

Zotero/Extensions/String+Extensions.swift
michalrentka: @dstillman I used existing `HtmlAttributedStringConverter` (https://github.com/zotero/zotero-ios/blob/master/Zotero/Controllers/HtmlAttributedStringConverter.swift) which is used in annotation comments. 

But now I see it supports only `b`, `i`, `sup`, `sub`, so I'll add support for those 2 `span`s: https://github.com/zotero/zotero-ios/blob/f9bfbf88afe8b2630d135fbe04cabb1287543782/Zotero/Models/StringAttributes.swift#L22-L30

Zotero/Models/Database/Database.swift
michalrentka: https://github.com/zotero/zotero-ios/pull/723/commits/376e67346d30e37e2f30474a47baec17ee3b4e32

Zotero/Controllers/Database/Requests/StoreItemsDbResponseRequest.swift
```diff
@@ -191,6 +191,10 @@ struct StoreItemDbRequest: DbResponseRequest {
 
             case (FieldKeys.Item.note, _) where item.rawType == ItemTypes.note:
                 item.baseTitle = NotePreviewGenerator.preview(from: value) ?? value
+                item.htmlFreeContent = value.isEmpty ? nil : value
```
mvasilak, line 194: Why is `value` not stripped of html tags as before?

Zotero/Scenes/Detail/Items/Views/ItemsTableViewHandler.swift
```diff
@@ -322,9 +322,17 @@ extension ItemsTableViewHandler: UITableViewDataSource {
             // Create and cache attachment if needed
             self.viewModel.process(action: .cacheItemAccessory(item: item))
 
+            let title: NSAttributedString
+            if let _title = self.viewModel.state.itemTitles[item.key] {
+                title = _title
+            } else {
+                self.viewModel.process(action: .cacheItemTitle(key: item.key, title: item.displayTitle))
+                title = self.viewModel.state.itemTitles[item.key] ?? NSAttributedString()
```
mvasilak, line 330:
```suggestion
                title = self.viewModel.state.itemTitles[item.key, default: NSAttributedString()]
```

mvasilak (approved):

Zotero/Controllers/Database/Requests/StoreItemsDbResponseRequest.swift
michalrentka: It's included in `NotePreviewGenerator.preview(from:)`, I just forgot about that initially.

mvasilak: That is assigned to `item.baseTitle`, `value` is created in `let value = data.fields[keyPair] ?? ""`.
Is `NotePreviewGenerator.preview(from:)` used also before saving the initial `value` in the database?
What am I missing here?

michalrentka: Oh you meant the `?? value` not `value` in `NotePreviewGenerator.preview(from: value)`? `NotePreviewGenerator.preview(from: value)` returns `nil` only when the string is empty, so I'll actually change this to `NotePreviewGenerator.preview(from: value) ?? ""` which makes more sense. And `NotePreviewGenerator` applies the stripping, so it should be applied to `htmlFreeContent`.

PR #733, opened 2023-07-28
Fixed reentrancy issue in pdf reader

michalrentka:
There was a RxSwift reentrancy issue in `AnnotationsViewController.loadAttributedComment(for:)` function.  It was caused by calling `self.viewModel.process(action: .parseAndCacheComment(key: annotation.key, comment: comment))` during ongoing update which then pushed another update.

### DIFF
--- a/Zotero/Scenes/Detail/PDF/Views/AnnotationsViewController.swift
+++ b/Zotero/Scenes/Detail/PDF/Views/AnnotationsViewController.swift
@@ -109,14 +109,28 @@ final class AnnotationsViewController: UIViewController {
             })
 
         case .options(let sender):
-            self.coordinatorDelegate?.showCellOptions(for: annotation, userId: self.viewModel.state.userId, library: self.viewModel.state.library, sender: sender, userInterfaceStyle: self.viewModel.state.interfaceStyle,
-                                                      saveAction: { [weak self] key, color, lineWidth, pageLabel, updateSubsequentLabels, highlightText in
-                                                          self?.viewModel.process(action: .updateAnnotationProperties(key: key.key, color: color, lineWidth: lineWidth, pageLabel: pageLabel,
-                                                                                                                      updateSubsequentLabels: updateSubsequentLabels, highlightText: highlightText))
-                                                      },
-                                                      deleteAction: { [weak self] key in
-                                                          self?.viewModel.process(action: .removeAnnotation(key))
-                                                      })
+            self.coordinatorDelegate?.showCellOptions(
+                for: annotation,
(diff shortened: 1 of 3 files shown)


Zotero/Scenes/Detail/PDF/Views/AnnotationsViewController.swift
```diff
@@ -478,29 +516,34 @@ final class AnnotationsViewController: UIViewController {
...
-                          })
-                          .disposed(by: self.disposeBag)
+        .keyboardWillShow
+        .observe(on: MainScheduler.instance)
+        .subscribe(with: self, onNext: { `self`, notification in
+            if let data = notification.keyboardData {
+                self.setupTableView(with: data)
+            }
+        })
+        .disposed(by: self.disposeBag)
```
mvasilak, line 526: Shouldn't these be moved a tab ahead to be properly aligned?

mvasilak (approved):

Zotero/Scenes/Detail/PDF/Views/AnnotationsViewController.swift
michalrentka: Yes, forgot to indent it, good catch

mvasilak:
@michalrentka this Reentrancy fix causes a [regression](https://forums.zotero.org/discussion/107608/ipad-app-not-displaying-text-comments-in-annotations), where annotation comments don't appear in the side bar.

The reason is that the returned value here doesn't use the updated state:
https://github.com/zotero/zotero-ios/blob/48e227dd89e9b0f94513d5385a13385034ffcb31/Zotero/Scenes/Detail/PDF/Views/AnnotationsViewController.swift#L337

I suggest we re-introduce the reentrancy for the upcoming release, so that the regression is fixed, and we open an issue to properly tackle the reentrancy afterwards.

michalrentka:
> @michalrentka this Reentrancy fix causes a [regression](https://forums.zotero.org/discussion/107608/ipad-app-not-displaying-text-comments-in-annotations), where annotation comments don't appear in the side bar.
> 
> The reason is that the returned value here doesn't use the updated state:
> 
> https://github.com/zotero/zotero-ios/blob/48e227dd89e9b0f94513d5385a13385034ffcb31/Zotero/Scenes/Detail/PDF/Views/AnnotationsViewController.swift#L337
> 
> I suggest we re-introduce the reentrancy for the upcoming release, so that the regression is reversed, and we open an issue to properly tackle the reentrancy afterwards.

Thanks for pointing the issue. I was able to fix it with a rather simple fix so we shouldn't need to do anything else - #763

PR #735, opened 2023-08-01
Update SwiftGen & localizable strings

mvasilak:
* Updates SwiftGen
* Converts localizable strings that need pluralization to stringsdict


dstillman:
Do we really need to include SwiftGen (a whole 12 MB binary, etc.) in our repo? That seems like it should be an external dependency, or at least a submodule.

mvasilak:
@dstillman 

> Do we really need to include SwiftGen (a whole 12 MB binary, etc.) in our repo? That seems like it should be an external dependency, or at least a submodule.

It's not available via SPM, which is the dependency system we use. Moreover we just need the binary, we don't want to compile it ourselves, so submodule would also not solve this.

Per their instructions:
> We recommend that you unarchive the ZIP inside your project directory and commit its content to git. This way, all coworkers will use the same version of SwiftGen for this project.

We can instead use a build phase, or a pre-build script, to install via homebrew or mint. That should be installed first manually by any developer. @michalrentka any preference between the two?

michalrentka:
@mvasilak would Xcode Cloud still work? Can it install stuff from homebrew/mint? Having the same version included in project makes it easier so that devs, tests (github actions) and CI (xcode cloud) are all in sync, so that you don't get random error on one end while others seemingly work fine.

@dstillman any reason not to include it in repo other than size of repo? You'd have to download it anyway separately and make sure that you're using the same version.

Zotero/Scenes/General/Views/CollectionsPickerViewController.swift
```diff
@@ -124,16 +124,7 @@ class CollectionsPickerViewController: UICollectionViewController {
     }
 
     private func updateTitle(with selectedCount: Int) {
```
michalrentka, line 126: This doesn't need to be a separate function anymore.

Zotero/Scenes/Master/Settings/Storage/Views/StorageSettingsRow.swift
```diff
@@ -34,7 +34,7 @@ struct StorageSettingsRow: View {
     private func storageDataString(for data: DirectoryData?) -> String {
         guard let data = data, data.fileCount > 0 else { return "-" }
         let mbString = String(format: "%.2f", data.mbSize)
-        return (data.fileCount == 1 ? L10n.Settings.Storage.oneFile : L10n.Settings.Storage.multipleFiles(data.fileCount)) + " (\(mbString) MB)"
+        return (L10n.Settings.Storage.files(data.fileCount)) + " (\(mbString) MB)"
```
michalrentka, line 37: Can we include the `(\(mbString) MB)` in that localized string too?

michalrentka (approved):

michalrentka:
> The size, and general principles. Unless we use Git LFS (which I don't think makes sense here), having it in the repo means every time we update the version, we have another copy in the history, so cloning the repo means downloading every version we ever had instead of just the current version. Adding someone else's binary — let alone an entire app package — to your repo is just kind of an anti-pattern…

This is a bit of a special case though, but ok I understand the size can be an issue. @mvasilak can you check whether we can make Xcode Cloud work with homebrew/mint and make adjustments? I didn't research that since we didn't need it.

Zotero/Scenes/General/Views/CollectionsPickerViewController.swift
mvasilak: It's called in 2 different places, so it's better if we keep it as is.

dstillman:
I don't think using Homebrew makes sense — that presumably wouldn't be available in Xcode Cloud — but we could just script downloading and unzipping a specific ZIP from their GitHub release page into a git-ignored directory in a pre-build script, if the current version is wrong.

mvasilak:
Xcode Cloud does include Homebrew! https://developer.apple.com/documentation/xcode/making-dependencies-available-to-xcode-cloud#Use-a-custom-build-script-to-install-a-third-party-dependency-or-tool
So we can utilize it in a custom build script.

Zotero/Scenes/Master/Settings/Storage/Views/StorageSettingsRow.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/735/commits/72345d1a83b3648e2b9fc0c3d37cee155306d87d

michalrentka (approved):

PR #755, opened 2023-08-25
Dont select rows without action

### LINKED ISSUE #753: Row without tap action shouldn't show tap effect
Following up on https://github.com/zotero/zotero-ios/issues/750, while it's no longer showing the details view, it's still flashing the row gray when I tap, which makes it look like something is supposed to be happening.

michalrentka:
Fixes #753


mvasilak (approved):

dstillman:
Superseded by #756

PR #762, opened 2023-09-05
Fix size class transition issues

mvasilak:
Modifies `MasterContainerViewController` to inherit from `UINavigationController`. This allows split view controller to merge/separate navigation stacks when it collapses/expands. 

`MasterContainerViewController` is used for both iPhone and iPad, with these differences:
* in iPhone the bottom controller is not created
* for iPhones Pro Max and Plus, the default behavior of showing a split view controller in landscape is kept, but without the bottom controller

So the case that was previously
`UIDevice.current.userInterfaceIdiom == .pad`
becomes (in most cases)
`traitCollection.horizontalSizeClass == .regular && UIDevice.current.userInterfaceIdiom == .pad`

and the case that was previously
`UIDevice.current.userInterfaceIdiom == .phone`
becomes (in most cases)
`traitCollection.horizontalSizeClass == .compact || UIDevice.current.userInterfaceIdiom == .phone`

Trait transition logic has been updated in most common cases. Other existing cases should work as is, but we can update them as well, if we get relevant feedback.

### DIFF
--- a/Zotero/Scenes/Detail/Items/ViewModels/ItemsToolbarController.swift
+++ b/Zotero/Scenes/Detail/Items/ViewModels/ItemsToolbarController.swift
@@ -11,7 +11,7 @@ import UIKit
 import RealmSwift
 import RxSwift
 
-protocol ItemsToolbarControllerDelegate: AnyObject {
+protocol ItemsToolbarControllerDelegate: UITraitEnvironment {
     func process(action: ItemAction.Kind, button: UIBarButtonItem)
 }
 
@@ -78,19 +78,17 @@ final class ItemsToolbarController {
     }
 
     private func deviceSpecificFilters(from filters: [ItemsFilter]) -> [ItemsFilter] {
-        // There is different functionality on iPad and iPhone. iPhone shows tag filters in filter popup in items screen while iPad shows tag filters in master controller.
-        // So filter icon and description should always show up on iPhone, while it should not show up on iPad for tag filters.
-        // Therefore we ignore `.tag` filter on iPad and keep it on iPhone.
-        switch UIDevice.current.userInterfaceIdiom {
-        case .pad:
+        // There is different functionality based on horizontal size class. iPhone and compact iPad show tag filters in filter popup in items screen while iPad shows tag filters in master controller.
+        // So filter icon and description should always show up on iPhone and compact iPad, while it should not show up on regular iPad for tag filters.
+        // Therefore we ignore `.tag` filter on iPhone and compact iPad, and keep it on regular iPad.
+        if delegate?.traitCollection.horizontalSizeClass == .regular && UIDevice.current.userInterfaceIdiom == .pad {
             return filters.filter({
                 switch $0 {
                 case .tags: return false
                 case .downloadedFiles: return true
                 }
             })
-
-        default:
+        } else {
             return filters
         }
     }
--- a/Zotero/Scenes/Detail/Items/ViewModels/ItemsToolbarController.swift
+++ b/Zotero/Scenes/Detail/Items/ViewModels/ItemsToolbarController.swift
@@ -63,7 +63,7 @@ final class ItemsToolbarController {
             self.viewController.toolbarItems = self.createEditingToolbarItems(from: self.editingActions)
             self.updateEditingToolbarItems(for: state.selectedItems, results: state.results)
         } else {
-            let filters = self.deviceSpecificFilters(from: state.filters)
+            let filters = self.sizeClassSpecificFilters(from: state.filters)
             self.viewController.toolbarItems = self.createNormalToolbarItems(for: filters)
             self.updateNormalToolbarItems(for: filters, downloadBatchData: state.downloadBatchData, results: state.results)
         }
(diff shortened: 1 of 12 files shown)


Zotero/Scenes/Detail/Items/Views/ItemsFilterViewController.swift
```diff
@@ -61,7 +61,7 @@ class ItemsFilterViewController: UIViewController {
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
 
-        guard UIDevice.current.userInterfaceIdiom == .pad else { return }
+        guard traitCollection.horizontalSizeClass == .regular && UIDevice.current.userInterfaceIdiom == .pad else { return }
```
michalrentka, line 64: Is `viewWillAppear` called when size class changes? If not then the `preferredSize` should be updated on size class change.

Zotero/Scenes/Detail/Items/Views/ItemsFilterViewController.swift
```diff
@@ -98,7 +98,7 @@ class ItemsFilterViewController: UIViewController {
         self.downloadsTitleLabel.text = L10n.Items.Filters.downloads
         self.downloadsSwitch.isOn = self.downloadsFilterEnabled
 
-        guard UIDevice.current.userInterfaceIdiom == .phone else {
+        guard traitCollection.horizontalSizeClass == .compact || UIDevice.current.userInterfaceIdiom == .phone else {
```
michalrentka, line 101: Similar as above, doesn't this need an update on size class change?

michalrentka:
I added some comments which you are most likely aware of. Otherwise the idea is there, if subclassing `UINavigationController` doesn't create additional issues then it looks good.

Zotero/Scenes/Detail/Items/Views/ItemsFilterViewController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/762/commits/882576cc235304112997182051826044dcc89adb

mvasilak:
@dstillman is there any reason to diverge from the default behavior in large iPhones (Pro Max  & Plus), and not use a split view in landscape?

Zotero/Scenes/Detail/Items/Views/ItemsViewController.swift
```diff
@@ -684,66 +686,64 @@ final class ItemsViewController: UIViewController {
     @discardableResult
     private func setupSearchBar(for windowSize: CGSize) -> SearchBarPosition {
```
michalrentka, line 687: It's outside of scope of this, but I think we can finally move away from this terrible search bar placement and move it to the iOS default which is in navigation bar now. If you don't want to do it here feel free to create a ticket for it.

michalrentka (approved):

Zotero/Scenes/Detail/Items/Views/ItemsViewController.swift
mvasilak: Opened https://github.com/zotero/zotero-ios/issues/764

Zotero/Scenes/Master/MasterContainerViewController.swift
```diff
@@ -71,278 +100,301 @@ final class MasterContainerViewController: UIViewController {
...
 
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
-        self.didAppear = true
+        guard !didAppear else { return }
+        didAppear = true
+        if let splitViewController {
+            // Split view controller collapsed status when the app launches is correct here, so it's used to show/hide bottom sheet for the first appearance.
+            // The app may be launched in collapsed mode, if it was in such mode the last time it was moved to background.
+            showBottomSheet(!splitViewController.isCollapsed)
```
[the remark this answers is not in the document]
mvasilak: The only change would happen if the app starts with regular horizontal size in iPad, and it's just an instantaneous hidden toggle, so nothing is noticable, even w/ slow animation, as far as I have tested.

Zotero/Scenes/Detail/Items/ViewModels/ItemsToolbarController.swift
```diff
@@ -78,19 +78,17 @@ final class ItemsToolbarController {
     }
 
     private func deviceSpecificFilters(from filters: [ItemsFilter]) -> [ItemsFilter] {
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/762/commits/1891ff8596c98bb69df7ebaede32d2f3a02fe396

Zotero/Scenes/Detail/Items/ViewModels/ItemsToolbarController.swift
```diff
@@ -78,19 +78,17 @@ final class ItemsToolbarController {
...
-        // There is different functionality on iPad and iPhone. iPhone shows tag filters in filter popup in items screen while iPad shows tag filters in master controller.
-        // So filter icon and description should always show up on iPhone, while it should not show up on iPad for tag filters.
-        // Therefore we ignore `.tag` filter on iPad and keep it on iPhone.
-        switch UIDevice.current.userInterfaceIdiom {
-        case .pad:
+        // There is different functionality based on horizontal size class. iPhone and compact iPad show tag filters in filter popup in items screen while iPad shows tag filters in master controller.
+        // So filter icon and description should always show up on iPhone and compact iPad, while it should not show up on regular iPad for tag filters.
+        // Therefore we ignore `.tag` filter on iPhone and compact iPad, and keep it on regular iPad.
+        if delegate?.traitCollection.horizontalSizeClass == .regular && UIDevice.current.userInterfaceIdiom == .pad {
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/762/commits/1891ff8596c98bb69df7ebaede32d2f3a02fe396

PR #765, opened 2023-09-12
Remove old style of search bar, replace with iOS default style

### LINKED ISSUE #764: Update items search bar setup
Modernize setup logic to adopt iOS default logic of presenting the search bar in the navigation bar.

Reference https://github.com/zotero/zotero-ios/pull/762#discussion_r1321516701

michalrentka:
Fixes #764


michalrentka:
@dstillman can we move to current iOS defaul style of search bar? The current implementation is a bit hacky and had some issues, which seem to have been fixed by OS over time (or at least I didn't see any reports for some time). Anyway the current default is very similar and also works a bit smoother.


IMG_0551
IMG_0550

michalrentka:
@dstillman do these icons in order look better or do we maybe want to try compressing search bar into search icon in portrait?

IMG_58BAE0DEE75D-1

dstillman:
That's better. If we can, is there any reason not to collapse to search icon in all modes? I guess bigger tap target, clearer differentiation between the two search icons with the placeholder text, Files doesn't do it…

dstillman:
Yeah, I just meant it seems a little silly to display the search box all the time, but not having two identical icons may be reason enough to keep it. So I guess until we need the room, we can leave it.

mvasilak (approved):

PR #767, opened 2023-09-18
iOS 17 state restoration crash fix

michalrentka:
The app crashed on iOS 17 during state restoration due to changes in #762. It crashed because https://github.com/zotero/zotero-ios/blob/27853fa4e30801f8a17de9ae3e27d986738f7c54/Zotero/Scenes/Main/Views/MainViewController.swift#L96 was called before https://github.com/zotero/zotero-ios/blob/27853fa4e30801f8a17de9ae3e27d986738f7c54/Zotero/Scenes/Master/MasterContainerViewController.swift#L104 was called and so the `bottomController` tried to add itself into `bottomContainer`, which didn't exist yet.

I fixed it by creating `bottomController` and not adding it as subview immediately, but I moved adding it as subview to `viewDidLoad` function together with other UI components.

I also noticed something I didn't notice during code review of #762 and that is that the distinction between `MasterCoordinator` and `MasterTopCoordinator` was not needed anymore, because the `MasterContainerViewController` took that role and connects "top" and "bottom" controllers together. The `MasterCoordinator` was almost empty anyway at this point, so I refactored a little and removed it.

Last fix is in https://github.com/zotero/zotero-ios/blob/27853fa4e30801f8a17de9ae3e27d986738f7c54/Zotero/Scenes/Master/MasterContainerViewController.swift#L320. Currently on state restoration, when you closed the PDF, the bottom tag picker would just pop up without animation, because it appered in `viewDidAppear` which is called too late. I moved it to `viewIsAppearing` which seems to have better effect.


Zotero/Scenes/Master/MasterContainerViewController.swift
```diff
@@ -103,9 +73,11 @@ final class MasterContainerViewController: UINavigationController {
         view.backgroundColor = .clear
         setupView()
         setupKeyboardObserving()
-        showBottomSheet(false)
+        setBottomSheet(hidden: true)
         
         func setupView() {
+            guard let bottomController = bottomController else { return }
```
mvasilak, line 79:
```suggestion
            guard let bottomController else { return }
```

mvasilak (approved):
`viewIsAppearing` works great and it's great that it is backwards compatible. One small note, it is called every time the view appears, and although in this case the actions would be noops after the first call, in some cases we may still need to guard with a Bool property, to truly run specific appearance logic only once.

mvasilak (changes requested):
In `MasterContainerViewController` since the setup of the bottom container happens only for iPad form, which is the best optimization, now in iPhone form `bottomYConstraint` & `bottomContainerBottomConstraint` may be nil, so they should become optional as well.

michalrentka:
> In `MasterContainerViewController` since the setup of the bottom container happens only for iPad form, which is the best optimization, now in iPhone form `bottomYConstraint` & `bottomContainerBottomConstraint` may be nil, so they should become optional as well.

Fixed in 6878aab49b0e52d136bdfe674fb98fa161d5671b, also there was another issue with keyboard apperance. When tag filter was in custom position and keyboard appeared, it just covered the filter if it was just too low. Fixed in 88f888f2b8e40a6b79723c5f6c0bd35ff1757a62

Zotero/Scenes/Master/MasterContainerViewController.swift
```diff
@@ -43,9 +43,9 @@ final class MasterContainerViewController: UINavigationController {
     lazy var bottomController: DraggableViewController? = {
         return coordinatorDelegate?.createBottomController()
     }()
-    private var bottomContainer: UIView?
-    private var bottomYConstraint: NSLayoutConstraint!
-    private var bottomContainerBottomConstraint: NSLayoutConstraint!
+    private weak var bottomContainer: UIView?
+    private weak var bottomYConstraint: NSLayoutConstraint?
+    private weak var bottomContainerBottomConstraint: NSLayoutConstraint?
```
mvasilak, line 55: Why make them weak? Just in case they are removed e.g. from the view controller's view, to be deallocated then? Is there a cyclic reference somwehere?

PR #777, opened 2023-10-03
Check for remote changes when quota limit is reached

### LINKED ISSUE #775: Quota error blocking data downloads?
https://forums.zotero.org/discussion/comment/444763/#Comment_444763

Similar to https://github.com/zotero/zotero-ios/issues/773

It says it's a non-fatal error, but I also don't see any attempt to check object versions for data syncing.

michalrentka:
If the user had only submissions/uploads queued and reached quota limit, sync controller wouldn't check for remote changes on 413 response.

Fixes #775

### DIFF
--- a/Zotero/Controllers/Sync/SyncController.swift
+++ b/Zotero/Controllers/Sync/SyncController.swift
@@ -987,7 +987,7 @@ final class SyncController: SynchronizationController {
         result.subscribe(on: self.workScheduler)
               .subscribe(onSuccess: { [weak self] toUpdate, toRemove in
                   guard let self = self else { return }
-                let actions = self.createGroupActions(updateIds: toUpdate, deleteGroups: toRemove, syncType: self.type)
+                  let actions = self.createGroupActions(updateIds: toUpdate, deleteGroups: toRemove, syncType: self.type)
                   self.accessQueue.async(flags: .barrier) { [weak self] in
                     self?.finishSyncGroupVersions(actions: actions, updateCount: toUpdate.count)
                   }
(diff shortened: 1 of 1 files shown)


Zotero/Controllers/Sync/SyncController.swift
```diff
@@ -2125,91 +2125,96 @@ final class SyncController: SynchronizationController {
...
+                default:
+                    return false
+                }
+            })
 
-        if !self.nonFatalErrors.contains(error) {
-            self.nonFatalErrors.append(error)
-        }
+            // Uploads failed, we need to check for remote changes for this library, in case there were no previous writes.
+            self.queue.insert(.createLibraryActions(.specific([libraryId]), .onlyDownloads), at: 0)
```
mvasilak, line 2160: Since there is already such an action in the queue for all libraries, with `automatic` options, why is this needed?

mvasilak (approved):

Zotero/Controllers/Sync/SyncController.swift
mvasilak: Got it. 
As I was checking a relevant log, there was this sequence:
```
Sync: createLibraryActions(Zotero.SyncController.Libraries.all, Zotero.SyncController.CreateLibraryActionsOptions.automatic)

Sync: createUploadActions(libraryId: Zotero.LibraryIdentifier.custom(Zotero.RCustomLibraryType.myLibrary), hadOtherWriteActions: false, canEditFiles: true)
```

before the single upload action failed, so I assumed that the first creation would have already enqueued any download actions, or would at least do so after the non fatal error (if the queue order was jumped).

PR #779, opened 2023-10-10
Improved API parsing logs

michalrentka:


mvasilak (commented):
Rather than having all error log messages verbosely use the same format 
`<response name> missing key \"<key name>\"`, maybe the extra argument could be `responseName: String` instead of `errorLogMessage: String`, and build the error message with that?

Even better, just add an extra convenience method
```
    static func parse<T>(key: String, from data: [String: Any], responseName: String) throws -> T {
        try Self.parse(key: key, from: data, errorLogMessage: "\(responseName) missing key \"\(key)\"")
    }
```
So we could easily differentiate the message if need be in the future?

michalrentka:
> Rather than having all error log messages verbosely use the same format `<response name> missing key \"<key name>\"`, maybe the extra argument could be `responseName: String` instead of `errorLogMessage: String`, and build the error message with that?
> 
> Even better, just add an extra convenience method
> 
> ```
>     static func parse<T>(key: String, from data: [String: Any], responseName: String) throws -> T {
>         try Self.parse(key: key, from: data, errorLogMessage: "\(responseName) missing key \"\(key)\"")
>     }
> ```
> 
> So we could easily differentiate the message if need be in the future?

I didn't really know what message I'd use initially so I just added a String and ended up using the same message everywhere :). Anyway I changed it to pass `Class.Type` so that you get autocomplete and avoid typos. If we want different message somewhere sometime in future we can overload the function. See https://github.com/zotero/zotero-ios/pull/779/commits/929316ddb0f9e0e6b4b638a43b9d430b92f6d7be

mvasilak (approved):

PR #780, opened 2023-10-13
Improvements for item creation, duplication and parent creation

### LINKED ISSUE #778: Attachment is deleted if Create Parent Item screen is canceled
https://forums.zotero.org/discussion/comment/445381/#Comment_445381

michalrentka:
Plus some bug fixes, fixes #778

### DIFF
--- a/Zotero/Controllers/Database/Requests/CreateItemFromDetailDbRequest.swift
+++ b/Zotero/Controllers/Database/Requests/CreateItemFromDetailDbRequest.swift
@@ -62,6 +62,7 @@ struct CreateItemFromDetailDbRequest: DbResponseRequest {
             guard let creator = self.data.creators[creatorId] else { continue }
 
             let rCreator = RCreator()
+            rCreator.uuid = UUID().uuidString
             rCreator.rawType = creator.type
             rCreator.firstName = creator.firstName
             rCreator.lastName = creator.lastName
--- a/Zotero/Controllers/Database/Requests/CreateItemFromDetailDbRequest.swift
+++ b/Zotero/Controllers/Database/Requests/CreateItemFromDetailDbRequest.swift
@@ -45,6 +45,7 @@ struct CreateItemFromDetailDbRequest: DbResponseRequest {
         item.dateAdded = self.data.dateAdded
         item.dateModified = self.data.dateModified
         item.libraryId = self.libraryId
+        item.changesSyncPaused = true
         database.add(item)
 
         var changes: RItemChanges = [.type, .fields]
@@ -105,6 +106,7 @@ struct CreateItemFromDetailDbRequest: DbResponseRequest {
                                                 collectionKey: nil,
                                                 parentKey: nil).process(in: database)
             rNote.parent = item
+            rNote.changesSyncPaused = true
             rNote.changes.append(RObjectChange.create(changes: RItemChanges.parent))
         }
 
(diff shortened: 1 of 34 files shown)


Zotero/Scenes/Detail/ItemDetail/Views/ItemDetailCollectionViewHandler.swift
```diff
@@ -387,10 +289,10 @@ final class ItemDetailCollectionViewHandler: NSObject {
     private func sections(for data: ItemDetailState.Data, isEditing: Bool) -> [Section] {
         if isEditing {
             if data.isAttachment {
-                return [.title, .type, .fields, .dates, .tags, .attachments]
+                return [.title, .type, .fields, .dates]
             } else {
                 // Each section is visible during editing, except dates section. Dates are filled automatically and the user can't change them manually.
```
mvasilak, line 294: `.dates` seem to be passed below, does this comment need to be updated?

Zotero/Scenes/Detail/ItemDetail/Views/ItemDetailFieldEditCell.swift
```diff
@@ -51,7 +62,9 @@ final class ItemDetailFieldEditCell: UICollectionViewListCell {
         }
 
         private func apply(configuration: ContentConfiguration) {
-            self.contentView.textChanged = configuration.textChanged
+            let disposable = self.contentView.textObservable.bind(to: configuration.textObservable)
+            _ = configuration.disposeBag.insert(disposable)
```
mvasilak, line 66: Could this lead to having more than one text observers, if the cell is reused?
Should the `disposeBag` be emptied when the cell is prepared for reuse?

mvasilak (approved):
Some minor changes and questions

Zotero/Controllers/Database/Requests/CreateItemFromDetailDbRequest.swift
```diff
@@ -61,6 +62,7 @@ struct CreateItemFromDetailDbRequest: DbResponseRequest {
             guard let creator = self.data.creators[creatorId] else { continue }
 
             let rCreator = RCreator()
+            rCreator.uuid = UUID().uuidString
```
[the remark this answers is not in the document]
michalrentka: It shouldn't really matter to be honest. Those `uuid`s are only used by UI to identify creators in current edit session. They don't sync to backend or matter otherwise at all. 

In this case we're duplicating existing item, so we're creating duplicates of `Creator`s too, which should probably have their own unique identifiers. `uuid` is not a primary key though, so even `creator.id` would still work.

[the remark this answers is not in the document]
mvasilak: I've seen that in other places in the code they are kept in sync, but yes, if UI is not affected, that's ok.

Zotero/Scenes/Detail/ItemDetail/Views/ItemDetailFieldEditCell.swift
michalrentka: I thought that since the dispose bag is stored in `ContentConfiguration` it should be enough, but disposing on `prepareForReuse` definitely won't hurt. See 662d34fb9f18d6dc52cda3f161fa3f554603b6a2

Zotero/Scenes/Detail/ItemDetail/Views/ItemDetailCollectionViewHandler.swift
michalrentka: Updated in 9c12b1a04cf149a362ee3557080404bfd9cffc3d

PR #782, opened 2023-10-16
WIP Support multiple open items

mvasilak:
Adds support for multiple open items.

- [x] Keep track of open items in order of most recently opened
- [x] Show in items navigation bar button that restores last item and shows count of currently open items
- [x] Save/restore state of open items
- [x] Add support for PDF items
- [x] Add support for note items
- [x] Show in presented item navigation bar button menu that allows to quickly switch to another open item 
- [ ] Allow user to change order of presented open items

In consideration
- [x] Support separate open items for multiple windows
- [ ] Opening an already opened item in another window, should switch to that

### DIFF
--- a/Zotero/Controllers/Database/Requests/ReadItemDbRequest.swift
+++ b/Zotero/Controllers/Database/Requests/ReadItemDbRequest.swift
@@ -25,18 +25,3 @@ struct ReadItemDbRequest: DbResponseRequest {
         return item
     }
 }
-
-struct ReadItemGloballyDbRequest: DbResponseRequest {
-    typealias Response = RItem
-
-    let key: String
-
-    var needsWrite: Bool { return false }
-
-    func process(in database: Realm) throws -> RItem {
-        guard let item = database.objects(RItem.self).filter(.key(key)).first else {
-            throw DbError.objectNotFound
-        }
-        return item
-    }
-}
--- a/Zotero/Controllers/Database/Requests/ReadItemDbRequest.swift
+++ b/Zotero/Controllers/Database/Requests/ReadItemDbRequest.swift
@@ -25,3 +25,18 @@ struct ReadItemDbRequest: DbResponseRequest {
         return item
     }
 }
+
+struct ReadItemGloballyDbRequest: DbResponseRequest {
+    typealias Response = RItem
+
+    let key: String
+
+    var needsWrite: Bool { return false }
+
+    func process(in database: Realm) throws -> RItem {
+        guard let item = database.objects(RItem.self).filter(.key(key)).first else {
+            throw DbError.objectNotFound
+        }
+        return item
+    }
+}
--- a/Zotero/Controllers/OpenItemsController.swift
+++ b/Zotero/Controllers/OpenItemsController.swift
@@ -17,8 +17,50 @@ protocol OpenItemsPresenter: AnyObject {
 
 final class OpenItemsController {
     // MARK: Types
-    enum Item: Hashable, Equatable {
-        case pdf(library: Library, key: String)
+    enum Item: Hashable, Equatable, Codable {
+        case pdf(libraryId: LibraryIdentifier, key: String)
+        
+        // MARK: Types
+        enum ItemType: String, Codable {
+            case pdf
+        }
+        
+        // MARK: Properties
+        var type: ItemType {
+            switch self {
+            case .pdf:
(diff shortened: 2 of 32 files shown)


mvasilak:
@michalrentka this is ready for a first look. Things to keep in mind:
- I tried not to upset existing architecture a lot, hence at the time being I refrained from using something like child view controllers to switch between open items.
- Instant presentation solution using screenshot window might be a bit extreme. I'll investigate if it can be done in a more elegant way, e.g. using custom transitions.
- In practice cold switch between items seems adequate, but we may need to consider hot switch of view controllers kept in memory, if there are UX issues.

Next steps could be keeping separate open items for separate windows, support for notes or other items, customization of the menu, or new UI (e.g. tabs) that makes use of this facility.

Zotero/Scenes/AppCoordinator.swift
```diff
@@ -33,33 +33,33 @@ protocol AppLoginCoordinatorDelegate: AnyObject {
 
 final class AppCoordinator: NSObject {
     private typealias Action = (UIViewController) -> Void
-
+
```
michalrentka, line 36: Careful about these empty lines with tabs, I guess you have it enabled in Xcode so if possible please disable it.

Zotero/Controllers/OpenItemsController.swift
```diff
@@ -0,0 +1,235 @@
...
+            }
+            var actions: [UIAction] = []
+            let items = self.items
+            let openItem: Item? = disableOpenItem ? items.last : nil
+            var itemTuples: [(Item, RItem)] = []
+            do {
+                try dbStorage.perform(on: .main) { coordinator in
+                    for item in items {
+                        do {
+                            let rItem = try coordinator.perform(request: ReadItemGloballyDbRequest(key: item.key))
```
michalrentka, line 164: The `Item` enum holds `LibraryIdentifier` too, so why is there a need for `ReadItemGloballyDbRequest`?

michalrentka:
Looks good to me so far 👍. Just one comment regarding the `ReadItemGloballyDbRequest` which I think is not necessary.

Zotero/Controllers/Database/Requests/ReadItemDbRequest.swift
```diff
@@ -25,3 +25,18 @@ struct ReadItemDbRequest: DbResponseRequest {
...
+
+struct ReadItemGloballyDbRequest: DbResponseRequest {
+    typealias Response = RItem
+
+    let key: String
+
+    var needsWrite: Bool { return false }
+
+    func process(in database: Realm) throws -> RItem {
+        guard let item = database.objects(RItem.self).filter(.key(key)).first else {
```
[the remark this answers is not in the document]
mvasilak: Removed in https://github.com/zotero/zotero-ios/pull/782/commits/114f136cd6bffe050c751ba132ae2a30d7511e0c

Zotero/Controllers/OpenItemsController.swift
mvasilak: Fixed in https://github.com/zotero/zotero-ios/pull/782/commits/114f136cd6bffe050c751ba132ae2a30d7511e0c to use proper request. In an interim version, based on my mistake of `key` being unique in the database, I had a single request that read all items at once, which I later modified to a loop, so that's how I got there 😅

mvasilak:
@michalrentka ready for another look. New additions:
* User order is maintained in the bar buttom menu (not editable at the time being, it is always the order of first time item is opened).
* Support for notes.

Zotero/Controllers/OpenItemsController.swift
```diff
@@ -0,0 +1,282 @@
...
+
+final class OpenItemsController {
+    // MARK: Types
+    struct Item: Hashable, Equatable, Codable {
+        enum Kind: Hashable, Equatable, Codable {
+            case pdf(libraryId: LibraryIdentifier, key: String)
+            case note(libraryId: LibraryIdentifier, key: String)
+
+            // MARK: Types
+            enum `Type`: String, Codable {
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/782/commits/69211f1ada34a1155765fe270e634c34210fad5e

mvasilak:
Due to many changes when rebasing, this PR is closed to use a new branch.

PR #784, opened 2023-10-19
Improve annotation page label

### LINKED ISSUE #781: Limit page label length
https://forums.zotero.org/discussion/comment/446026/#Comment_446026

App currently doesn't limit the input length when editing page labels (macOS app limits that to 16 characters, although API upper limit is larger).

This leads to these issues:
- Annotation header buttons are pushed of view.
Screenshot 2023-10-16 at 15 39 18

- Sync fails due to page label being too long.
Screenshot 2023-10-16 at 15 39 40

- [x] Limit page label to specified length, e.g. 16 characters. 
- [x] Fix annotation header UI, so that the page label truncates when too large, and buttons don't disappear.
- [x] Migrate existing page labels by truncating to specified length.

mvasilak:
Closes #781

The same annotation page label max length as in the macOS client is used, but since the API does allow a larger limit, we may want to revisit that.

### DIFF
--- a/Zotero/Models/Database/Database.swift
+++ b/Zotero/Models/Database/Database.swift
@@ -13,7 +13,7 @@ import RealmSwift
 import Network
 
 struct Database {
-    private static let schemaVersion: UInt64 = 39
+    private static let schemaVersion: UInt64 = 40
 
     static func mainConfiguration(url: URL, fileStorage: FileStorage) -> Realm.Configuration {
         var config = Realm.Configuration(fileURL: url,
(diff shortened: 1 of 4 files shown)


Zotero/Models/Database/Database.swift
```diff
@@ -50,9 +50,26 @@ struct Database {
...
+        migration.enumerateObjects(ofType: RItem.className()) { oldObject, newObject in
+            let maxLength = 16
+            guard oldObject?["rawType"] as? String == ItemTypes.annotation,
+                  var fields = oldObject?["fields"] as? List<MigrationObject>,
+                  let index = fields.firstIndex(where: { $0["key"] as? String == FieldKeys.Item.Annotation.pageLabel }),
+                  let pageLabel = fields[index]["value"] as? String,
+                  pageLabel.count > maxLength
+            else { return }
+            (newObject?["fields"] as? List<MigrationObject>)?[index]["value"] = pageLabel.prefix(maxLength)
+            (newObject?["fields"] as? List<MigrationObject>)?[index]["changed"] = true
```
michalrentka, line 69: If we want this to sync, you need to add a `RObjectChange` with value of `RItemChanges.fields` for this `RItem` too.

michalrentka (changes requested):

Zotero/Models/Database/Database.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/784/commits/3649765e8a45246a2910e20b07f99fc6c62613e4

michalrentka (approved):

PR #791, opened 2023-10-25
Improve master container bottom additional safe area inset

mvasilak:
https://forums.zotero.org/discussion/108739/ios-tag-pane-prevents-scrolling-collections/

Adjusts additional safe area insets for interested view controllers, so they can always scroll to bottom of their content, regardless of the bottom sheet height.

### DIFF
--- a/Zotero/Scenes/Master/MasterContainerViewController.swift
+++ b/Zotero/Scenes/Master/MasterContainerViewController.swift
@@ -16,6 +16,17 @@ protocol DraggableViewController: UIViewController {
     func disablePanning()
 }
 
+protocol BottomSheetObserver: UIViewController {
+    func bottomSheetUpdated(hidden: Bool, containerHeight: CGFloat, topOffset: CGFloat)
+}
+
+extension BottomSheetObserver {
+    func bottomSheetUpdated(hidden: Bool, containerHeight: CGFloat, topOffset: CGFloat) {
+        let bottomInset: CGFloat = hidden ? .zero : (containerHeight - topOffset)
+        additionalSafeAreaInsets = UIEdgeInsets(top: 0, left: 0, bottom: bottomInset, right: 0)
+    }
+}
+
 final class MasterContainerViewController: UINavigationController {
     enum BottomPosition {
         case mostlyVisible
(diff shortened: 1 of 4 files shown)


Zotero/Scenes/Master/MasterContainerViewController.swift
```diff
@@ -16,37 +16,48 @@ protocol DraggableViewController: UIViewController {
...
+}
+
 final class MasterContainerViewController: UINavigationController {
     enum BottomPosition {
         case mostlyVisible
         case `default`
         case hidden
         case custom(CGFloat)
-
+
```
michalrentka, line 36: Lots of tabs added again

Zotero/Scenes/Master/MasterContainerViewController.swift
```diff
@@ -304,34 +316,38 @@ final class MasterContainerViewController: UINavigationController {
...
-        let visibleHeight = size.height - keyboardHeight
-        set(bottomPosition: bottomPosition, containerHeight: visibleHeight)
-
+        
+        set(bottomPosition: bottomPosition, containerHeight: size.height, keyboardHeight: keyboardHeight)
+        
         coordinator.animate(alongsideTransition: { _ in
             self.view.layoutIfNeeded()
         }, completion: nil)
+        coordinator.animate { _ in
```
michalrentka, line 339: Why does the coordinator animate view layout changes twice? Wasn't there supposed to be just `completion` block added to existing `coordinator.animate(alongsideTransition: ...`?

michalrentka (changes requested):

Zotero/Scenes/Master/MasterContainerViewController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/791/commits/240b793795fe6c73a9740e9a38c0a2278bc72044, still some spaces were removed now, instead of added (a lot less though).

Zotero/Scenes/Master/MasterContainerViewController.swift
mvasilak: Didn't delete the old code 😅 Good catch. Check https://github.com/zotero/zotero-ios/pull/791/commits/22bf5330bce3f553ed23f6183aca53263ccca839

michalrentka (approved):

PR #795, opened 2023-10-27
Added support for new page index formats

### LINKED ISSUE #794: Support new page index formats
https://forums.zotero.org/discussion/comment/446971/#Comment_446971

michalrentka:
Fixes #794


mvasilak (approved):
Looks good. 

Just a couple of questions.
In the decimal case, are we guaranteed that there won't ever be more than 3 decimal points by the API? 
Is there an edge case that the page index is a string of an integer, so converting it we might lose its actual type?

michalrentka:
> In the decimal case, are we guaranteed that there won't ever be more than 3 decimal points by the API? 

We round rect values for annotation position to 3 decimal points, so I applied the same logic, I don't think more precision will be needed, but I'll ask to be sure.

> Is there an edge case that the page index is a string of an integer, so converting it we might lose its actual
 type?

I don't think so, string page indices will probably contain some CSS values like `#content > div > div:first-child > div:nth-child(3)`, but I'll confirm to be sure.

PR #796, opened 2023-10-27
Hardcoded most used extension and mimetypes as possible crash fix

michalrentka:
There are numerous crashes with no real info in them, but they all seem to be processing translators in the background and the crash points to `mimeTypeFromExtension` being called, even though it's not specifically mentioned that it crashes there. We encountered similar issue with `extensionFromMimeType` a long time ago, so it might be worth a try to hard code most used values.


mvasilak (approved):
An interesting improvement in the future would be to have a bidirectional map, that could also be initialized from a configuration file. But for the time being 👌

PR #807, opened 2023-11-14
Improve annotations filter

mvasilak:
* Fixes issue reported in https://forums.zotero.org/discussion/109348/ios-filtering-annotations-by-color-should-hide-comment-icon/p1
* Fixes annotations filter UI layout issues


michalrentka (approved):
Looks good, but if you could extract the color picker into separate view and just reuse the view in `AnnotationsFilterViewController.swift` and `AnnotationToolOptionsViewController.swift` instead of having the same code in both, that would be ideal.

mvasilak:
@michalrentka ready for another look. Created common widget that is used in all current use cases, and also fixed some related small UI bugs. In the future, if necessary, we could improve it by implementing the widget e.g. using a collection view, so layout would be more dynamic.

Zotero/Scenes/Detail/Annotation Filter Popover/Views/AnnotationsFilterViewController.swift
```diff
@@ -11,165 +11,219 @@ import UIKit
...
+            updateFilter()
+            updatePreferredContentSize()
         }
-    }
 
-    private func set(tags: Set<String>, availableTags: [Tag]) {
-        if availableTags.isEmpty {
-            self.tagsContainer.isHidden = true
-            return
+        func setSelected(colors: Set<String>) {
```
michalrentka, line 171: At this point you can remove this function and just call `colorPicker?.setSelected(hexColors:)` directly.

Zotero/Scenes/Detail/Annotation Popover/Views/ColorPickerCell.swift
```diff
@@ -11,42 +11,66 @@ import UIKit
 import RxSwift
 
 final class ColorPickerCell: UITableViewCell {
-    @IBOutlet private weak var stackView: UIStackView!
+    let colorChange: PublishSubject<String>
+    let disposeBag: DisposeBag
```
michalrentka, line 15: Doesn't this `disposeBag` need to be disposed on cell reuse? I guess there's always only one used, but just to be sure in case we use this somewhere else.

michalrentka (approved):

Zotero/Scenes/Detail/Annotation Popover/Views/ColorPickerCell.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/807/commits/07b7b802048eca7e64129908c684f0be3e3a0300

I see that there is the `RxTableViewCell` class. Perhaps the same logic should apply there, that is upon reuse replace the `disposeBag` (and therefore dispose the old as it is deinitialized), rather than expecting `newDisposeBag` to be called explictly?

Zotero/Scenes/Detail/Annotation Filter Popover/Views/AnnotationsFilterViewController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/807/commits/2fae4dd91e37662605dfed6ee1237574ead4ea68

Zotero/Scenes/Detail/Annotation Popover/Views/ColorPickerCell.swift
michalrentka: @mvasilak sure, I think the same should apply there. You can try changing it and test whether everything works as expected.

mvasilak: Check https://github.com/zotero/zotero-ios/pull/807/commits/70f5ff7c73a63499f9093c63eb82caf098f75da6
Even if there is only one cell, it might be reused if it is not visible due to scrolling, and then is setup again. So it's indeed better to properly prepare for reuse.

michalrentka (approved):

Issue #308, opened 2021-06-18
Embedded newlines in highlight text


dstillman:
It seems like we're including newlines in extracted text on iOS, which doesn't happen on desktop. @mrtcode can comment on our approach there — not sure if we get that for free from pdf.js or we're doing something special there.

![IMG_0035](https://user-images.githubusercontent.com/783082/122610427-9ffab900-d04d-11eb-84de-8615e5b38be9.PNG)

mrtcode:
Currently we don't use any PDF.js logic to extract text because it's inferior. Instead, we only pull characters data from PDF.js that includes position, font, etc. and process it with another logic that is ported from Xpdf `pdftotext` which allows to get words from characters. There is a plan to port the logic that makes lines from words and paragraphs from lines. For lines extraction we are currently using our own logic.

Speaking about new lines, we're currently just checking if the last character in the line [has dash](https://github.com/zotero/pdf-reader/blob/6a126b1cd0a67c0c25565bed2dd2621c6e29489f/src/lib/text/range.js#L4) or not, and removing it if so. I don't remember if Xpdf is doing anything more intelligent here or not, but I'll investigate when I'll get back to this. Of course the result is not always correct, especially for scanned papers.

Also, I would expect PSPDFKit to be more intelligent here, maybe it allows to configure text extraction?

michalrentka:
@dstillman can you share the pdf?

dstillman:
Shared (though I would imagine it happens for any PDF with columns)

michalrentka:
So the PSPDFKit implementation just adds a \r\n to the end of each line. I could remove all newline characters but then we'd loose actual new lines. There's always just single \r\n even if there is a bigger gap.

dstillman:
Probably better to strip all newlines than to have extra newlines in all highlights. I think highlights will mostly not have real newlines in them.

michalrentka:
So i tested this a bit. The best results I observed when I replaced newlines with whitespace. I don't replace newlines if there's a "." before them. That way we get to keep newlines between paragraphs.

Before:
![before](https://user-images.githubusercontent.com/1022956/162739354-a7a1658d-bb5b-4fb0-8c6c-13e8f2fbd29c.jpeg)
After:
![after](https://user-images.githubusercontent.com/1022956/162739386-5604c1f5-d9e2-4457-ab94-3d38774c72f1.jpeg)

PR #808, opened 2023-11-15
Improve text conversion for copying and highlighting from PDF documents

mvasilak:
* Improves copied text from PDF documents by glueing hyphenated words in new lines and removing new lines.
* Improves annotation text by by glueing hyphenated before certain new lines are removed.

Improvement mentioned in this [forum discussion](https://forums.zotero.org/discussion/109222/ios-pdf-reader-improvement-request)
With some info from from https://github.com/zotero/zotero-ios/issues/308

~- [ ] Also add database migration for annotation text if needed.~

### DIFF
--- a/Zotero/Controllers/TextConverter.swift
+++ b/Zotero/Controllers/TextConverter.swift
@@ -1,0 +1,78 @@
+//
+//  TextConverter.swift
+//  Zotero
+//
+//  Created by Miltiadis Vasilakis on 15/11/23.
+//  Copyright © 2023 Corporation for Digital Scholarship. All rights reserved.
+//
+
+import Foundation
+
+import CocoaLumberjackSwift
+
+struct TextConverter {
+    private static var hyphenAndNewlineExpression: NSRegularExpression? = {
+        do {
+            let pattern = #"[\x2D\u058A\u05BE\u1400\u1806\u2010-\u2015\u2E17\u2E1A\u2E3A\u2E3B\u301C\u3030\u30A0\uFE31\uFE32\uFE58\uFE63\uFF0D][\n\r]+"#
+            return try NSRegularExpression(pattern: pattern)
+        } catch let error {
+            DDLogError("CopiedTextConverter: can't create hyphen and newline expression - \(error)")
+            return nil
+        }
+    }()
+
+    private static var newlineExpression: NSRegularExpression? = {
+        do {
+            let pattern = #"[\n\r]+"#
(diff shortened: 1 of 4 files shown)


Zotero/Controllers/TextConverter.swift
```diff
@@ -0,0 +1,78 @@
...
+        }
+        return newString
+    }
+
+    static func removeHyphenAndNewlinePairs(from string: String) -> String {
+        guard let expression = hyphenAndNewlineExpression else { return string }
+        return replaceMatches(of: expression, with: "", in: string)
+    }
+
+    static func replaceNewlineOptionallySurroundedByWhitespaceExcludingEndingSentencesExpressionWithSingleSpace(from string: String) -> String {
```
michalrentka, line 59: Probably one of the longest names I've seen in Swift

michalrentka (approved):

michalrentka:
I probably wouldn't do db migration just so that all of your highlight annotations don't suddenly change text (even if it's for the better) values. @dstillman do you agree?

dstillman:
How would migration even work? We can't actually know what annotation text was created by the iOS app vs. the desktop app, right? It might be one thing to automatically update annotation text when a PDF was opened if we knew that the previous text was created on iOS, but otherwise I don't think it's really an option.

mvasilak:
@dstillman the transformation would just affect annotations created on iOS, where it would glue hyphened words separated by newlines, and then would remove any extraneous newlines (it keeps those newlines after periods, as they might be paragraphs). Since the desktop app already does a superset of these actions, text of highlight annotations created with it, won't change.

Of course, it still might not worth doing.

dstillman:
You're saying it wouldn't reprocess the text from the PDF but rather adjust a specific pattern (hyphen plus newline) in the stored annotation text itself?

mvasilak:
Correct, the migration would act on the stored highlight annotation text. However, that means that it would act only on those synced currently. And if the user is still using an older client in another device, it might still create new ones with the old transformer. Hmmm, maybe it makes more sense to only update those documents that are opened, and keep track of this per document? Or do nothing as @michalrentka suggested?

mvasilak:
Let's do nothing then for the time being, and we can see if we get relevant user feedback. To be honest, the best "migration" in this case might be to let users trigger such a transformation centrally in the backend database, not the local iOS database.

mvasilak:
@dstillman you agree with doing nothing for the time being migration-wise ☝️

dstillman:
Sure.

PR #809, opened 2023-11-17
Text and Underline annotation support

### LINKED ISSUE #712: Support underline and text annotations

michalrentka:
Closes #712


Zotero/Scenes/Detail/PDF/Views/CustomFreeTextAnnotationView.swift
```diff
@@ -0,0 +1,186 @@
...
+                    guard let sizePicker else { return }
+                    sizePicker.value = size
+                })
+            })
+            .disposed(by: self.disposeBag)
+
+        let colorButton = UIButton()
+        let deleteButton = UIButton()
+
+        if #available(iOS 15.0, *) {
```
mvasilak, line 80: Latest minimum deployment should make this always available.

Zotero/Scenes/Detail/PDF/Views/CustomFreeTextAnnotationView.swift
```diff
@@ -0,0 +1,186 @@
...
+    private func attributedString(from tags: [Tag]) -> AttributedString {
+        if tags.isEmpty {
+            var string = AttributedString(L10n.Pdf.AnnotationsSidebar.addTags)
+            string.foregroundColor = Asset.Colors.zoteroBlueWithDarkMode.color
+            return string
+        } else {
+            let nsAttributedString = AttributedTagStringGenerator.attributedString(from: tags, limit: 3)
+            return (try? AttributedString(nsAttributedString, including: \.uiKit)) ?? AttributedString()
+        }
+    }
```
mvasilak, line 181: Is this used?

mvasilak (commented):
Code looks good so far 👌 

 Interesting use of keyboard accessories, we could discuss if this UX is more appropriate for other annotation types as well. There is an issue of discoverability, I get those with double tap, is that correct? 

 FInally, a couple of times the text annotation drawing would zoom, without me changing anything about it. I only noticed after it happened, so I don’t have any hints as to what may have caused it. Dismissing and reopening the PDF, would restore what I expected.

Zotero/Scenes/Detail/PDF/PDFCoordinator.swift
```diff
@@ -334,6 +369,14 @@ extension PDFCoordinator: PdfReaderCoordinatorDelegate {
         controller.popoverPresentationController?.barButtonItem = sender
         self.navigationController?.present(controller, animated: true)
     }
+
+    func showFontSizePicker(sender: UIView, picked: @escaping (UInt) -> Void) {
+        let controller = FontSizePickerViewController(pickAction: picked)
+        controller.preferredContentSize = CGSize(width: 200, height: 400)
+        controller.modalPresentationStyle = UIDevice.current.userInterfaceIdiom == .pad ? .popover : .formSheet
```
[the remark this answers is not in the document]
michalrentka: Sure we can do this. I don't remember why, but we chose this way some time ago and we did it in other popovers as well, so it might be worth creating a ticket to revise all `modalPresentationStyle` instances and switch them all to popovers.

When testing on phone I noticed some usability issues as well, so I'll have to add navigation bar on phone and fix some constraints too.

michalrentka:
>  Interesting use of keyboard accessories, we could discuss if this UX is more appropriate for other annotation types as well.
Other annotations don't really use keyboard as their main way of editing though. For text it makes sense.

> There is an issue of discoverability, I get those with double tap, is that correct?
Yes. First tap is to select and the ability to move and scale text. Second tap is to confirm we actually want to edit text or its settings.

>  FInally, a couple of times the text annotation drawing would zoom, without me changing anything about it. I only noticed after it happened, so I don’t have any hints as to what may have caused it. Dismissing and reopening the PDF, would restore what I expected.

I'll try to reproduce, but it might be PSPDFKit issue.

Zotero/Scenes/Detail/PDF/Views/CustomFreeTextAnnotationView.swift
michalrentka: It's used above, where you made your previous comment. But now that we support iOS 15.6+, I can actually get rid of the other `func attributedString(from tags: [Tag]) -> NSAttributedString` and keep this one.

Zotero/Scenes/Detail/PDF/PDFCoordinator.swift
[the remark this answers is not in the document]
michalrentka: Ok so as discussed personally, there were some issues with popover and we're not using that here.

Zotero/Scenes/Detail/PDF/Views/CustomFreeTextAnnotationView.swift
michalrentka: Fixed in e5162c413a6ad621c4e1b0283d49c839b54a4e72

Zotero/Scenes/Detail/PDF/Views/CustomFreeTextAnnotationView.swift
michalrentka: Yeah so I thought I could use the `Configuration` API for tag button, but it doesn't support button having just single line of text. So I had to keep classic setup and I removed this as you suggested. e5162c413a6ad621c4e1b0283d49c839b54a4e72

michalrentka:
@mvasilak can you take another look at this? I updated and fixes some things, we'll be merging this with free text and underline annotations disabled in UI for the time being.

mvasilak (commented):
`AnnotationViewController.swift` is not part of the project, is it still in use?

 Icons for the new types are still missing, right?

michalrentka:
> Icons for the new types are still missing, right?

Yes, I created a ticket for it #871. They won't be visible for now though.

Zotero/Scenes/Detail/PDF/Models/PDFDatabaseAnnotation.swift
```diff
@@ -14,58 +14,59 @@ import RxSwift
...
 
-    var key: String {
-        return self.item.key
-    }
-
-    var _type: AnnotationType? {
-        guard let rawValue = self.item.fieldValue(for: FieldKeys.Item.Annotation.type) else {
-            DDLogError("DatabaseAnnotation: \(self.key) missing annotation type!")
+    init?(item: RItem) {
+        guard let _type = AnnotationType(rawValue: item.annotationType) else {
```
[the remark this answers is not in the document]
michalrentka: Check acc54bc27306ae68de7d7e7ac975930c28636a11. Also removed `AnnotationViewController.swift`.

mvasilak (approved):

PR #812, opened 2023-11-27
Emojis in tags improvements

### LINKED ISSUE #797: Changes to emoji tags
- Emoji tags should show up in items list - they should show up first, then overlapped colored circles.
- If tag contains text and emojis, strip text and pick first emoji group and show that in item list. Show whole tag in other places.
- Emoji tags don't have to be colored to show up as emojis
- Tags in tag picker should sort in this order: colored, emojis, other.

https://github.com/zotero/zotero/pull/3330/files#diff-6d32b0ceca7a10685e44094f16b83bb7f59b8db13ee9b3b388cd7ad24b7f10f5R226-R241
https://github.com/zotero/zotero/issues/3324#issuecomment-1679683989

michalrentka:
Closes #797

### DIFF
--- a/Zotero/Controllers/Database/Requests/CreateNoteDbRequest.swift
+++ b/Zotero/Controllers/Database/Requests/CreateNoteDbRequest.swift
@@ -81,12 +81,7 @@ struct CreateNoteDbRequest: DbResponseRequest {
             if let existing = allTags.filter(.name(tag.name)).first {
                 rTag = existing
             } else {
-                rTag = RTag()
-                rTag.name = tag.name
-                rTag.emojiGroup = EmojiExtractor.extractFirstContiguousGroup(from: tag.name)
-                rTag.updateSortName()
-                rTag.color = tag.color
-                rTag.libraryId = self.libraryId
+                rTag = .create(name: tag.name, color: tag.color, libraryId: libraryId)
                 database.add(rTag)
             }
 
--- a/Zotero/Controllers/Database/Requests/CreateNoteDbRequest.swift
+++ b/Zotero/Controllers/Database/Requests/CreateNoteDbRequest.swift
@@ -83,6 +83,7 @@ struct CreateNoteDbRequest: DbResponseRequest {
             } else {
                 rTag = RTag()
                 rTag.name = tag.name
+                rTag.emojiGroup = EmojiExtractor.extractFirstContiguousGroup(from: tag.name)
                 rTag.updateSortName()
                 rTag.color = tag.color
                 rTag.libraryId = self.libraryId
--- a/Zotero/Controllers/EmojiExtractor.swift
+++ b/Zotero/Controllers/EmojiExtractor.swift
@@ -14,7 +14,7 @@ struct EmojiExtractor {
         var endIndex: Int?
 
         for (idx, character) in text.enumerated() {
-            let isEmoji = character.unicodeScalars.first?.properties.isEmojiPresentation ?? false
+            let isEmoji = isEmoji(character: character)
             if startIndex == nil && isEmoji {
                 startIndex = idx
             }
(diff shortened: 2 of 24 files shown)


Zotero/Controllers/EmojiExtractor.swift
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    private static func isEmoji(character: Character) -> Bool {
+        guard let firstScalar = character.unicodeScalars.first else { return false }
+
+        if character.unicodeScalars.count > 1 {
+            return firstScalar.properties.isEmoji
+        }
+
+        return firstScalar.properties.isEmoji && firstScalar.value > 0x238C
```
mvasilak, line 42: Is the lower limit arbitrary?

Zotero/Scenes/General/ViewModels/TagPickerActionHandler.swift
```diff
@@ -92,9 +92,10 @@ struct TagPickerActionHandler: ViewModelActionHandler {
         do {
             let request = ReadTagPickerTagsDbRequest(libraryId: viewModel.state.libraryId)
             let results = try self.dbStorage.perform(request: request, on: .main)
-            let colored = results.filter("color != \"\"").sorted(byKeyPath: "name")
-            let others = results.filter("color = \"\"").sorted(byKeyPath: "name")
-            let tags = Array(colored.map(Tag.init)) + Array(others.map(Tag.init))
+            let colored = results.filter("color != \"\"").sorted(byKeyPath: "order")
```
mvasilak, line 95: Why did the sorting changed here?

Zotero/Models/Tag.swift
```diff
@@ -11,25 +11,29 @@ import UIKit
 struct Tag: Identifiable, Equatable, Hashable {
     let name: String
     let color: String
+    let emojiGroup: String?
     let type: RTypedTag.Kind
 
     var id: String { return self.name }
 
-    init(name: String, color: String) {
+    init(name: String, color: String, emojiGroup: String?) {
```
mvasilak, line 19: A convenience `init` could be added that passes then `EmojiExtractor.extractFirstContiguousGroup(from: name)` to the regular `init`.

mvasilak (approved):
Some questions and minimal changes proposed.

Zotero/Scenes/General/ViewModels/TagPickerActionHandler.swift
michalrentka: I matched desktop behavior. I didn't notice the difference before.

Zotero/Controllers/EmojiExtractor.swift
michalrentka: It should be the first emoji UTF16 value. I can't use only the `isEmoji`, because it also detects some other symbols (#, *, 6) as emojis, so this check avoids incorrect detections. (I've got this from this discussion: https://stackoverflow.com/questions/30757193/find-out-if-character-in-string-is-emoji)

Zotero/Models/Tag.swift
michalrentka: Actually I'll just get rid of `emojiGroup` and include the `EmojiExtractor` in initializer. This initializer is used in one place only anyway.

Zotero/Controllers/Database/Requests/CreateNoteDbRequest.swift
```diff
@@ -83,6 +83,7 @@ struct CreateNoteDbRequest: DbResponseRequest {
             } else {
                 rTag = RTag()
                 rTag.name = tag.name
+                rTag.emojiGroup = EmojiExtractor.extractFirstContiguousGroup(from: tag.name)
```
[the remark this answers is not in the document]
michalrentka: Good idea 👍
9c9f0ddee18257234209941ba5e4b02b454a5da7

Zotero/Controllers/EmojiExtractor.swift
michalrentka: You're right, that lower bound seems better, updated 41a8c98636ebd233ef491a27db3181cee453107b

PR #815, opened 2023-12-07
Page thumbnails in PDF reader sidebar

### LINKED ISSUE #742: Page thumbnails view
https://forums.zotero.org/discussion/106681/comment-on-zotero-beta-7-feature-request

As in the desktop app

michalrentka:
Closes #742

### DIFF
--- a/Zotero/Scenes/Detail/PDF/Views/PDFSidebarViewController.swift
+++ b/Zotero/Scenes/Detail/PDF/Views/PDFSidebarViewController.swift
@@ -44,7 +44,7 @@ class PDFSidebarViewController: UIViewController {
     private let disposeBag: DisposeBag
 
     private weak var picker: UISegmentedControl!
-    private weak var thumbnailsController: PdfThumbnailsViewController!
+    private weak var thumbnailsController: PDFThumbnailsViewController!
     private weak var annotationsController: AnnotationsViewController!
     private weak var outlineController: TableOfContentsViewController!
     weak var parentDelegate: (PDFReaderContainerDelegate & SidebarDelegate)?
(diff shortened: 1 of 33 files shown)


Zotero/Scenes/Detail/PDF/Views/PdfThumbnailsViewController.swift
```diff
@@ -0,0 +1,133 @@
...
+        if state.changes.contains(.userInterface) {
+            var snapshot = dataSource.snapshot()
+            snapshot.reloadSections([0])
+            dataSource.apply(snapshot)
+        }
+
+        if state.changes.contains(.scrollToSelection) {
+            let indexPath = IndexPath(row: state.selectedPageIndex, section: 0)
+            collectionView.selectItem(at: indexPath, animated: true, scrollPosition: .centeredVertically)
+        }
```
mvasilak, line 118: I got a crash when restoring a document at launch, but could be a race issue in general.

```suggestion
        // The following updates should be ignored if the collection hasn't loaded yet for the first time.
        var snapshot = dataSource.snapshot()
        guard snapshot.numberOfSections > 0 else { return }

        if state.changes.contains(.userInterface) {
            snapshot.reloadSections([0])
            dataSource.apply(snapshot)
        }

        if state.changes.contains(.scrollToSelection) {
            let indexPath = IndexPath(row: state.selectedPageIndex, section: 0)
            collectionView.selectItem(at: indexPath, animated: true, scrollPosition: .centeredVertically)
        }
```

Zotero/Scenes/Detail/PDF/Views/PDFSidebarViewController.swift
```diff
@@ -56,91 +64,132 @@ class PDFSidebarViewController: UIViewController {
...
         ])
     }
 
     private func setupControllers() {
-        let annotationsController = AnnotationsViewController(viewModel: self.viewModel)
-        annotationsController.parentDelegate = self.parentDelegate
-        annotationsController.coordinatorDelegate = self.coordinatorDelegate
-        annotationsController.boundingBoxConverter = self.boundingBoxConverter
-        self.addChild(annotationsController)
+        let annotationsController = AnnotationsViewController(viewModel: viewModel)
```
mvasilak, line 147: Since we create several controllers when the document appears, that may not even be used at all, maybe we can convert them to lazy load, in a future update. Or maybe just pre-load the first, or some.

michalrentka: I considered this to be pretty much the same as `UITabbarController` which allocates all controllers ahead too. But we can definitely improve this by lazy loading.

michalrentka:
> Is there a reason the new files are named e.g. `PdfThumbnailsViewController` instead of `PDFThumbnailsViewController`?

Just my memory I'm afraid :), I'll rename to uppercase.

Zotero/Scenes/Detail/PDF/Views/PdfThumbnailsViewController.swift
mvasilak: Yes, indeed that was what caused the crash, as it caused an update.

mvasilak (approved):

PR #817, opened 2023-12-08
Fixed firefox share extension

michalrentka:
Firefox sends multiple `NSItemProviders` and we always took the first. In case of Firefox, the first one was plaintext and second one was web. Web contains more info than plaintext, so I order them based on priority so that we always get most useful info for translation.

### DIFF
--- a/ZShare/ViewModels/ExtensionViewModel.swift
+++ b/ZShare/ViewModels/ExtensionViewModel.swift
@@ -121,7 +121,7 @@ final class ExtensionViewModel {
                     if error.isFatal {
                         return false
                     }
-                    
+
                     switch error {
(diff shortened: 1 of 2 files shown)


ZShare/ViewModels/ExtensionViewModel.swift
```diff
@@ -356,25 +356,48 @@ final class ExtensionViewModel {
...
-                            case .failure(let error): return Single.error(error)
-                            }
-                        }
+                case .failure:
+                    return false
+                }
+            })
+            .first()
+            .flatMap { value in
+                guard let value = value else {
```
mvasilak, line 389:
```suggestion
                guard let value else {
```

PR #821, opened 2023-12-13
Fix launch with document

mvasilak:
When launching the app with a restored document, the presented view controller that is also used for the initial screenshot wouldn't have its `viewDidAppear` and `viewDidDisappear` method called by the system, resulting in an interim state of the transition windows' subviews. Thus, if a popover was presented, and the app changes size, an empty transition window would come to front, making the app unresponsive.

To overcome this, we use separate view controllers for screenshot and presentation, so that the former properly disappears, leaving the expected state for transition windows.

### DIFF
--- a/Zotero/Scenes/AppCoordinator.swift
+++ b/Zotero/Scenes/AppCoordinator.swift
@@ -382,7 +382,6 @@ final class AppCoordinator: NSObject {
 
         presentAction(oldController, {
             // New window is visible with a screenshot, return old view controller and present the new one
-            window.rootViewController = oldController
             self.presentedRestoredControllerWindow = nil
         })
     }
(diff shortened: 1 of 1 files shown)


Zotero/Scenes/AppCoordinator.swift
```diff
@@ -368,8 +382,12 @@ final class AppCoordinator: NSObject {
...
 
         // Create a temporary `UIWindow` which will be shown above current window until it successfully presents the new view controller.
-        let tmpWindow = UIWindow(frame: window.frame)
-        tmpWindow.windowScene = window.windowScene
+        let tmpWindow: UIWindow
+        if let windowScene = window.windowScene {
+            tmpWindow = UIWindow(windowScene: windowScene)
+        } else {
+            tmpWindow = UIWindow(frame: window.frame)
+        }
```
michalrentka, line 390: You could simplify this into one liner:
`let tmpWindow = window.windowScene.flatMap(UIWindow.init) ?? UIWindow(frame: window.frame)`

michalrentka (approved):
Good catch!

Zotero/Scenes/AppCoordinator.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/821/commits/93c1c0b001ce45d02cdd8da350908d2c0bff8ab1

PR #823, opened 2023-12-13
Fix document search source when navigation bar is hidden

mvasilak:
Fixes issue in https://forums.zotero.org/discussion/109983/ipados-17-2-zotero-beta-1-0-25-crash-when-searching-in-pdf-cmd-f

If the document screen is restored without a navigation bar, then the navigation item of the document view controller is not yet utilized by the navigaton controller. Hence, the bar button item used as the source for the search popover, is not part of the view hierarchy. Since search can be shown via external keyboard, with combination `command+F`, to be on the safe side, if the navigation bar is hidden, we use a source view and rect instead.

### DIFF
--- a/Zotero/Scenes/Detail/PDF/PDFCoordinator.swift
+++ b/Zotero/Scenes/Detail/PDF/PDFCoordinator.swift
@@ -188,7 +188,6 @@ extension PDFCoordinator: PdfReaderCoordinatorDelegate {
             }
 
             existing.overrideUserInterfaceStyle = userInterfaceStyle
-            existing.modalPresentationStyle = UIDevice.current.userInterfaceIdiom == .pad ? .popover : .formSheet
             setupPresentation(for: existing, with: sender)
             existing.text = text
 
@@ -213,7 +212,6 @@ extension PDFCoordinator: PdfReaderCoordinatorDelegate {
                     popoverPresentationController.barButtonItem = sender
                 }
                 popoverPresentationController.sourceView = nil
-                popoverPresentationController.sourceRect = .null
             } else {
                 if #available(iOS 17, *) {
                     popoverPresentationController.sourceItem = nil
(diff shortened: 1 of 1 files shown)


Zotero/Scenes/Detail/PDF/PDFCoordinator.swift
```diff
@@ -199,10 +199,31 @@ extension PDFCoordinator: PdfReaderCoordinatorDelegate {
...
+            pdfSearchController.modalPresentationStyle = UIDevice.current.userInterfaceIdiom == .pad ? .popover : .formSheet
+            guard let popoverPresentationController = pdfSearchController.popoverPresentationController else { return }
+            if navigationController?.isNavigationBarHidden == false {
+                if #available(iOS 17, *) {
+                    popoverPresentationController.sourceItem = sender
+                } else {
+                    popoverPresentationController.barButtonItem = sender
+                }
+                popoverPresentationController.sourceView = nil
+                popoverPresentationController.sourceRect = .null
```
michalrentka, line 216: Why use `.null` here and `.zero` below? Is there a difference for this case?

mvasilak: Yes, `.null` is the default value, which instructs the system to use the current frame of `sourceView`, that I think would put the popover outside of the visible screen.
`.zero` anchors it on the top left corner of the screen, which is what was already happening if the search popover was triggered by key press, when the navigation bar is hidden (but has been properly setup first).

mvasilak: On the other hand, we can probably always set it to `.zero`, since it will be ignored if `sourceView` is `nil` anyway.

michalrentka: Sure you can change it but this should be fine too.

mvasilak: Check https://github.com/zotero/zotero-ios/pull/823/commits/d9abe2729c398b5bdf3a87fb770858f76ed97616

michalrentka (approved):

PR #827, opened 2023-12-27
Update PDF document share menu

mvasilak:
* Converts PDF share menu from popover to UIMenu with separate actions for including Zotero annotations or not
* Adds copy citation and bibliography actions to the menu, where available (feature requested in this [forum post](https://forums.zotero.org/discussion/108280/ios-feature-request-add-copy-citation-and-bibliography-commands-to-share-menu))


Zotero/Scenes/Detail/CopyBibliography/Views/CopyBibliographyViewController.swift
```diff
@@ -0,0 +1,171 @@
...
+
+        case .error(let message):
+            overlayText.text = message
+            overlayActivityIndicator.stopAnimating()
+            overlayErrorIcon.isHidden = false
+        }
+
+        view.layoutIfNeeded()
+
+        UIView.animate(withDuration: 0.2) {
```
michalrentka, line 158: Maybe add a guard in case the overlay is already visible.

Zotero/Scenes/Detail/DetailCoordinator.swift
```diff
@@ -912,23 +924,60 @@ extension DetailCoordinator: DetailNoteEditorCoordinatorDelegate {
...
+            coordinator.start(animated: false)
+            presenter.present(containerController, animated: true)
+        }
+    }
+}
+
+extension DetailCoordinator: DetailCopyBibliographyCoordinatorDelegate {
+    func overlay(for viewController: CopyBibliographyViewController, isHidden: Bool) {
+        guard isHidden else { return }
+        viewController.willMove(toParent: nil)
```
michalrentka, line 978: I'd move this to `CopyBibliographyViewController`. The controller removing itself from parent controller is similar to `popViewController()` or `dismiss` which we call from the controller too. Also having the `isHidden` flag and ignoring half of it is a bit unnecessary.

Zotero/Scenes/Detail/PDF/PDFCoordinator.swift
```diff
@@ -341,19 +345,56 @@ extension PDFCoordinator: PdfReaderCoordinatorDelegate {
...
-                completed(settings)
-            })
-        })
-        let controller = UIHostingController(rootView: view)
-        controller.overrideUserInterfaceStyle = userInterfaceStyle
-        controller.preferredContentSize = CGSize(width: 400, height: 140)
-        controller.modalPresentationStyle = UIDevice.current.userInterfaceIdiom == .pad ? .popover : .formSheet
-        controller.popoverPresentationController?.barButtonItem = sender
-        self.navigationController?.present(controller, animated: true)
+    func showCitation(for itemId: String, libraryId: LibraryIdentifier) {
```
michalrentka, line 348: You don't have to re-implement this function. You can just call `(parentCoordinator as? DetailsCoordinator)?.showCitation(for: itemId, libraryId: libraryId)`

Zotero/Scenes/Detail/PDF/PDFCoordinator.swift
```diff
@@ -341,19 +345,56 @@ extension PDFCoordinator: PdfReaderCoordinatorDelegate {
...
+
+        let controller = SingleCitationViewController(viewModel: viewModel)
+        controller.coordinatorDelegate = self
+        let navigationController = UINavigationController(rootViewController: controller)
+        let containerController = ContainerViewController(rootViewController: navigationController)
+        self.navigationController?.present(containerController, animated: true, completion: nil)
+    }
+
+    func copyBibliography(for itemId: String, libraryId: LibraryIdentifier, showOverlayOn viewController: UIViewController) {
+        guard let citationController = controllers.userControllers?.citationController else { return }
```
michalrentka, line 371: Same as above.

Zotero/Scenes/Detail/PDF/PDFCoordinator.swift
```diff
@@ -542,3 +583,30 @@ extension PDFCoordinator: PdfAnnotationsCoordinatorDelegate {
...
+    func showMissingStyleError(using presenter: UINavigationController?) {
+        guard let coordinator = parentCoordinator as? DetailCoordinator else { return }
+        coordinator.showMissingStyleError(using: navigationController)
+    }
+}
+
+extension PDFCoordinator: DetailCopyBibliographyCoordinatorDelegate {
+    func overlay(for viewController: CopyBibliographyViewController, isHidden: Bool) {
+        guard isHidden else { return }
+        viewController.willMove(toParent: nil)
```
michalrentka, line 607: Same as before, no need to have this implemented in coordinator. I'd keep this in view controller.

Zotero/Scenes/Detail/PDF/Views/PDFReaderViewController.swift
```diff
@@ -91,19 +91,49 @@ class PDFReaderViewController: UIViewController {
...
+                elements.append(copyCitationAction)
+                let copyBibliographyAction = UIAction(title: L10n.Citation.copyBibliography, image: .init(systemName: "doc.on.doc")) { [weak self] _ in
+                    guard let self, let coordinatorDelegate else { return }
+                    coordinatorDelegate.copyBibliography(for: parentKey, libraryId: viewModel.state.library.identifier, showOverlayOn: self)
+                }
+                elements.append(copyBibliographyAction)
+            }
+
+            let exportAttributes: UIMenuElement.Attributes = viewModel.state.document.isLocked ? [.disabled] : []
+            let exportOriginalPDFAction = UIAction(title: L10n.Pdf.Export.export, attributes: exportAttributes) { [weak self] _ in
```
michalrentka, line 122: Can you add some icon here and below too?

Zotero/Scenes/Detail/PDF/Views/PDFReaderViewController.swift
```diff
@@ -91,19 +91,49 @@ class PDFReaderViewController: UIViewController {
...
+        let deferredMenu = UIDeferredMenuElement.uncached { [weak self] elementProvider in
+            var elements: [UIMenuElement] = []
+            defer {
+                elementProvider(elements)
+            }
+            guard let self else { return }
+
+            if let parentKey = viewModel.state.parentKey {
+                let copyCitationAction = UIAction(title: L10n.Citation.copyCitation, image: .init(systemName: "doc.on.doc")) { [weak self] _ in
+                    guard let self, let coordinatorDelegate else { return }
```
michalrentka, line 110: Hmm isn't the closure missing `[weak coordinatorDelegate]`? Or does this guard translate to `guard let self = self, let coordinatorDelegate = self.coordinatorDelegate else { return }` (that's why I hate these shortened versions)?

Same with `viewModel` below. Shouldn't be either be `[weak viewModel]` or `self.viewModel`?

Zotero/Scenes/Detail/SingleCitation/Views/SingleCitationViewController.swift
```diff
@@ -46,95 +46,144 @@ final class SingleCitationViewController: UIViewController {
...
-//                    self.previewTitleLabel.numberOfLines = 0
-//                    self.previewTitleLabel.attributedText = attr
-//                }
-//            } catch let error {
-//                NSLog("ERROR: \(error)")
-//            }
-            self.updatePreferredContentSize()
+        func setupNavigationBar() {
+            let cancel = UIBarButtonItem(title: L10n.cancel, style: .plain, target: nil, action: nil)
+            cancel.rx.tap.subscribe(onNext: { [weak self] in
```
michalrentka, line 83: You can change this to `subscribe(with: self, onNext: { self in`

Zotero/Scenes/Detail/SingleCitation/Views/SingleCitationViewController.swift
```diff
@@ -46,95 +46,144 @@ final class SingleCitationViewController: UIViewController {
...
+                .disposed(by: disposeBag)
+
+            omitAuthorSwitch.rx.controlEvent(.valueChanged)
+                .subscribe(onNext: { [weak self] _ in
+                    guard let self else { return }
+                    viewModel.process(action: .setOmitAuthor(omitAuthorSwitch.isOn))
+                })
+                .disposed(by: disposeBag)
+
+            func update(state: SingleCitationState) {
```
michalrentka, line 114: Same as above, I'd keep `update` outside of `viewDidLoad`. Doesn't make sense to have this here and then for example have a function `private func setupRightButtonItem(isLoading: Bool)` outside of here.

Zotero/Scenes/Detail/SingleCitation/Views/SingleCitationViewController.swift
```diff
@@ -46,95 +46,144 @@ final class SingleCitationViewController: UIViewController {
...
+        func setupObserving() {
+            viewModel.stateObservable
+                .subscribe(onNext: { state in
+                    update(state: state)
+                })
+                .disposed(by: disposeBag)
+
+            locatorTextField.rx.controlEvent(.editingChanged).flatMap({ Observable.just(self.locatorTextField.text ?? "") })
+                .debounce(.milliseconds(150), scheduler: MainScheduler.instance)
+                .subscribe(onNext: { [weak self] value in
```
michalrentka, line 102: `subscribe(with: self, onNext: { self, value in`

Zotero/Scenes/Detail/SingleCitation/Views/SingleCitationViewController.swift
```diff
@@ -46,95 +46,144 @@ final class SingleCitationViewController: UIViewController {
...
+                    self?.viewModel.process(action: .setLocatorValue(value))
+                })
+                .disposed(by: disposeBag)
+
+            omitAuthorSwitch.rx.controlEvent(.valueChanged)
+                .subscribe(onNext: { [weak self] _ in
```
michalrentka, line 108: `subscribe(with: self, onNext: { self, _ in`

Zotero/Scenes/Detail/SingleCitation/Views/SingleCitationViewController.swift
```diff
@@ -46,95 +46,144 @@ final class SingleCitationViewController: UIViewController {
...
+                    guard let self else { return }
+                    viewModel.process(action: .setOmitAuthor(omitAuthorSwitch.isOn))
```
michalrentka, line 110: Doesn't this need `self.viewModel.process(...)`? Or is `self` automatically added here too?

Zotero/Scenes/General/Views/ContainerViewController.swift
```diff
@@ -35,30 +35,120 @@ class ContainerViewController: UIViewController {
...
-        height.constant = container.preferredContentSize.height
-        width.constant = container.preferredContentSize.width
+        containerHeight.constant = container.preferredContentSize.height
+        containerWidth.constant = container.preferredContentSize.width
 
-        guard self.didAppear else { return }
+        guard didAppear else { return }
 
         UIView.animate(withDuration: 0.2) {
             self.view.layoutIfNeeded()
```
michalrentka, line 154: Since you removed `self`s everywhere, you can remove this one too.

michalrentka (changes requested):

Zotero/Scenes/Detail/CopyBibliography/ViewModels/CopyBibliographyActionHandler.swift
```diff
@@ -0,0 +1,78 @@
...
+                .flatMap { [weak webView] _ -> Single<String> in
+                    guard let webView else { return Single.error(CitationController.Error.prepareNotCalled) }
+                    return citationController.bibliography(for: itemIds, format: .html, in: webView)
+                }
+                .flatMap { [weak webView] html -> Single<(String, String?)> in
+                    if exportAsHtml { return Single.just((html, nil)) }
+                    guard let webView else { return Single.error(CitationController.Error.prepareNotCalled) }
+                    return citationController.bibliography(for: itemIds, format: .text, in: webView).flatMap({ Single.just((html, $0)) })
+                }
+                .subscribe(onSuccess: { [weak viewModel] data in
```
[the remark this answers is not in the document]
mvasilak: Using the convenience method, in case the object is not retained anymore, not even the part before the guard will be executed though. Of course in this instance it seems an unlikely edge case. Check https://github.com/zotero/zotero-ios/pull/827/commits/6230e70ee55e6670556eae5516d2c25628df4f0d

Zotero/Scenes/Detail/CopyBibliography/Views/CopyBibliographyViewController.swift
```diff
@@ -0,0 +1,171 @@
...
+    deinit {
+        viewModel.process(action: .cleanup)
+        DDLogInfo("CopyBibliographyViewController deinitialized")
+    }
+
+    // MARK: - View Lifecycle
+    override func loadView() {
+        super.loadView()
+
+        let overlayBody = UIView()
```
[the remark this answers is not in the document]
mvasilak: `super.loadView()` in this case is equal to `view = UIView()`, but you are right, better not call it, unless there is reason to. 
I have to say, I don't totally agree with the documented guideline. In cases where there is a mix of interface builder and programmatic UI, I think that `loadView` is a good place semantically to create all the UI you would have also build with the interface builder, including constraints. However, I'm happy to have us agree on a certain guideline and adhere to it in the codebase. Would that be, only create the main `view` of the controller in `loadView`, whenever this is needed for a reason, and setup the rest of the views in `viewDidLoad`?

Zotero/Scenes/Detail/DetailCoordinator.swift
mvasilak: Initially I did have it as part of `CopyBibliographyViewController`, but it doesn't seem the right place semantically. We cannot assume that the view controller knows that it is is presented in a container, so that it would remove itself as such. That's why I added the logic for dismissal, in the same level that handled the initial presentation. 
The cases you refer to only work when the circumstances exist, which admittedly are of the most common ones. E.g. for `dismiss` we assume there is a presenting view controller to handle the dismissal.

Zotero/Scenes/Detail/PDF/Views/PDFReaderViewController.swift
mvasilak: It's exactly what you describe from this implemented swift evolution [proposal](https://github.com/apple/swift-evolution/blob/main/proposals/0365-implicit-self-weak-capture.md)

Zotero/Scenes/Detail/CopyBibliography/Views/CopyBibliographyViewController.swift
```diff
@@ -0,0 +1,171 @@
...
+
+        func setupObserving() {
+            viewModel.stateObservable
+                .observe(on: MainScheduler.instance)
+                .subscribe(onNext: { state in
+                    update(state: state)
+                })
+                .disposed(by: disposeBag)
+
+            func update(state: CopyBibliographyState) {
```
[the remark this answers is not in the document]
mvasilak: This should also be added in our guidelines, that function for updating from a `ViewModel` state, should not be nested, but rather be a (private) method of the relevant view controller.
Check https://github.com/zotero/zotero-ios/pull/827/commits/7e5eccebcb0a2044d80477c3a59d9ada02329699

Zotero/Scenes/Detail/CopyBibliography/Views/CopyBibliographyViewController.swift
mvasilak: A guard against `view.alpha` has been added.
Check https://github.com/zotero/zotero-ios/pull/827/commits/7e5eccebcb0a2044d80477c3a59d9ada02329699

Zotero/Scenes/Detail/SingleCitation/Views/SingleCitationViewController.swift
mvasilak: Both styles are used throughout the codebase, and as I mentioned in a previous comment, the convenience `subscribe(with: ...)` method may be limiting in certain edge cases. Of course we can opt to not use it only then, and add also a comment in the code, to justify this change.
However, this convenience method, when user with `self`, as it is needed to be differentiated from the `object` parameter, by writing
```
subscribe(with: self, onNext: { `self`, _ in ...
```
means that you cannot allow implicit self for weak self captures, after self is unwrapped.
I.e. if you need to refer to multiple `self` properties you need to be explicit, such as
```
subscribe(with: self, onNext: { `self`, _ in
    self.viewModel.process(action: .setOmitAuthor(self.omitAuthorSwitch.isOn))
})
```
vs.
```
subscribe(onNext: { [weak self] _ in
    viewModel.process(action: .setOmitAuthor(omitAuthorSwitch.isOn))
})
```
Which direction should we follow?

Zotero/Scenes/Detail/CopyBibliography/Views/CopyBibliographyViewController.swift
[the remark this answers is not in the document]
michalrentka: I agree that `loadView` would kind of make more sense, but I'd still rather adhere to their recommendation, whatever reason they may have (some time ago I read somewhere that `loadView` should return as fast as possible and you can do whatever needed in `viewDidLoad` 🤷‍♂️).

Also the whole codebase initializes views in `viewDidLoad` currently and it's such a small detail that I don't think we need to rethink this one.

So if you need to create some kind of special root view you can do that in `loadView` (otherwise if it's just a `UIView` then `loadView` doesn't even need to be implemented). Then create all views, adjustments and constraints in `viewDidLoad`.

Zotero/Scenes/Detail/DetailCoordinator.swift
michalrentka: When hiding any view controller we always assume how it's shown and hide it based on that assumption. It works because all view controllers are (almost) always shown the same way (either in navigation controller or as modals). This view controller is pretty much only an overlay, so you can safely assume this won't even be pushed to a navigation controller or it won't be presented as a modal.

This also brings up a question, why isn't this presented as a modal over current context?

Zotero/Scenes/Detail/CopyBibliography/Views/CopyBibliographyViewController.swift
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/827/commits/a40cb59c457b715c97449228f79febb8bb01157e

Zotero/Scenes/Detail/PDF/PDFCoordinator.swift
michalrentka: Sure, you can make a common `func showCitation(for itemIds: Set<String>, libraryId: LibraryIdentifier, coordinatorDelegate: DetailCitationCoordinatorDelegate)` and call that from both coordinators.

Zotero/Scenes/Detail/CopyBibliography/Views/CopyBibliographyViewController.swift
```diff
@@ -0,0 +1,171 @@
...
+            stackView.topAnchor.constraint(equalTo: overlayBody.topAnchor, constant: 20),
+            stackView.leadingAnchor.constraint(equalTo: overlayBody.leadingAnchor, constant: 20),
+            overlayBody.bottomAnchor.constraint(equalTo: stackView.bottomAnchor, constant: 20),
+            overlayBody.trailingAnchor.constraint(equalTo: stackView.trailingAnchor, constant: 20)
+        ])
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        view.backgroundColor = UIColor(white: 0, alpha: 0.1)
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/827/commits/a40cb59c457b715c97449228f79febb8bb01157e

Zotero/Scenes/Detail/PDF/Views/PDFReaderViewController.swift
michalrentka: Right, I assumed that was the case, just forgot this was already in, since I mostly used explicit `self`s before. Alright then we can keep this like this, I don't like it, but it's the way the language is evolving, so no need to fight it.

Zotero/Scenes/Detail/SingleCitation/Views/SingleCitationViewController.swift
michalrentka: We use both styles because the `with: self` didn't exist initially and I started moving forward to it after it was introduced. But yes now that we use implicit `self` with `guard` it makes more sense to not use this feature. At least we'll have the same style in `RxSwift` and other closures.

So yes, sorry about the confusion, we'll keep doing the `weak` + `guard` dance to be able to use implicit `self` in closures.

Zotero/Scenes/Detail/DetailCoordinator.swift
mvasilak: I refactored this from a number of views that were part of `ItemsViewController`, so it is used in a similar manner to maintain the same functionality (check https://github.com/zotero/zotero-ios/pull/827/commits/196133df7240a06446d437de13bc44797621b4e1). We can improve it afterwards of course.

michalrentka: Yes, I implemented it as a simple subview overlay before, because it was only used there. But since it's extracted into a view controller, which can be shown in multiple view controllers, I think you can safely keep everything the same as it's now and just present it with `modalPresentationStyle = .overCurrentContext`?

mvasilak: Check https://github.com/zotero/zotero-ios/pull/827/commits/7b4c1a6a242b77c079e1ac995801d4ab7e30008f

Zotero/Scenes/Detail/PDF/PDFCoordinator.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/827/commits/6b4670aa6acab125bde828bfbe392feb681253dd

Zotero/Scenes/Detail/PDF/PDFCoordinator.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/827/commits/7b4c1a6a242b77c079e1ac995801d4ab7e30008f

Zotero/Scenes/Detail/PDF/PDFCoordinator.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/827/commits/7b4c1a6a242b77c079e1ac995801d4ab7e30008f

Zotero/Scenes/Detail/PDF/Views/PDFReaderViewController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/827/commits/8c9eb6f57987cb66e2a1929c4ab9451c3ea57ddb

Zotero/Scenes/General/Views/ContainerViewController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/827/commits/72089cea9a35bc6efa387ee9896b4b38a16a659b

Zotero/Scenes/Detail/SingleCitation/Views/SingleCitationViewController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/827/commits/db2e5780f0cc0667fa8537dbb94c3ce3aa466a60

mvasilak:
@michalrentka @dstillman, as part of these changes I propose we also make the locator button have the default color for buttons, as right now it is not immediately evident that it is a user control. Furthermore, we can change it to display a menu, instead of pushing a separate screen. What do you think?

michalrentka:
> @michalrentka @dstillman, as part of these changes I propose we also make the locator button have the default color for buttons, as right now it is not immediately evident that it is a user control. Furthermore, we can change it to display a menu, instead of pushing a separate screen. What do you think?

Do you mean in the "Quick Copy" settings screen?

mvasilak:
> Do you mean in the "Quick Copy" settings screen?

The `Citation Preview` screen, that is shown from the `Copy Citation` action.

michalrentka:
Sure, that would make sense. I don't know if button color will look best there, but I guess underline or the right arrow won't look any better. 

@yexingsha might have better ideas. Just to introduce you into the issue a bit, this is the window and I highlighted the button we're talking about. Currently it's just black as the rest of the text, so users probably don't even know it's a button.

IMG_00C7EFA900F4-1

yexingsha:
Does this button open up a menu/screen where users can select or type in page numbers? What does the menu/screen look like? (I can't get zotero-ios to build in Xcode 15, keep getting the error `Command PhaseScriptExecution failed with a nonzero exit code`)

If the button does what I think it does, I think it makes more sense for the touch target to be the entire row, like the rows with options in the iOS calendar new event screen. If it opens up a menu, then the trailing symbol should be `chevron.up.chevron.down`; if it opens up a new screen, then the trailing symbol should be `chevron.right`.

Citation Preview


(Also: what's the consideration behind making this UIMenu instead of popover?)

mvasilak:
@yexingsha here is a video of the current UI:

https://github.com/zotero/zotero-ios/assets/617841/6324cba4-843e-4996-a66f-70ba14bf6b85

There is a button and a textfield next to it, so it's not simply a row that can act as a selection menu, like the calendar case you mentioned. The reason to make it a menu is because it's simpler and it follows the latest guidelines for this type of interaction (we couldn't use a menu before, due to supporting older iOS versions). 

From what you propose, I think just adding the `chevron.up.chevron.down` next to the button would be a good indicattion that this is a user control. In my opinion, we should also use the default or another color for buttons, instead of using the same color as in simple text labels. 

If you have some other proposal, please do tell.

michalrentka (approved):
Current PR looks good, I'll review the UI changes when they come, but if needed, this can be merged as is.

dstillman:
What would the menu look like? That's what we use on the desktop, of course.

mvasilak:
@dstillman check this video for what it would look like (the button design is indicative, based on the previous discussion):

https://github.com/zotero/zotero-ios/assets/617841/0133c6c3-0eda-45ff-8e74-3e49dbc603d6

yexingsha:
I think this works, though I'm not sure about putting the `chevron.up.chevron.down` before the text. I think we just need use a native [pop-up button](https://developer.apple.com/design/human-interface-guidelines/pop-up-buttons) here, with the proper spacing between text and symbol, in default button color. We can also give the button a larger fixed width, so that "Number" starts off more to the right and doesn't move when selecting labels of different lengths:

Citation Preview

Also, making this screen/modal into grouped lists, like the mock here, will make the layout more readable.

mvasilak:
@yexingsha check the result of applying these changes (spacing between button title and image is the default one):
copy_citation_pop-up
copy_citation_pop-up_menu

Since the current implementation is not table view based, but much simpler, I omited the group list suggestion, for a future iteration.

@dstillman @michalrentka any comments on these suggestions?

michalrentka:
I'd just add that this window just had very temporary UI until we get a designer, so I'm glad we're taking a look at this now.

To comment those screenshots, I'd move those arrows a bit further from the button, it seems too close. Otherwise I'd create a new ticket to redesign the whole screen with suggestions from @yexingsha and we can continue further design discussion there (after we fix the button let's say).

yexingsha:
I think putting 4px spacing between the text and the symbol will be about right.

mvasilak:
@michalrentka I added the following UI changes in the PR (the rest will be discussed in a new issue)
- Increase the locator button label and image spacing per @yexingsha 's suggestion.
- Make said button a pop-up one (that shows a menu instead of pushing a screen).
- Change the keyboard `Return` button to `Done` when editing the locator text field, and have it dismiss the keyboard when tapped.
- Show the default clear button while the locator text field is edited.

michalrentka (approved):

PR #829, opened 2024-01-03
Support multiple open items

mvasilak:
Adds support for multiple open items.

- [x] Keep track of open items in order of most recently opened
- [x] Show in items navigation bar button that restores last item and shows count of currently open items
- [x] Save/restore state of open items
- [x] Add support for PDF items
- [x] Add support for note items
- [x] Show in presented item navigation bar button menu that allows to quickly switch to another open item 
- [x] Allow user to change order of presented open items (only move first & last currently)

In consideration
- [x] Support separate open items for multiple windows
- [x] Opening an already opened item in another window, should switch to that

Previously discussed in https://github.com/zotero/zotero-ios/pull/782

### DIFF
--- a/Zotero/Controllers/OpenItemsController.swift
+++ b/Zotero/Controllers/OpenItemsController.swift
@@ -322,8 +322,8 @@ final class OpenItemsController {
         setItemsSortedByUserIndex(existingItems, for: sessionIdentifier, validate: false)
     }
 
-    func restore(_ item: Item, using presenter: OpenItemsPresenter, completion: @escaping (Bool) -> Void) {
-        loadPresentation(for: item.kind) { [weak presenter] presentation in
+    func restore(_ kind: Item.Kind, using presenter: OpenItemsPresenter, completion: @escaping (Bool) -> Void) {
+        loadPresentation(for: kind) { [weak presenter] presentation in
             guard let presenter, let presentation else {
                 completion(false)
                 return
@@ -380,7 +380,7 @@ final class OpenItemsController {
         openItemPresenterProvider: @escaping () -> OpenItemsPresenter?,
         completion: @escaping (_ changedCurrentItem: Bool, _ openItemsChanged: Bool) -> Void
     ) -> UIDeferredMenuElement {
-        UIDeferredMenuElement.uncached { [weak self] elementProvider in
+        return UIDeferredMenuElement.uncached { [weak self] elementProvider in
             guard let self else {
                 elementProvider([])
                 return
@@ -392,7 +392,7 @@ final class OpenItemsController {
             let itemsCount = itemTuples.count
             for (index, (item, rItem)) in itemTuples.enumerated() {
                 if item == openItem {
-                    var currentItemActions: [UIAction] = []
+                    var currentItemActions: [UIMenuElement] = []
                     let closeAction = UIAction(title: L10n.Accessibility.Pdf.currentItemClose, image: .init(systemName: "xmark.circle")) { [weak self] _ in
                         guard let self else { return }
                         close(item.kind, for: sessionIdentifier)
@@ -429,17 +429,17 @@ final class OpenItemsController {
                         }
                         currentItemActions.append(closeOtherAction)
                     }
+                    if let siblingNotesMenu = siblingNotesMenu(rItem: rItem) {
+                        currentItemActions.append(siblingNotesMenu)
+                    }
+                    if let siblingAttachmentsMenu = siblingAttachmentsMenu(rItem: rItem) {
+                        currentItemActions.append(siblingAttachmentsMenu)
+                    }
                     let currentItemMenu = UIMenu(title: L10n.Accessibility.Pdf.currentItem, options: [.displayInline], children: currentItemActions)
                     let currentItemElement = UIMenu(title: rItem.displayTitle, image: item.kind.icon, children: [currentItemMenu])
                     elements.append(currentItemElement)
                 } else {
-                    let itemAction = UIAction(title: rItem.displayTitle, image: item.kind.icon) { [weak self] _ in
-                        guard let self, let presenter = openItemPresenterProvider() else { return }
-                        restore(item, using: presenter) { restored in
-                            completion(restored, false)
-                        }
-                    }
-                    elements.append(itemAction)
+                    elements.append(createKindAction(title: rItem.displayTitle, kind: item.kind, openItemsChanged: false))
                 }
             }
 
@@ -453,6 +453,68 @@ final class OpenItemsController {
             elements.append(closeAllElement)
 
             elementProvider(elements)
+        }
+
+        func siblingNotesMenu(rItem: RItem) -> UIMenu? {
+            guard let parent = rItem.parent else { return nil }
+            let notes = Array(parent.children.filter(.items(type: ItemTypes.note, notSyncState: .dirty, trash: false)).sorted(byKeyPath: "displayTitle"))
+            let actions: [UIAction] = notes.compactMap({
+                guard $0.key != rItem.key else { return nil }
+                let kind = Item.Kind.note(libraryId: $0.libraryIdentifier, key: $0.key)
+                return createKindAction(title: $0.displayTitle, kind: kind, openItemsChanged: true)
+            })
+            guard !actions.isEmpty else { return nil }
+            return UIMenu(title: L10n.Accessibility.Pdf.itemNotes, children: actions)
+        }
+
+        func siblingAttachmentsMenu(rItem: RItem) -> UIMenu? {
+            guard let parent = rItem.parent else { return nil }
+            let attachments: [Attachment] = Array(
+                parent.children.filter(.items(type: ItemTypes.attachment, notSyncState: .dirty, trash: false))
+                    .sorted(byKeyPath: "displayTitle")
+                    .compactMap({ [weak fileStorage] in
+                        guard let fileStorage else { return nil }
+                        return AttachmentCreator.attachment(for: $0, fileStorage: fileStorage, urlDetector: nil)
+                    })
+            )
+            let titleAndKindTuples: [(String, Item.Kind)] = attachments.compactMap({
+                guard $0.key != rItem.key else { return nil }
+                let kind: Item.Kind
+                switch $0.type {
+                case .file(_, let contentType, _, _, _):
+                    switch contentType {
+                    case "application/pdf":
+                        kind = .pdf(libraryId: $0.libraryId, key: $0.key)
+
+                    case "text/html":
+                        guard FeatureGates.enabled.contains(.htmlEpubReader) else { return nil }
+                        kind = .html(libraryId: $0.libraryId, key: $0.key)
+
+                    case "application/epub+zip":
+                        guard FeatureGates.enabled.contains(.htmlEpubReader) else { return nil }
+                        kind = .epub(libraryId: $0.libraryId, key: $0.key)
+
+                    default:
+                        return nil
+                    }
+
+                case .url:
+                    return nil
+                }
+                return ($0.title, kind)
+            })
+            let actions = titleAndKindTuples.map({ createKindAction(title: $0.0, kind: $0.1, openItemsChanged: true) })
+            guard !actions.isEmpty else { return nil }
+            return UIMenu(title: L10n.Accessibility.Pdf.itemAttachments, children: actions)
+        }
+
+        func createKindAction(title: String, kind: Item.Kind, openItemsChanged: Bool) -> UIAction {
+            return UIAction(title: title, image: kind.icon) { [weak self] _ in
+                guard let self, let presenter = openItemPresenterProvider() else { return }
+                restore(kind, using: presenter) { restored in
+                    completion(restored, openItemsChanged)
+                }
+            }
         }
     }
 
--- a/Zotero/Assets/en.lproj/Localizable.strings
+++ b/Zotero/Assets/en.lproj/Localizable.strings
@@ -593,6 +593,8 @@
 "accessibility.pdf.toggle_annotation_toolbar" = "Toggle annotation toolbar";
 "accessibility.pdf.show_more_tools" = "Show more";
 "accessibility.pdf.open_items" = "Open Items";
+"accessibility.pdf.item_notes" = "Item Notes";
+"accessibility.pdf.item_attachments" = "Item Attachments";
 "accessibility.pdf.current_item" = "Current Item";
 "accessibility.pdf.current_item_close" = "Close";
 "accessibility.pdf.current_item_move_to_start" = "Move to start";


mvasilak:
@michalrentka @dstillman 
this PR is ready for some more eyes and perhaps beta testing. It has the underlying facility to support:
* multiple open items
* switching between open items 
* opening new items
* closing existing items
* moving open items to another position (user index)
* keeping track of last opened date for each item
* opening an item in its corresponding window/scene if it's already open

It currently supports PDF documents and notes, and should be straight-forward to add support for e.g. web snapshots and ePubs.

UI-wise, it includes a bare minimum implementation using simple UIMenus, with these features:
* In the items screen, a bar button appears with a number indicating the number of open items.
Screenshot 2024-04-18 at 16 10 00

* Tapping the button, will open the last opened item.
* Tapping and holding will show a menu that allows opening specific item, or closing all open items.
Screenshot 2024-04-18 at 16 13 57

* Similar bar button appears in the PDF document/note screen.
Screenshot 2024-04-18 at 16 15 22
Screenshot 2024-04-18 at 16 16 01

* Tapping the button will always show a menu. It contains all the open items as before, but also adds a sub-menu for the current item. Said sub-menu allows to close current item, move to start (if not already), move to end (if not already), close other items (if other items are open).

Screenshot 2024-04-18 at 16 18 24

Next (possible) steps are (a) an actual tab bar and (b) instead of using UIMenus create a custom view that allows e.g. closing any item and re-ordering.

Feedback regarding icons and copy used will be much appreciated.

Zotero/Controllers/OpenItemsController.swift
```diff
@@ -0,0 +1,481 @@
...
+            for item in items {
+                let libraryId = item.kind.libraryId
+                let key = item.kind.key
+                var keys = keysByLibraryIdentifier[libraryId, default: .init()]
+                keys.insert(key)
+                keysByLibraryIdentifier[libraryId] = keys
+            }
+            do {
+                try dbStorage.perform(on: .main) { coordinator in
+                    let objects = try coordinator.perform(request: ReadItemsWithKeysFromMultipleLibrariesDbRequest(keysByLibraryIdentifier: keysByLibraryIdentifier))
```
michalrentka, line 170: Same as above, this doesn't need to happen inside closure with coordinator. You can just get `objects` directly from `dbStorage` and observe them.

Zotero/Controllers/OpenItemsController.swift
```diff
@@ -0,0 +1,481 @@
...
+    // MARK: Object Lifecycle
+    init(dbStorage: DbStorage, fileStorage: FileStorage) {
+        self.dbStorage = dbStorage
+        self.fileStorage = fileStorage
+        disposeBag = DisposeBag()
+    }
+    
+    // MARK: Actions
+    func observable(for sessionIdentifier: String) -> PublishSubject<[Item]> {
+        if let observable = observableBySessionIdentifier[sessionIdentifier] {
```
michalrentka, line 124: Are these always accessed from main thread? Otherwise we'll need some kind of lock or queue to access them. Same with `sessionIdentifier` below.

Zotero/Controllers/OpenItemsController.swift
```diff
@@ -0,0 +1,481 @@
...
+            let newItems = existingItems + [item]
+            // setItems will produce next observable event
+            setItems(newItems, for: sessionIdentifier, validate: false)
+        }
+    }
+
+    func close(_ kind: Item.Kind, for sessionIdentifier: String) {
+        DDLogInfo("OpenItemsController: closed open item \(kind) for \(sessionIdentifier)")
+        var existingItems = itemsSortedByUserOrder(for: sessionIdentifier)
+        if let index = existingItems.firstIndex(where: { $0.kind == kind }) {
```
michalrentka, line 228: This would look nicer as a guard, but not mandatory. Same with `move` function below.

Zotero/Controllers/OpenItemsController.swift
```diff
@@ -0,0 +1,481 @@
...
+        }
+        return itemTuples
+    }
+
+    private func filterValidItems(_ items: [Item]) -> [Item] {
+        filterValidItemsWithRItem(items).map { $0.0 }
+    }
+
+    private func loadPresentation(for item: Item) -> Presentation? {
+        switch item.kind {
```
michalrentka, line 401: Here you could do the `do-catch` block with `try dbStorage.perform(on: .main) { coordinator in` in this function and pass `coordinator` to `loadPDFPresentation` and `loadNotePresentation?` and both would be throwable, something like: 
```
func loadPDFPresentation(key: String, libraryId: LibraryIdentifier, coordinator: DbCoordinator) throws -> Presentation?
func loadNotePresentation(key: String, libraryId: LibraryIdentifier, coordinator: DbCoordinator) throws -> Presentation?
```

and then the child functions would be a bit cleaner without repeating `do-catch` blocks.

michalrentka:
Some bugs and smaller issues I caught:
1) On iPhone, the square bar button item for tabs seems to have a lot of space around it
bar button
2) When I open multiple PDFs, then I scroll to some page on one, switch to another one and switch back to the original one, the original one is not on the page I scrolled to, but on page that it was when I opened it

Zotero/Extensions/NSUserActivity+Activities.swift
```diff
@@ -9,29 +9,47 @@
...
+        activity.addUserInfoEntries(from: userInfo)
+        return activity
     }
 
-    static func pdfActivity(for key: String, libraryId: LibraryIdentifier, collectionId: CollectionIdentifier) -> NSUserActivity {
+    static func pdfActivity(with openItems: [OpenItem], libraryId: LibraryIdentifier, collectionId: CollectionIdentifier) -> NSUserActivity {
         let activity = NSUserActivity(activityType: self.pdfId)
-        var pdfUserInfo: [AnyHashable: Any] = ["key": key, "libraryId": libraryIdToString(libraryId)]
+        activity.addUserInfoEntries(from: openItemsToUserInfo(openItems: openItems))
+        var userInfo: [AnyHashable: Any] = [libraryIdKey: libraryIdToString(libraryId), restoreMostRecentlyOpenedItemKey: true]
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/829/commits/0e0296c49ff172a982cbf1ed56f558e9cbbeca09

Zotero/Scenes/AppCoordinator.swift
```diff
@@ -176,85 +178,49 @@ extension AppCoordinator: AppDelegateCoordinatorDelegate {
...
-                        } catch let error {
-                            DDLogError("AppCoordinator: can't load restored data - \(error)")
-                            return nil
-                        }
+                    var collection: Collection?
 
-                        guard let url  else { return nil }
-                        return (url, collection, parentKey)
+                    do {
+                        try dbStorage.perform(on: .main, with: { coordinator in
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/829/commits/2ca8013a7d8c096af5b47443b89d562365a64f92

Zotero/Scenes/AppCoordinator.swift
```diff
@@ -863,4 +836,30 @@ extension AppCoordinator: SyncRequestReceiver {
     }
 }
 
+extension AppCoordinator: OpenItemsPresenter {
+    func showItem(with presentation: ItemPresentation?) {
+        guard let presentation, let window, let mainController = window.rootViewController as? MainViewController else { return }
+        mainController.getDetailCoordinator { [weak self] coordinator in
+            guard let self else { return }
+            self.show(
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/829/commits/2ca8013a7d8c096af5b47443b89d562365a64f92

Zotero/Scenes/Detail/DetailCoordinator.swift
```diff
@@ -910,15 +958,50 @@ extension DetailCoordinator: DetailNoteEditorCoordinatorDelegate {
...
+                return
+            }
+            openItemsController.open(kind, for: sessionIdentifier)
+        }
+
+        let viewControllerProvider: () -> DetailNavigationViewController = {
+            self.createNoteController(library: library, kind: kind, text: text, tags: tags, title: title, saveCallback: amendedSaveCallback)
         }
-        navigationController.present(controller, animated: true)
+        if let presentedViewController = navigationController.presentedViewController {
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/829/commits/08d13a23c148316b549c3b3b2a755b5e20debb47

Zotero/Scenes/Detail/Items/Views/ItemsViewController.swift
```diff
@@ -473,7 +486,15 @@ final class ItemsViewController: UIViewController {
         item.isEnabled = enabled
     }
 
+    private func updateRestoreOpenItemsButton(withCount count: Int) {
+        guard let item = self.navigationItem.rightBarButtonItems?.first(where: { button in RightBarButtonItem(rawValue: button.tag) == .restoreOpenItems }) else { return }
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/829/commits/0ea653eec6b01116e302f8b3b800fc1ddcad0dc4

Zotero/Scenes/Detail/Items/Views/ItemsViewController.swift
```diff
@@ -473,7 +486,15 @@ final class ItemsViewController: UIViewController {
...
+        item.image = UIImage(systemName: "\(count).square")
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/829/commits/0ea653eec6b01116e302f8b3b800fc1ddcad0dc4
For the time being I just opted for another image if the count is more than 50, as it seems an edge case. We can easily modify it in this single place if we need to, based on feedback.

Zotero/Extensions/UIViewController+Extensions.swift
```diff
@@ -66,4 +66,8 @@ extension UIViewController {
         // Parent also didn't return a scene. Trying presenting view controller.
         return presentingViewController?.scene
     }
+
+    func getSessionIdentifier() -> String? {
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/829/commits/095ac4fe156fd8c5f3d7836a63662cbdfddfc57e
Previously, it was clashing with the property passed to `MainViewController`. I removed it and it is fetched by the optional property instead.

Zotero/Controllers/OpenItemsController.swift
mvasilak: Yes, these are used when a view loads, a view controller is presented, or a menu action is called.

Zotero/Controllers/OpenItemsController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/829/commits/ae2376422f3f11d32545767dac3d7b35913fac7a

Zotero/Controllers/OpenItemsController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/829/commits/ae2376422f3f11d32545767dac3d7b35913fac7a

Zotero/Controllers/OpenItemsController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/829/commits/ae2376422f3f11d32545767dac3d7b35913fac7a

Zotero/Scenes/Detail/PDF/Views/PDFReaderViewController.swift
```diff
@@ -105,6 +106,28 @@ class PDFReaderViewController: UIViewController {
         share.menu = UIMenu(children: [deferredMenu])
         return share
     }()
+    private lazy var openItemsButton: UIBarButtonItem = {
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/829/commits/250fdb4804b13eebe91d5314042565bf60b30b32
The items view controller case is somewhat different. The PDF reader and note editor cases have more similarities, especially regarding their menu, but they do make use of different view model and state structs. I suggest to make at some point a detail view controller base class (or protocol w/ default implementations), with similar base view model and state, so each specific case can extend to inherint the common logic.

mvasilak:
@michalrentka 
>     1. On iPhone, the square bar button item for tabs seems to have a lot of space around it

It's the standard spacing for bar button items created with an image, it just seems like that because is next to another such button. We can use a custom view button instead, to decrease the image insets. Maybe we should consider using `UIBarButtonItemGroup` to group related buttons when there are space constraints.
Screenshot 2024-04-25 at 16 37 42

>     2. When I open multiple PDFs, then I scroll to some page on one, switch to another one and switch back to the original one, the original one is not on the page I scrolled to, but on page that it was when I opened it

Good catch, fixed in https://github.com/zotero/zotero-ios/pull/829/commits/33b5f638ad7318e2a3a6d3aed84fbcb89a9beed7

michalrentka:
Looks good to me now, I didn't find new bugs yet. We should probably make an internal build for others to test or push it to beta. There are a ton of changes, so we should definitely test this thoroughly. @dstillman can decide.

mvasilak:
@michalrentka rebased to master, to include the activity/scene title logic, and made some relevant improvements for the note editor.

Zotero/Scenes/General/Models/NoteEditorState.swift
```diff
@@ -64,6 +66,13 @@ struct NoteEditorState: ViewModelState {
         self.title = title
         changes = []
         self.openItemsCount = openItemsCount
+        displayTitle = generateDisplayTitle()
+    }
+
+    func generateDisplayTitle() -> String? {
```
michalrentka, line 72: Should tab display title for a note consist of the "title" in this context? The "title" in note editor is parent's item display title and item type. So when viewing tabs you'd see "Parent title / actual note title", but it would be too big, so you probably wouldn't see much.

I'd do the following: 
1. Rename `displayTitle` to something like `sceneTitle` or `tabTitle` because it's a bit confusing. The `displayTitle` is not really displayed anywhere in the app, the `title` is displayed in navigation bar when available.
2. Don't generate `displayTitle` here, pass it from `RItem.displayTitle`, same as `title` is being passed. 
3. When the note updates, the `displayTitle` updates too, so you could update `displayTitle` on each save in `NoteEditorActionHandler.save`
4. Use only `RItem.displayTitle` of the note item, don't include parent item, it's too long anyway.

michalrentka (changes requested):

Zotero/Scenes/General/Models/NoteEditorState.swift
mvasilak: Since this string is used for the current activity, as well as for the scene title in the app switcher, I picked `activityTitle` as a more apt name. Check https://github.com/zotero/zotero-ios/pull/829/commits/635fbde25259baa27e62e1b92b66f1f9e2fc8d01
Also made the same name change for `PDFReaderState` in https://github.com/zotero/zotero-ios/pull/829/commits/631b1459dd13e0c8386c9dc7f1245a67902cecf6

Zotero/Scenes/General/NoteEditorCoordinator.swift
```diff
@@ -75,7 +75,15 @@ final class NoteEditorCoordinator: NSObject, Coordinator {
...
 
-        let state = NoteEditorState(kind: kind, library: library, title: title, text: initialText, tags: initialTags, openItemsCount: openItemsController.getItems(for: sessionIdentifier).count)
+        let state = NoteEditorState(
+            kind: kind,
+            library: library,
+            title: title,
+            text: initialText,
+            tags: initialTags,
+            openItemsCount: openItemsController.getItems(for: sessionIdentifier).count,
+            activityTitle: kind.key.flatMap { (try? dbStorage.perform(request: ReadItemDbRequest(libraryId: library.identifier, key: $0), on: .main))?.displayTitle }
```
michalrentka, line 85: You don't need to re-read this from database. When showing a note you already start with an RItem (or Note), so you can just pass activityTitle from that when creating NoteEditorCoordinator. So you'd add activityTitle parameter to DetailCoordinator.showNote and I see we call showNote in ItemDetailViewController and ItemsViewController, where you have either RItem or Note available.

michalrentka (changes requested):

Zotero/Scenes/General/NoteEditorCoordinator.swift
mvasilak: There are cases now where the Note is not available, just individual parameters, but let me check if I can refactor this up the chain.

mvasilak: Check https://github.com/zotero/zotero-ios/pull/829/commits/008228c3f8c93d7d76003a5a6286f34cf6961d03
I selected to rename `title: NoteEditorState.TitleData?` to `parentTitleData` and `activityTitle: String?` to simply `title` for further clarification.

michalrentka: Looks good now, I guess you can rename the `activityTitle` in PDF reader too, not sure whether it conflicted with another title there as well.

Zotero/Controllers/OpenItemsController.swift
michalrentka: I don't really mind either approach, but why did you move back to previous format? Did it cause any issues?

Zotero/Controllers/OpenItemsController.swift
```diff
@@ -98,4 +101,438 @@ final class OpenItemsController {
...
+        }
+    }
+}
+
+extension OpenItemsController {
+    func openItemsUserActivity(for sessionIdentifier: String?, libraryId: LibraryIdentifier, collectionId: CollectionIdentifier? = nil) -> NSUserActivity {
+        var items: [Item] = []
+        if let sessionIdentifier {
+            items = getItems(for: sessionIdentifier)
+        }
```
michalrentka, line 513: `let items = sessionIdentifier.flatMap({ getItems(for: $0) }) ?? []`

Zotero/Scenes/AppCoordinator.swift
```diff
@@ -207,8 +211,12 @@ extension AppCoordinator: AppDelegateCoordinatorDelegate {
                     collection = Collection(custom: .all)
                 }
                 mainController.showItems(for: collection, in: data.libraryId)
-                guard data.restoreMostRecentlyOpenedItem, let item = data.openItems.first else { return }
-                restoreMostRecentlyOpenedItem(using: self, item: item)
+                guard data.restoreMostRecentlyOpenedItem else { return }
+                openItemsController.restoreMostRecentlyOpenedItem(using: self, sessionIdentifier: sessionIdentifier) { item in
+                    if item == nil {
+                        DDLogInfo("AppCoordinator: no open item to restore")
```
michalrentka, line 217: You can add log for non-nil `item` too.

Zotero/Scenes/AppCoordinator.swift
```diff
@@ -424,11 +375,23 @@ extension AppCoordinator: AppDelegateCoordinatorDelegate {
...
         }
         return true
     }
+
+    func continueUserActivity(_ userActivity: NSUserActivity, for sessionIdentifier: String) {
+        guard userActivity.activityType == NSUserActivity.contentContainerId, let window, let mainController = window.rootViewController as? MainViewController else { return }
+        mainController.getDetailCoordinator { [weak self] coordinator in
+            self?.controllers.userControllers?.openItemsController.restoreMostRecentlyOpenedItem(using: coordinator, sessionIdentifier: sessionIdentifier) { item in
+                if item == nil {
+                    DDLogInfo("AppCoordinator: no open item to restore for continued user activity")
```
michalrentka, line 390: Same as above

mvasilak: Check https://github.com/zotero/zotero-ios/pull/829/commits/055044308a8c2ccae4bb61b6472cc1e9676a3c9d

Zotero/Scenes/AppCoordinator.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/829/commits/055044308a8c2ccae4bb61b6472cc1e9676a3c9d

Zotero/Controllers/OpenItemsController.swift
```diff
@@ -98,4 +101,438 @@ final class OpenItemsController {
...
+            var keysByLibraryIdentifier: [LibraryIdentifier: Set<String>] = [:]
+            for item in items {
+                let libraryId = item.kind.libraryId
+                let key = item.kind.key
+                var keys = keysByLibraryIdentifier[libraryId, default: .init()]
+                keys.insert(key)
+                keysByLibraryIdentifier[libraryId] = keys
+            }
+            do {
+                let objects = try dbStorage.perform(request: ReadItemsWithKeysFromMultipleLibrariesDbRequest(keysByLibraryIdentifier: keysByLibraryIdentifier), on: .main)
```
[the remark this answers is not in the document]
mvasilak: `keysByLibraryIdentifier` is a local variable that is populated as a parameter for the db request, it's not used outside of the function scope.

Zotero/Controllers/OpenItemsController.swift
```diff
@@ -98,4 +101,438 @@ final class OpenItemsController {
...
+        let observable = PublishSubject<[Item]>()
+        observableBySessionIdentifier[sessionIdentifier] = observable
+        return observable
+    }
+
+    func getItems(for sessionIdentifier: String) -> [Item] {
+        itemsBySessionIdentifier[sessionIdentifier, default: []]
+    }
+
+    func setItems(_ items: [Item], for sessionIdentifier: String, validate: Bool) {
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/829/commits/055044308a8c2ccae4bb61b6472cc1e9676a3c9d

Zotero/Controllers/OpenItemsController.swift
mvasilak: I rewrote the part to add a completion, and I guess I followed the same steps 😅 Check https://github.com/zotero/zotero-ios/pull/829/commits/a607962689c73afac08a839640c22cc3a5d4d0d0

mvasilak:
@michalrentka ready for another look:
- rebased to master
- it simplifies open URL logic by letting the open items controller to download any attachment if needed
- it improves the note editor by opening any attachment links with multiple open items support

Zotero/Extensions/UIViewController+Extensions.swift
```diff
@@ -69,4 +69,8 @@ extension UIViewController {
         // Parent also didn't return a scene. Trying presenting view controller.
         return presentingViewController?.scene
     }
+
+    var sessionIdentifier: String? {
```
michalrentka: All `UIViewController`s which support multiple items controls also contain the `UIButton` for triggering the deferred `UIMenu` and observing of item count. I know the UI will change to tabs at some points, but it might be worth putting those two into this extension or some new extension which would handle the tabbing UI.

I think we even mentioned this at some point, so just to keep that in mind.

PR #831, opened 2024-01-10
Fix memory leaks

mvasilak:
Fixes several memory leaks. More fixes may follow in a separate PRs.

### DIFF
--- a/Zotero/Scenes/Detail/Annotation Popover/Views/AnnotationViewController.swift
+++ b/Zotero/Scenes/Detail/Annotation Popover/Views/AnnotationViewController.swift
@@ -213,8 +213,9 @@ final class AnnotationViewController: UIViewController {
             commentView.textObservable
                        .debounce(.milliseconds(500), scheduler: MainScheduler.instance)
                        .subscribe(with: self, onNext: { `self`, data in
-                           self.viewModel.process(action: .setComment(key: annotation.key, comment: data.0))
-                           if data.1 {
+                           guard let (text, needsHeightReload) = data else { return }
+                           self.viewModel.process(action: .setComment(key: annotation.key, comment: text))
+                           if needsHeightReload {
                                self.updatePreferredContentSize()
                                self.scrollToCursorIfNeeded()
                            }
--- a/Zotero/Scenes/Detail/PDF/Views/Annotation View/AnnotationView.swift
+++ b/Zotero/Scenes/Detail/PDF/Views/Annotation View/AnnotationView.swift
@@ -17,7 +17,7 @@ final class AnnotationView: UIView {
 
     enum Action {
         case tags
-        case options(UIButton)
+        case options(UIButton?)
         case reloadHeight
         case setComment(NSAttributedString)
         case setCommentActive(Bool)
--- a/Zotero/Scenes/Detail/PDF/Views/Annotation View/AnnotationView.swift
+++ b/Zotero/Scenes/Detail/PDF/Views/Annotation View/AnnotationView.swift
@@ -232,7 +232,8 @@ final class AnnotationView: UIView {
         })
         self.commentTextView.textObservable
             .debounce(.milliseconds(500), scheduler: MainScheduler.instance)
-            .subscribe(onNext: { [weak self] text, needsHeightReload in
+            .subscribe(onNext: { [weak self] data in
+                guard let (text, needsHeightReload) = data else { return }
                 self?.actionPublisher.on(.next(.setComment(text)))
                 if needsHeightReload {
                     self?.actionPublisher.on(.next(.reloadHeight))
--- a/Zotero/Scenes/Detail/PDF/Views/Annotation View/AnnotationView.swift
+++ b/Zotero/Scenes/Detail/PDF/Views/Annotation View/AnnotationView.swift
@@ -232,11 +232,11 @@ final class AnnotationView: UIView {
         })
         self.commentTextView.textObservable
             .debounce(.milliseconds(500), scheduler: MainScheduler.instance)
-            .subscribe(onNext: { text, needsHeightReload in
-                self.actionPublisher.on(.next(.setComment(text)))
+            .subscribe(onNext: { [weak self] text, needsHeightReload in
+                self?.actionPublisher.on(.next(.setComment(text)))
                 if needsHeightReload {
-                    self.actionPublisher.on(.next(.reloadHeight))
-                    self.scrollToBottomIfNeeded()
+                    self?.actionPublisher.on(.next(.reloadHeight))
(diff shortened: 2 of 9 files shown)


Zotero/Scenes/Detail/Annotation Popover/Views/AnnotationViewController.swift
```diff
@@ -213,8 +213,9 @@ final class AnnotationViewController: UIViewController {
             commentView.textObservable
                        .debounce(.milliseconds(500), scheduler: MainScheduler.instance)
                        .subscribe(with: self, onNext: { `self`, data in
-                           self.viewModel.process(action: .setComment(key: annotation.key, comment: data.0))
-                           if data.1 {
+                           guard let (text, needsHeightReload) = data else { return }
```
michalrentka, line 216: Since you're changing this, can you switch to the style we agreed on previously? In this case remove `with: self` and just to `[weak self]` and `let self` in `guard` so that we don't need explicit `self`s in closure.

Anyway what was the leak here? The fact that we used `data` directly?

Zotero/Scenes/Detail/PDF/Views/Annotation View/AnnotationView.swift
```diff
@@ -232,11 +232,12 @@ final class AnnotationView: UIView {
         })
         self.commentTextView.textObservable
             .debounce(.milliseconds(500), scheduler: MainScheduler.instance)
-            .subscribe(onNext: { text, needsHeightReload in
-                self.actionPublisher.on(.next(.setComment(text)))
+            .subscribe(onNext: { [weak self] data in
+                guard let (text, needsHeightReload) = data else { return }
```
michalrentka, line 236: Add `let self` to `guard` and remove `self?`.

Zotero/Scenes/Detail/SingleCitation/Views/SingleCitationViewController.swift
```diff
@@ -131,8 +138,9 @@ final class SingleCitationViewController: UIViewController {
         updatePreferredContentSize()
     }
 
-    deinit {
-        viewModel.process(action: .cleanup)
+    override func viewWillDisappear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
+        previewWebView.configuration.userContentController.removeAllScriptMessageHandlers()
```
michalrentka, line 143: Would it be ok to do this in `deinit` too? These handlers are setup in `viewDidLoad` and removed here in `viewWillDisappear`. So if this view disappears and re-appears again they won't work. I know that UI probably doesn't allow that now, but if we add something later it might cause an issue.

Zotero/Scenes/General/Views/ContainerViewController.swift
```diff
@@ -156,8 +145,19 @@ class ContainerViewController: UIViewController {
     }
 
     // MARK: - Actions
-
     @objc private func hide() {
         presentingViewController?.dismiss(animated: true, completion: nil)
     }
+
+    private func moveToKeyboard(_ data: KeyboardData, willShow: Bool) {
```
michalrentka, line 152: I guess we should decide on a rule here. If I understand correctly you moved this away from `viewDidLoad` because it was capturing `self`? 

So if we want to stick to our rule, that if a function is used only in one place, it should be moved to that place as child function, this should be moved back and you should add `self: ContainerViewController` parameter and use explicit `self` in that child function.

Or we want to make an exception to our rule so that this case is simpler. Then we'll have some one-use standalone functions as this one while others are "hidden" as child functions.

I don't really know which is best. We already make an exception for `update(state: SomeState)` functions because otherwise we'd have the whole view controller included in a `viewDidLoad`. So that makes sense to me. I'm not really sure with this one. I guess I'm slightly more inclined to include `self` as parameter to keep it a bit cleaner. But then we'll have explicit `self`s in some functions (which we already have to have in some places anyway).

Thoughts? :)

michalrentka (changes requested):
Some minor things only. Good job on figuring out these!

Zotero/Scenes/Detail/PDF/Views/Annotation View/AnnotationView.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/831/commits/712a9acd8014d8279613bf2f9f1670d42f1cd8e6
The `let self = self` style is necessary, otherwise the compiler complains. Maybe they forgot to update some compiler logic for function builders?

Zotero/Scenes/Detail/SingleCitation/Views/SingleCitationViewController.swift
mvasilak: This needs to be called before, so that the reference is removed, in order to trigger deinitialization. We can move the handler addition in `viewWillAppear`, so that they are always balanced. I tested it, and it works as expected. 
Check https://github.com/zotero/zotero-ios/pull/831/commits/6e44151ba8a5b15455e029416b3a08976d9aad23

Zotero/Scenes/Detail/Annotation Popover/Views/AnnotationViewController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/831/commits/66585bb108b54a05de0ecf27945dfbae633d64c7

The leak is in the `textObservable`, because it creates a circular capture to it's owner. The change here just reflects that `data is now an optional tuple, and I just made it similar to the other place it is used, in `AnnotationView`.

Zotero/Scenes/General/Views/ContainerViewController.swift
mvasilak: I guess we should just discuss these cases and see if we can come with a list of rules/suggestions to follow.
Indeed `update(state: SomeState)` is a very clear and cut case.

Here the issue is the indirect capture of `self` in the notification blocks, as the nested function can access the context of its parent function. 

So the rule to not rely on the context, but pass directly what is needed, forces as to be more aware. But still, it is easy to write the name of a class property without thinking about it, and thus cause `self` capture (granted, one of the pitfalls of not having to prefix `self.` explicitly).

The counter argument rule could be to not nest any function called by an escaped block, so that we are certain there are no indirect captures, but not all such cases are problematic.

I'd say let's go with the first one for the time being, and re-evaluate if it becomes too cumbersome. I'm guessing that this is not a common issue anyway.

Check https://github.com/zotero/zotero-ios/pull/831/commits/1758c9ef6bd22d7fef78216fa5b37dd4b487361c

michalrentka (approved):

PR #836, opened 2024-01-29
Html / Epub reader implementation

michalrentka:


jarbus:
I've literally been stalking this repo since last November because I've been so excited for epub support in Zotero haha, so great to see this PR

Zotero/Scenes/Detail/HTML:EPUB/ViewModels/HtmlEpubReaderActionHandler.swift
```diff
@@ -0,0 +1,876 @@
...
+            var jsons: [[String: Any]] = []
+
+            for item in items {
+                guard let (annotation, json) = item.htmlEpubAnnotation else { continue }
+                jsons.append(json)
+                sortedKeys.append(annotation.key)
+                annotations[item.key] = annotation
+            }
+
+            let jsonString = WebViewEncoder.encodeAsJSONForJavascript(jsons)
```
mvasilak, line 613: This should not be encoded in base 64.
E.g.:
```suggestion
            let jsonString = WebViewEncoder.encodeAsJSONForJavascript(jsons)
```
Better yet if `WebViewEncoder.encodeAsJSONForJavascript` accepts an extra parameter that determines the final encoding. I guess that this method should return `String?` instead of `String`, so the caller determine what the value should be in case it fails, e.g. here it is `[]`.

Zotero/Models/Files.swift
```diff
@@ -174,6 +174,12 @@ struct Files {
         return FileData(rootPath: Files.cachesRootPath, relativeComponents: ["Zotero", "shareext", "downloads"], name: "item_\(key)", contentType: contentType)
     }
 
+    // MARK: Epub/Html reader
+
+    static var tmpReaderDirectory: File {
+        return FileData.directory(rootPath: Files.cachesRootPath, relativeComponents: ["Zotero", UUID().uuidString])
```
mvasilak, line 180: It seems that we use each tmp reader directory only once. If it's fast enough to create it, we can also delete it when the reader closes. Otherwise, maybe we can re-use it from the cache if it is still there?

mvasilak (commented):
Some first comments. More will follow.

Zotero/Scenes/Detail/DetailCoordinator.swift
```diff
@@ -286,7 +286,7 @@ final class DetailCoordinator: Coordinator {
         navigationController.present(controller, animated: true, completion: nil)
     }
 
-    func createPDFController(key: String, parentKey: String?, libraryId: LibraryIdentifier, url: URL, page: Int? = nil, preselectedAnnotationKey: String? = nil) -> NavigationViewController {
+    func createPdfController(key: String, parentKey: String?, libraryId: LibraryIdentifier, url: URL, page: Int? = nil, preselectedAnnotationKey: String? = nil) -> NavigationViewController {
```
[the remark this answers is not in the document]
michalrentka: Yeah that was a mistake during rebase, it's fixed now.

Zotero/Models/Files.swift
michalrentka: We already delete the directory in `HtmlEpubReaderActionHandler.deinitialise`. But I moved the call from `HtmlEpubDocumentViewController` to `HtmlEpubReaderViewController` which makes more sense. 450ed7e35c728f4f47d6c7199543ec5ad3528e63

scripts/bundle_reader.sh
```diff
@@ -0,0 +1,32 @@
...
+
+SCRIPT_PATH=`realpath "$0"`
+SCRIPT_DIR=`dirname "$SCRIPT_PATH"`
+READER_SUBMODULE_DIR="$SCRIPT_DIR/../reader"
+READER_DIR="$SCRIPT_DIR/../bundled/reader"
+HASH_FILE="$READER_DIR/reader_hash.txt"
+CURRENT_HASH=`git ls-tree --object-only HEAD "$READER_SUBMODULE_DIR"`
+
+if [ -d "$READER_DIR" ]; then
+    if [ -d "$HASH_FILE" ]; then
```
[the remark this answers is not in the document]
michalrentka: Fixed in d62891c9949ec5f5f8005eff6a82832f4dc48244

Zotero/Controllers/Database/Requests/CreateHtmlEpubAnnotationsDbRequest.swift
```diff
@@ -0,0 +1,127 @@
...
+                rField.value = value
+
+            default: break
+            }
+
+            item.fields.append(rField)
+        }
+    }
+
+    private func addTags(for annotation: HtmlEpubAnnotation, to item: RItem, database: Realm) {
```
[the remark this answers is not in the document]
michalrentka: Fixed in cf2f8cf77d7ed670d7dca80d2117edfb99a1a9ac

chenggiant:
Looking forward to the epub and html support!

haidar47x:
Glad to see it being implemented. :)

mvasilak:
@michalrentka some comments from testing ePUB attachments.

### general
* When I download an attachment to open, rarely, it opens empty and I have to close the reader and open it again, for the document to appear. Could there be a potential race issue when downloading, copying to temporary folder, and loading the document?
* For images in the document, e.g. a cover, I can hold-tap to get a context menu, which is nice, unless we think it may cause some confusion.
Wuthering Heights

* I can even select text on an image, which may be confusing if a user tries to annotate it (like I tried 😅).
Wuthering Heights

### Annotations
* Creating an annotation may create 2 annotations with some text overlapping.
Screenshot 2024-10-15 at 15 10 23

* Creating further annotations, may not be rendered, although text is selected, and you get an error like:
> [ERROR] Zotero(+0000001): HtmlEpubDocumentViewController: navigating to 6NREE26N failed - Error Domain=WKErrorDomain Code=4 "A JavaScript exception occurred" UserInfo={WKJavaScriptExceptionLineNumber=31037, WKJavaScriptExceptionMessage=Error: Unsupported FragmentSelector.conformsTo: undefined, WKJavaScriptExceptionColumnNumber=103, WKJavaScriptExceptionSourceURL=file:///.../view.js, NSLocalizedDescription=A JavaScript exception occurred}. [(155) HtmlEpubDocumentViewController.selectInDocument(key:); com.apple.main-thread]
* Closing and re-opening the document after, in both iOS & macOS, will show a blank screen, as it seems corrupted. (No errors are logged in iOS log when re-opening)
* Trying to delete an annotation from the sidebar in this state, logs this error (although the annotation is deleted):
> [ERROR] Zotero(+0000000): HtmlEpubDocumentViewController: updating document failed - Error Domain=WKErrorDomain Code=4 "A JavaScript exception occurred" UserInfo={WKJavaScriptExceptionLineNumber=31037, WKJavaScriptExceptionMessage=Error: Unsupported FragmentSelector.conformsTo: undefined, WKJavaScriptExceptionColumnNumber=103, WKJavaScriptExceptionSourceURL=file:///.../view.js, NSLocalizedDescription=A JavaScript exception occurred}. [(167) HtmlEpubDocumentViewController.updateView(modifications:insertions:deletions:); com.apple.main-thread]
  * Deleting all the annotations (only possible in iOS), eliminates the corrupted state, and re-opening the document displays it as expected.

### Rendering
* It’s not always the same in macOS and iOS, which makes sense, but I may see empty pages in iOS, that I don’t see in macOS.
* Decreasing reader height, e.g. by moving the toolbar to the top, may hide some of the bottom content.
* Changing reader size in any way, e.g. by rotating, will render annotation rects in the wrong place.
Screenshot 2024-10-15 at 15 14 32

### Navigation
* It would be nice to have a table of contents in the sidebar.
* It would be nice to have page controls in the toolbar.
* Tapping a link would be nice to create a back button.

mvasilak:
@michalrentka some comments from testing snapshot attachments.

### General
* Taping an external link doesn't open it in a the browser or an in-app safari view controller. In macOS it is opened in browser.
* Snapshots with existing annotations, perhaps created in previous versions of the macOS client, seem to have breaking changes, resulting in annotations not showing in iOS and logging errors even for selecting an annotation tool. I'm not sure if this will be an issue when we release the feature for iOS.

### Annotations
Creating annotations from text selection is not very accurate, and could result in:
* Missing characters from start or end.
* Adding more characters than selected.
* Drawing an annotation that e.g. underlines more than the annotation's text, usually a whole paragraph.
* Creating more than one annotations at the same time.

These issues seem to stem from the inconsistency of the text selection box and the annotation being drawn, as if two separate text selections are considered in the end. Could be related to web view configuration. 

Check the following example:


https://github.com/user-attachments/assets/90fde1a8-b5a8-481b-a5bf-c32a09bb84d4

michalrentka:
> * For images in the document, e.g. a cover, I can hold-tap to get a context menu, which is nice, unless we think it may cause some confusion.
> 
> Wuthering Heights
> * I can even select text on an image, which may be confusing if a user tries to annotate it (like I tried 😅).
> 
> Wuthering Heights

@dstillman what do you think about this? In my opinion this is actually fine as it's a feature that iOS provides. Yes it might be confusing when you try to select text and annotate it and you don't notice that you're actually selecting text in an image (so you don't have any option to annotate), but I guess that's on you. It should be possible to disable this functionality if you think it's counterproductive though.

michalrentka:
@mvasilak you can try to re-test. Hopefully all or at least most issues should be fixed now. Some were caused by me incorrectly passing the data to reader, others were in the reader itself. I left image handling as is for now.

mvasilak:
@michalrentka things look improved. Some new observations:
### both ePUBS & snapshots
* `onSetSelectionPopup` is ignored if an annotation tool is selected, is this intended? This breaks muscle memory learned from PDF reader.
* Page number label in annotation sidebar cells/popovers, as well as page edit actions, should be omitted, same as in desktop.
* Annotation tools remains open after toolbar has been closed.
### snapshots only
* Selecting an annotation with sidebar open doesn’t scroll to its cell.
* Annotations that cover more than the visible screen, e.g. created in desktop by selecting text and scrolling, may not always show a popover, but still may cause a dismissal event, which w/o presented popover will dismiis the reader.

michalrentka:
@mvasilak should be ready for another look

jarbus:
Guys I think this year is gonna be the year

Timtonix:
Hello, do you have any update for this feature ?

hexclover:
What's the status of this PR? Is there a plan to merge this soon?

scripts/bundle_reader.sh
```diff
@@ -0,0 +1,32 @@
...
+    fi
+
+    if [ $CACHED_HASH == $CURRENT_HASH ]; then
+        exit
+    else
+        rm -rf "$READER_DIR"
+    fi
+fi
+
+cd "$READER_SUBMODULE_DIR"
```
mvasilak, line 28: These are some changes that I also made in the pdf-worker branch, so that Xcode build fails if npm build also does, for any reason, and it should provide a meaningful message.

```suggestion
set -eo pipefail

realpath() {
    [[ $1 = /* ]] && echo "$1" || echo "$PWD/${1#./}"
}

SCRIPT_PATH=`realpath "$0"`
SCRIPT_DIR=`dirname "$SCRIPT_PATH"`
READER_SUBMODULE_DIR="$SCRIPT_DIR/../reader"
READER_DIR="$SCRIPT_DIR/../bundled/reader"
HASH_FILE="$READER_DIR/reader_hash.txt"
CURRENT_HASH=`git ls-tree --object-only HEAD "$READER_SUBMODULE_DIR"`

# Check if the reader submodule is initialized
if ! git -C "$SCRIPT_DIR" submodule status "$READER_SUBMODULE_DIR" | grep -qv '^-'; then
    echo "Error: The reader submodule is not initialized. Run:"
    echo "    git submodule update --init --recursive reader"
    exit 1
fi

if [ -d "$READER_DIR" ]; then
    if [ -f "$HASH_FILE" ]; then
        CACHED_HASH=`cat "$HASH_FILE"`
    else
        CACHED_HASH=0
    fi

    if [ "$CACHED_HASH" == "$CURRENT_HASH" ]; then
        exit
    else
        rm -rf "$READER_DIR"
    fi
fi

cd "$READER_SUBMODULE_DIR"
export NVM_DIR="$HOME/.nvm"
[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
nvm install 21.7.3
```

mvasilak (commented):
When tapping an outline link, it fails with a JS error in the console.

michalrentka:
> When tapping an outline link, it fails with a JS error in the console.

Should be fixed now in reader commit.

> Highlight/Underline annotations allow type switch, but it doesn't seem to actually change the annotation, or even store the change at that.

Added this functionality

> When the width of the reader decreases, sometimes all rendering of text disappears, and only annotation colors are rendered. E.g. have an iPad in portrait mode with side bar open, or have an iPad with another app in split view, in equal widths.

I can't really reproduce this yet

mvasilak:
@michalrentka please also update the `reader` submodule to point to the proper commit.

michalrentka:
> @michalrentka please also update the `reader` submodule to point to the proper commit.

Sorry, forgot to push latest change. Should be there now.

jd-santos:
I no longer have an excuse to not try Zotero, great job team!

jarbus:
Hallelujah, all that's left is the app update

adrian-fuertes:
This is huge! Congrats!!

zsrobinson:
thank you guys so much :)

girdzijauskas:
It's in the beta! Well done!

reagle:
I have 1.0.51 on iPad. Is there a date or version where this feature will be available on that platform?

reagle:
I have zotero 1.0.54 on desktop and ipad, and on the ipad, when I click on an ePub I get the pop-up asking me which app to share/open it in. When is this likely to be released?

PR #843, opened 2024-02-09
Improve code

mvasilak:


Zotero/Scenes/Detail/PDF/Views/PDFReaderViewController.swift
```diff
@@ -764,83 +764,84 @@ extension PDFReaderViewController: PDFDocumentDelegate {
...
     }
 
     func interfaceVisibilityDidChange(to isHidden: Bool) {
-        let shouldChangeNavigationBarVisibility = !self.toolbarState.visible || self.toolbarState.position != .pinned
+        let shouldChangeNavigationBarVisibility = !toolbarState.visible || toolbarState.position != .pinned
 
-        if !isHidden && shouldChangeNavigationBarVisibility && self.navigationController?.navigationBar.isHidden == true {
-            self.navigationController?.setNavigationBarHidden(false, animated: false)
-            self.navigationController?.navigationBar.alpha = 0
+        if !isHidden, shouldChangeNavigationBarVisibility, navigationController?.navigationBar.isHidden == true {
```
michalrentka, line 785: Why change from `&&` to `,`? I'd vote for keeping normal logic operators (`&&`, `||`) and use `,` only when actually needed.

Zotero/Scenes/Detail/PDF/Views/PDFReaderViewController.swift
```diff
@@ -764,83 +764,84 @@ extension PDFReaderViewController: PDFDocumentDelegate {
...
-            guard finished && shouldChangeNavigationBarVisibility else { return }
-            self.navigationController?.setNavigationBarHidden(isHidden, animated: false)
+        }, completion: { [weak self] finished in
+            guard let self, finished, shouldChangeNavigationBarVisibility else { return }
+            navigationController?.setNavigationBarHidden(isHidden, animated: false)
         })
 
-        if isHidden && self.isSidebarVisible {
-            self.toggleSidebar(animated: true)
+        if isHidden, isSidebarVisible {
```
michalrentka, line 805: Same as above.

Zotero/Scenes/Detail/PDF/Views/PDFReaderViewController.swift
```diff
@@ -764,83 +764,84 @@ extension PDFReaderViewController: PDFDocumentDelegate {
...
+            toggleSidebar(animated: true)
         }
     }
 }
 
 extension PDFReaderViewController: ConflictViewControllerReceiver {
     func shows(object: SyncObject, libraryId: LibraryIdentifier) -> String? {
-        guard object == .item && libraryId == self.viewModel.state.library.identifier else { return nil }
-        return self.viewModel.state.key
+        guard object == .item, libraryId == viewModel.state.library.identifier else { return nil }
```
michalrentka, line 813: Same as above

michalrentka (approved):

Zotero/Scenes/Detail/PDF/Views/PDFReaderViewController.swift
mvasilak: For clauses with only and operators, this seems easier to read for me, but I understand there might be cases where a logic operator is preferred. I suggest that we always use `,` for `guard` statements, and for `if` statements that contain `let` clauses. WDYT?

mvasilak: No strong feelings on my side, so to be on the same side, only let clauses, may be separated by `,` and there is no need to re-style existing cases using `&&`.

mvasilak: Check https://github.com/zotero/zotero-ios/compare/c00971089803a374cdf9edf03132eeab0663457a..4293218569cd0a3852646beda65ed98e02353d22

Zotero/Scenes/Detail/PDF/Views/PDFReaderViewController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/compare/c00971089803a374cdf9edf03132eeab0663457a..4293218569cd0a3852646beda65ed98e02353d22

Zotero/Scenes/Detail/PDF/Views/PDFReaderViewController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/compare/c00971089803a374cdf9edf03132eeab0663457a..4293218569cd0a3852646beda65ed98e02353d22

michalrentka (approved):

PR #844, opened 2024-02-12
Citation popup redesign

### LINKED ISSUE #830: Redesign Copy Citation modal screen
Look the conversation starting at https://github.com/zotero/zotero-ios/pull/827#issuecomment-1875626227 for more info.

michalrentka:
Closes #830


Zotero/Scenes/Detail/SingleCitation/Views/CitationLocatorContentView.swift
```diff
@@ -0,0 +1,87 @@
...
+    }
+}
+
+extension CitationLocatorContentView: UITextFieldDelegate {
+    func textField(_ textField: UITextField, shouldChangeCharactersIn range: NSRange, replacementString string: String) -> Bool {
+        guard let text = textField.text as? NSString else { return true }
+        let newString = text.replacingCharacters(in: range, with: string)
+        valueChanged?(newString)
+        return true
+    }
```
mvasilak, line 81: This change misses the debounce from the previous implementation. Couldn't this instance just use the previous react observation mechanism to do the same?

Zotero/Scenes/Detail/SingleCitation/Views/SingleCitationViewController.swift
```diff
@@ -164,64 +236,66 @@ final class SingleCitationViewController: UIViewController {
...
+        preferredContentSize = CGSize(width: width, height: height)
         navigationController?.preferredContentSize = preferredContentSize
     }
 }
 
 extension SingleCitationViewController: WKScriptMessageHandler {
     func userContentController(_ userContentController: WKUserContentController, didReceive message: WKScriptMessage) {
         guard message.name == "heightHandler", let height = message.body as? CGFloat else { return }
-        webViewHeight.constant = height
+        viewModel.process(action: .setPreviewHeight(height))
```
mvasilak, line 270: In my tests, the `height` value is always the same `20`. Can you check if there is an issue here?

mvasilak (changes requested):
Some points that need clarification. Furthermore, I just noticed that when a locator is used, the comma is before the title end quote. Not sure if this is valid, or it is a citation logic bug.

Zotero/Scenes/Detail/SingleCitation/Views/CitationLocatorContentView.swift
michalrentka: Fixed in dc6833d8a4bb2131c12dec7d81d0de20f80ee190

Zotero/Scenes/Detail/SingleCitation/Views/SingleCitationViewController.swift
michalrentka: You're right, height was not calculated properly. The web view which did the citation didn't have the same width as web view that's used to show the citation, so the height differed. I also synced their styles so that the same fonts are used in both. Fixed in 1d2d7ab521af4fb12717934f4d8cdfb22b7e657b.

mvasilak (approved):

PR #850, opened 2024-02-19
Fix initial caching of annotation comment not yet visible in the sidebar

mvasilak:
Fixes regression where only annotation comments in the visible cells of the sidebar, are initially cached.
This leads to opening an annotation popover not cached, without an empty comment value.

### DIFF
--- a/Zotero/Scenes/Detail/PDF/PDFCoordinator.swift
+++ b/Zotero/Scenes/Detail/PDF/PDFCoordinator.swift
@@ -56,7 +56,6 @@ protocol PdfAnnotationsCoordinatorDelegate: AnyObject {
         userInterfaceStyle: UIUserInterfaceStyle,
         completed: @escaping (AnnotationsFilter?) -> Void
     )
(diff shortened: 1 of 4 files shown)


Zotero/Scenes/Detail/PDF/PDFCoordinator.swift
```diff
@@ -583,6 +584,18 @@ extension PDFCoordinator: PdfAnnotationsCoordinatorDelegate {
...
+        let comment = annotation.comment
+        guard !comment.isEmpty else { return nil }
+
+        if let attributedComment = viewModel.state.comments[annotation.key] {
+            return attributedComment
+        }
+
+        viewModel.process(action: .parseAndCacheComment(key: annotation.key, comment: comment))
+        return viewModel.state.comments[annotation.key]
+    }
```
michalrentka, line 598: I agree this needs to be shared, but it doesn't really make sense to have this in `*Coordinator` class. I'd either move it to the view model or `PDFReaderViewController` or some other shared controller.

mvasilak: Check https://github.com/zotero/zotero-ios/pull/850/commits/6efff2582f1a7c0d1f857821c57922aaffb54454

michalrentka (approved):

PR #863, opened 2024-03-01
Option to download files at sync time

### LINKED ISSUE #220: Option to download files at sync time
https://forums.zotero.org/discussion/88303/feature-request-ios-beta-option-to-download-all-files-at-sync-time

I guess some people will want this as an option. As I mention there, I'm not sure we want to do this before the end of the beta, since getting people onto the App Store version without deleting local data may be tricky, and redownloading gigabytes of files would be somewhat annoying.

We'll want this to be a per-library option, with new libraries defaulting to "as needed".

On a low-storage condition, I guess we just stop downloading and show a sync error?

michalrentka:
Closes #220


mvasilak (commented):
Code looks good so far, a couple of questions from testing.
* Snapshots are not download as well. Is this planned for the future?
* Sync seems to be stuck, even after all files are downloaded.

Issue #873, opened 2024-03-12
Modal not closed when deleting annotation on iPhone


dstillman:
https://forums.zotero.org/discussion/112723/zotero-beta-ios-deleting-annotation-should-dismiss-dialog

This must be a recent regression?

PR #875, opened 2024-03-12
Add PDFReaderState change for selection deletion

mvasilak:
When annotation is deleted in its popover, it is not dismissed and deselected in the document.
Happens for both document and sidebar popovers.

Fixes https://github.com/zotero/zotero-ios/issues/873

### DIFF
--- a/Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
+++ b/Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
@@ -993,7 +993,6 @@ final class PDFReaderActionHandler: ViewModelActionHandler, BackgroundDbProcessi
         state.changes.insert(.selection)
 
         guard let key else {
-            state.changes.insert(.selectionDeletion)
             state.selectedAnnotationKey = nil
             return
         }
@@ -1990,6 +1989,7 @@ final class PDFReaderActionHandler: ViewModelActionHandler, BackgroundDbProcessi
             if let key = selectKey {
                 self._select(key: key, didSelectInDocument: true, state: &state)
             } else if selectionDeleted {
+                state.changes.insert(.selectionDeletion)
                 self._select(key: nil, didSelectInDocument: true, state: &state)
             }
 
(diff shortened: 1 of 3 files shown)


michalrentka:
Hmm, it feels like the initial thing should have worked too, did you check where it fails now?

mvasilak:
The previous check
```
            if (presentedViewController as? UINavigationController)?.viewControllers.first is AnnotationPopover, let key = state.selectedAnnotationKey, !state.sortedKeys.contains(key) {
```
fails always because `state.selectedAnnotationKey` is already `nil`. Probably there is a more elegant way to fix this.

Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
```diff
@@ -992,7 +992,8 @@ final class PDFReaderActionHandler: ViewModelActionHandler, BackgroundDbProcessi
 
         state.changes.insert(.selection)
 
-        guard let key = key else {
+        guard let key else {
+            state.changes.insert(.selectionDeletion)
```
michalrentka, line 996: This line doesn't mean that selection has been deleted, but that selected annotation was unselected (by hand gesture, deletion or whatever reason). I'd move this to the end of  `update(objects:deletions:insertions:modifications:viewModel:)` where you have flag `selectionDeleted`.

Zotero/Scenes/Detail/PDF/Views/PDFReaderViewController.swift
```diff
@@ -411,9 +411,9 @@ class PDFReaderViewController: UIViewController {
             viewModel.process(action: .loadDocumentData(boundingBoxConverter: documentController))
         }
 
-        if state.changes.contains(.annotations) {
+        if state.changes.contains(.annotations) && state.changes.contains(.selectionDeletion) {
```
michalrentka, line 414: We can just leave `if state.changes.contains(.selectionDeletion) {` here.

michalrentka (changes requested):

Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/875/commits/e4f821f0e933122b71fd926b16252e8bf46f3f44

Zotero/Scenes/Detail/PDF/Views/PDFReaderViewController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/875/commits/e4f821f0e933122b71fd926b16252e8bf46f3f44

michalrentka (approved):

PR #883, opened 2024-03-20
Add privacy manifest

mvasilak:
Resources:
* https://mastodon.social/@chockenberry/112095474802156488
* https://developer.apple.com/documentation/bundleresources/privacy_manifest_files#4284009
* https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api

### DIFF
--- a/Zotero/PrivacyInfo.xcprivacy
+++ b/Zotero/PrivacyInfo.xcprivacy
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>NSPrivacyAccessedAPITypes</key>
+	<array>
+		<dict>
+			<key>NSPrivacyAccessedAPIType</key>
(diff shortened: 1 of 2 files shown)


Zotero/PrivacyInfo.xcprivacy
```diff
@@ -0,0 +1,46 @@
...
+			</array>
+		</dict>
+		<dict>
+			<key>NSPrivacyAccessedAPIType</key>
+			<string>NSPrivacyAccessedAPICategorySystemBootTime</string>
+			<key>NSPrivacyAccessedAPITypeReasons</key>
+			<array>
+				<string>35F9.1</string>
+				<string>8FFB.1</string>
+				<string>3D61.1</string>
```
dstillman, line 32: What do we use this for? Just for a crash report that includes system uptime?

mvasilak: We may not use them directly, but through 3rd party SDKs. Given that, since we don't have to rush this, we can update first those SDKs that implement it, and see if that silences the complaint.

User defaults is the one that we certainly use.

mvasilak:
We have updated external dependencies, so that most have added their own privacy manifest. I've also used this [tool](https://github.com/stelabouras/privacy-manifest) to scan our code, and only keep those APIs that we do use. In any case, we'll know if more action is required, upon the next app store submission.

Issue #882, opened 2024-03-19
Share extension causing updates to be skipped?


dstillman:
This looks a lot like the share extension is advancing the library version without actually downloading objects.

```
api.zotero.org x.x.x.x - - [03/Mar/2024:11:59:08 -0500] "GET /users/xxxxxx/collections?format=versions&since=6887 HTTP/1.1" 200 2 "-" "ZShare/1.0.24 (org.zotero.ios.Zotero.ZShare; build:222; iOS 17.3.0) Alamofire/5.8.0"
api.zotero.org x.x.x.x - - [03/Mar/2024:11:59:08 -0500] "GET /users/xxxxxx/groups?format=versions HTTP/1.1" 200 2 "-" "ZShare/1.0.24 (org.zotero.ios.Zotero.ZShare; build:222; iOS 17.3.0) Alamofire/5.8.0"
api.zotero.org x.x.x.x - - [03/Mar/2024:11:59:16 -0500] "POST /users/xxxxxx/items/CIZHEBTQ/file HTTP/1.1" 200 12 "-" "ZShare/1.0.24 (org.zotero.ios.Zotero.ZShare; build:222; iOS 17.3.0) Alamofire/5.8.0"
api.zotero.org x.x.x.x - - [03/Mar/2024:11:59:15 -0500] "POST /users/xxxxxx/items HTTP/1.1" 200 548 "-" "ZShare/1.0.24 (org.zotero.ios.Zotero.ZShare; build:222; iOS 17.3.0) Alamofire/5.8.0"

api.zotero.org x.x.x.x - - [03/Mar/2024:11:59:21 -0500] "GET /users/xxxxxx/groups?format=versions HTTP/1.1" 200 2 "-" "Zotero/1.0.24 (org.zotero.ios.Zotero; build:222; iOS 17.3.0) Alamofire/5.8.0"
api.zotero.org x.x.x.x - - [03/Mar/2024:11:59:21 -0500] "GET /users/xxxxxx/collections?format=versions&since=7195 HTTP/1.1" 200 2 "-" "Zotero/1.0.24 (org.zotero.ios.Zotero; build:222; iOS 17.3.0) Alamofire/5.8.0"
[...]

PR #884, opened 2024-03-20
Fix share extension sync

mvasilak:
Syncs item versions as well, when creating download actions for a collections only type of sync. Fixes the edge case mentioned in #882


dstillman:
Can you explain why this is the fix for #882? It doesn't seem like there should be any need to do an item versions request (assuming that's what `syncVersions` does) whenever the share extension is used. The problem is specific to file uploads, so why wouldn't the fix be specific to that? Is there not just a way to have the file-upload request not advance the library version, the same way that the item data request already isn't advancing the version?

mvasilak:
@dstillman improved as you suggested.

dstillman:
That looks better. Let's make a build from this and I'll test.

PR #887, opened 2024-03-26
Improve Report ID submission

michalrentka:


Zotero/Scenes/AppCoordinator.swift
```diff
@@ -677,6 +678,14 @@ extension AppCoordinator: CrashReporterCoordinator {
         }
 
         showAlert(title: L10n.Settings.CrashAlert.title, message: L10n.Settings.CrashAlert.message(id), actions: actions)
+
+        func submit(reportId: String, coordinator: AppCoordinator, completion: @escaping () -> Void) {
+            guard let viewController, var components = URLComponents(string: "https://forums.zotero.org/post/discussion") else { return }
+            components.queryItems = [URLQueryItem(name: "name", value: "iOS Crash Report: \(reportId)"), URLQueryItem(name: "body", value: "[Describe what you were doing when the crash occurred.]")]
```
mvasilak, line 684: Should we make these strings localizable as well, or it doesn't matter?

mvasilak (approved):

Zotero/Scenes/AppCoordinator.swift
michalrentka: Since this is posted to forums I didn't localize it, but @dstillman what do you think?

PR #890, opened 2024-03-27
Added full sync debugging to settings

michalrentka:


Zotero/Controllers/Sync/SyncScheduler.swift
```diff
@@ -70,6 +72,21 @@ final class SyncScheduler: SynchronizationScheduler, WebSocketScheduler {
     }
 
     var inProgress: BehaviorRelay<Bool>
+    var syncTypeInProgressObservable: Observable<SyncController.Kind?> {
+        return inProgress.map({ [weak self] progress in
```
mvasilak, line 76: Why return a new observable instance for each access to this property, rather than e.g. initializing just one lazily?

Zotero/Scenes/AppCoordinator.swift
```diff
@@ -463,24 +463,23 @@ extension AppCoordinator: DebugLoggingCoordinator {
             }
 
             func share(debugId: String, customMessage: String?, userId: Int) {
-                var actions = [
+                let actions = [
                     UIAlertAction(title: L10n.ok, style: .cancel, handler: nil),
-                    UIAlertAction(title: L10n.copy, style: .default, handler: { _ in
+                    UIAlertAction(title: L10n.Settings.CrashAlert.submitForum, style: .default, handler: { _ in
                         UIPasteboard.general.string = debugId
```
mvasilak, line 469: The `debugId` will be copied to paste board only if the user tries to submit to the forum. It might be better to be copied every time, i.e. just before showing the alert, regardless of the action selected.

mvasilak (approved):
A couple of minor comments.

Zotero/Controllers/Sync/SyncScheduler.swift
michalrentka: Fixed in d219e47b67c93d92933ce16a90215d5bb83168c3

Zotero/Scenes/AppCoordinator.swift
michalrentka: Makes sense, changed in d219e47b67c93d92933ce16a90215d5bb83168c3

PR #904, opened 2024-04-12
Page spreads implemented

### LINKED ISSUE #900: Odd/Even Spreads option for Page Mode
https://forums.zotero.org/discussion/107448/ios-page-spreads

Are we able to do this?

michalrentka:
Closes #900

### DIFF
--- a/Zotero/Scenes/Detail/PDF/Models/PDFSettings.swift
+++ b/Zotero/Scenes/Detail/PDF/Models/PDFSettings.swift
@@ -17,15 +17,24 @@ struct PDFSettings {
     var pageFitting: PDFConfiguration.SpreadFitting
     var appearanceMode: ReaderSettingsState.Appearance
     var idleTimerDisabled: Bool
+    var isFirstPageAlwaysSingle: Bool
 
     static var `default`: PDFSettings {
-        return PDFSettings(transition: .scrollContinuous, pageMode: .automatic, direction: .horizontal, pageFitting: .adaptive, appearanceMode: .automatic, idleTimerDisabled: false)
+        return PDFSettings(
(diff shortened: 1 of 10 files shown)


Zotero/Scenes/Detail/PDF/Models/PDFSettings.swift
```diff
@@ -35,12 +44,14 @@ extension PDFSettings: Codable {
         let appearanceRaw = try container.decode(UInt.self, forKey: .appearanceMode)
         let modeRaw = (try? container.decode(UInt.self, forKey: .pageMode)) ?? 2
         let fittingRaw = (try? container.decode(Int.self, forKey: .pageFitting)) ?? 2
+        let isFirstPageAlwaysSingle = (try? container.decode(Bool.self, forKey: .isFirstPageAlwaysSingle)) ?? false
```
mvasilak, line 47: We should use the PSPDFKIt default here, which is `true`.

mvasilak (commented):
One minor change. 
I undestand that `Even` (`true`) is `Even-Odd` and `Odd` (`false`) is `Odd-Even`.
Could the full names be more clear for users?

michalrentka:
> One minor change. I undestand that `Even` (`true`) is `Even-Odd` and `Odd` (`false`) is `Odd-Even`. Could the full names be more clear for users?

@dstillman what do you think?

Zotero/Scenes/Detail/PDF/Models/PDFSettings.swift
michalrentka: You're right, I got that mixed up in here, got it right just few lines above 😅

PR #907, opened 2024-04-22
Fix WebDAV not found error handling

### LINKED ISSUE #906: "Error: Unable to unzip snapshot" if file is missing on WebDAV server?
https://forums.zotero.org/discussion/comment/461279/#Comment_461279

mvasilak:
Closes #906

### DIFF
--- a/Zotero/Controllers/Attachment Downloader/AttachmentDownloader.swift
+++ b/Zotero/Controllers/Attachment Downloader/AttachmentDownloader.swift
@@ -796,10 +796,26 @@ extension AttachmentDownloader: URLSessionDownloadDelegate {
             return
         }
 
+        var error: Swift.Error?
+        do {
+            if let fileResponseError = checkFileResponse(for: Files.file(from: location), fileStorage: fileStorage) {
+                throw fileResponseError
+            }
+        } catch let fileResponseError {
+            error = fileResponseError
+        }
+
         if let data = activeDownload.logData {
-            logResponse(for: data, task: downloadTask, error: nil)
+            logResponse(for: data, task: downloadTask, error: error)
         }
         DDLogInfo("AttachmentDownloader: didFinishDownloadingTo \(downloadTask.taskIdentifier)")
+
+        if let error {
+            accessQueue.sync(flags: .barrier) { [weak self] in
+                self?.finish(activeDownload: activeDownload, download: download, compressed: nil, result: .failure(error))
+            }
+            return
+        }
 
         var zipFile: File?
         var shouldExtractAfterDownload = activeDownload.extractAfterDownload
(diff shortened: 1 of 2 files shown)


Zotero/Controllers/Attachment Downloader/AttachmentDownloader.swift
```diff
@@ -796,11 +796,27 @@ extension AttachmentDownloader: URLSessionDownloadDelegate {
             return
         }
 
+        var error: Swift.Error?
+        do {
```
michalrentka, line 800: No need to have this in `do catch` block, just have something like
```
let error = checkFileResponse(for: Files.file(from: location), fileStorage: fileStorage, downloadTask: downloadTask)

if let data = activeDownload.logData {
    logResponse(for: data, task: downloadTask, error: error)
}
...
if let error {
    ...
}
```

michalrentka (changes requested):

Zotero/Controllers/Attachment Downloader/AttachmentDownloader.swift
mvasilak: Correct, I just merged this, I'll just push another commit to improve this.

Zotero/Controllers/Attachment Downloader/AttachmentDownloader.swift
```diff
@@ -848,11 +861,17 @@ extension AttachmentDownloader: URLSessionDownloadDelegate {
             }
         }
 
-        func checkFileResponse(for file: File, fileStorage: FileStorage) -> Swift.Error? {
+        func checkFileResponse(for file: File, fileStorage: FileStorage, downloadTask: URLSessionDownloadTask) -> Swift.Error? {
             let size = fileStorage.size(of: file)
-            if size == 0 || (size == 9 && (try? fileStorage.read(file)).flatMap({ String(data: $0, encoding: .utf8) }) == "Not found") {
+            if size == 0 || (size == 9 && (try? fileStorage.read(file)).flatMap({ String(data: $0, encoding: .utf8) })?.caseInsensitiveCompare("Not found") == .orderedSame) {
```
michalrentka, line 866: And while you're fixing the code above, you can extract this bit into a standalone function and use it both here and in `RemoteAttachmentDownloadOperation`

Zotero/Controllers/Attachment Downloader/AttachmentDownloader.swift
mvasilak: Check https://github.com/zotero/zotero-ios/commit/919a1d40e913e3b4182d0bc76cc0bbf4bab870f6 (commited directly)

Zotero/Controllers/Attachment Downloader/AttachmentDownloader.swift
mvasilak: Check https://github.com/zotero/zotero-ios/commit/919a1d40e913e3b4182d0bc76cc0bbf4bab870f6 (commited directly)

PR #909, opened 2024-04-24
Remove unnecessary custom auth challenge delegate implementation

michalrentka:


Zotero/Controllers/API/ZoteroApiClient.swift
```diff
@@ -64,20 +64,24 @@ final class ZoteroApiClient: ApiClient {
...
+    }
+
+    func set(credentials: (String, String)?) {
+        guard let (username, password) = credentials else {
+            token = nil
+            return
         }
+
+        token = .credentials(username: username, password: password)
+        (manager.delegate as? CredentialSessionDelegate)?.credential = URLCredential(user: username, password: password, persistence: .forSession)
```
mvasilak, line 84:
```suggestion
        (manager.delegate as? CredentialSessionDelegate)?.credential = self.credentials
```

Zotero/Controllers/Attachment Downloader/AttachmentDownloader.swift
```diff
@@ -896,6 +896,39 @@ extension AttachmentDownloader: URLSessionDownloadDelegate {
...
+        if urlString.starts(with: ApiConstants.baseUrlString) {
+            // Perform default handling because Zotero uses auth token instead of credentials
+            completionHandler(.performDefaultHandling, nil)
+            return
+        }
+
+        let webDavUrl = webDavController.sessionStorage.scheme.rawValue + "://" + webDavController.sessionStorage.url
+        if !urlString.starts(with: webDavUrl) {
+            // Perform default handling if it's not WebDAV
+            completionHandler(.performDefaultHandling, nil)
```
mvasilak, line 917: Why it doesn't return here?

mvasilak (approved):

dstillman:
Before we merge this, I'd like to understand what this is fixing, and ideally have a test — or at least a Debug ID, but preferably a test — that shows it failing before. We've had a bunch of WebDAV problems lately, but WebDAV downloads are certainly not generally broken, so I'd like to make sure we're clear on what the issue here is.

PR #912, opened 2024-04-26
Set scene title

mvasilak:
Suggested in this forum [post](https://forums.zotero.org/discussion/comment/461420/#Comment_461420)

### DIFF
--- a/Zotero/Scenes/Main/Views/MainViewController.swift
+++ b/Zotero/Scenes/Main/Views/MainViewController.swift
@@ -32,9 +32,7 @@ final class MainViewController: UISplitViewController {
     private var detailCoordinator: DetailCoordinator? {
         didSet {
             guard let detailCoordinator else { return }
-            let userActivity: NSUserActivity = .mainActivity
-            userActivity.title = detailCoordinator.collection.name
-            set(userActivity: userActivity)
+            set(userActivity: .mainActivity.set(title: detailCoordinator.collection.name))
             if let detailCoordinatorGetter {
                 detailCoordinatorGetter(detailCoordinator)
                 self.detailCoordinatorGetter = nil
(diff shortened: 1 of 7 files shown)


mvasilak:
> We should probably also provide a title for main activity? Like "Home" or "Items", not sure.

For the time being I left it `nil`, that will also not set the scene title, resulting to the default which is the app name. 
We can consider using the current collection as the name.

dstillman:
Collection name makes sense to me. We change the first tab — and therefore the window title — to the selected collection in the desktop app.

Zotero/Scenes/Main/Views/MainViewController.swift
```diff
@@ -75,8 +78,10 @@ final class MainViewController: UISplitViewController {
 
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
-        self.set(userActivity: .mainActivity)
-        self.didAppear = true
+        guard let detailCoordinator else { return }
+        let userActivity: NSUserActivity = .mainActivity
```
michalrentka, line 82: You could maybe create `NSUserActivity` extension, something like: `func create(activity: NSUserActivity, title: String) -> NSUserActivity` so that we can just do `set(userActivity: .create(activity: .mainActivity, title: detailCoordinator.collection.name)` instead of 3 lines every time. But also we only have this in 2 places now, it's fine as it is too. So up to you :).

michalrentka (approved):

Zotero/Scenes/Main/Views/MainViewController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/912/commits/e08d4039b87fe839bd332797a73dc22687c92343

PR #913, opened 2024-04-27
Explicitly call viewIsAppearing for nested view controllers in iOS < 16

mvasilak:
Fixes this user reported [issue](https://forums.zotero.org/discussion/comment/461590/#Comment_461590), that is caused by an iOS [bug](https://forums.developer.apple.com/forums/thread/749169) that doesn't call `viewIsAppearing` for nested view controllers in iOS 15.

### DIFF
--- a/Zotero/Scenes/Detail/PDF/Views/PDFDocumentViewController.swift
+++ b/Zotero/Scenes/Detail/PDF/Views/PDFDocumentViewController.swift
@@ -45,7 +45,6 @@ final class PDFDocumentViewController: UIViewController {
     }
     private var searchResults: [SearchResult] = []
     private var pageIndexCancellable: AnyCancellable?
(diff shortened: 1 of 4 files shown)


michalrentka:
Wouldn't it be better to call `viewIsAppearing` from parent `UIViewController` for all child controllers?

So parent controller would look like:
```
func viewIsAppearing(animated: Bool) {
    for child in childViewControllers {
        child.viewIsAppearing(animated: animated)
    }
}
```

And then child controllers would just stay the same.

michalrentka (approved):

mvasilak:
@michalrentka very good suggestion, it also helped catch an omission in the previous solution. Check https://github.com/zotero/zotero-ios/pull/913/commits/cd3cb229ddd9dcb41e2f8b2b6ddcb17fcfdc5cd1

PR #915, opened 2024-04-29
Fix save of local file when over quota limit

### LINKED ISSUE #910: Locally shared file not being saved to app when over storage quota?
https://forums.zotero.org/discussion/comment/461466/#Comment_461466

I haven't tested, but this looks like we may not be saving the file to the app when the share extension gets a quota error from the API?

mvasilak:
Closes #910

### DIFF
--- a/ZShare/ViewModels/ExtensionViewModel.swift
+++ b/ZShare/ViewModels/ExtensionViewModel.swift
@@ -1184,8 +1184,7 @@ final class ExtensionViewModel {
                 return alamoErrorRequiresAbort(.responseValidationFailed(reason: .unacceptableStatusCode(code: response.code)), url: nil, libraryId: libraryId)
 
             case .authorizationFailed(let statusCode, let response, _):
-                guard statusCode == 413 else { return .unknown }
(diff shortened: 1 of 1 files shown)


ZShare/ViewModels/ExtensionViewModel.swift
```diff
@@ -1183,6 +1183,10 @@ final class ExtensionViewModel {
                 guard let response = responses.first else { return .unknown }
                 return alamoErrorRequiresAbort(.responseValidationFailed(reason: .unacceptableStatusCode(code: response.code)), url: nil, libraryId: libraryId)
 
+            case .authorizationFailed(let statusCode, let response, _):
+                guard statusCode == 413 else { return .unknown }
```
michalrentka, line 1187: Do we want to limit it to 413? Why not just pass status code for all authorization failures?

michalrentka (approved):

ZShare/ViewModels/ExtensionViewModel.swift
mvasilak: I guess it doesn't hurt. Check https://github.com/zotero/zotero-ios/pull/915/commits/bfb065874c193df2175909eefbf9f7107963eddf

PR #916, opened 2024-04-29
Re-added copy debug id button

### LINKED ISSUE #911: Add/restore "Copy to Clipboard" button when generating Debug ID
https://forums.zotero.org/discussion/comment/461466/#Comment_461466

I can't remember what we had before, but I assume we had a way to copy the Debug ID to the clipboard? We still need that — if someone has an existing forum thread, "Submit to Zotero Forums" isn't helpful.

michalrentka:
Closes #911


mvasilak:
Why can't we have both actions, i.e. copy to clipboard and submit to forums?

mvasilak (approved):

michalrentka:
> Why can't we have both actions, i.e. copy to clipboard and submit to forums?

We do, I guess you're referring to the deleted string. We had the same string listed twice for some reason, so I re-used one, you can see "Submit to Zotero Forums" is still available 2 lines above it.

Issue #880, opened 2024-03-18
Highlight annotations created in iOS may contain gaps in drawing


mvasilak:
https://forums.zotero.org/discussion/comment/458539
The issue appeared when we upgraded PSPDFKit from `13.3.1` to `13.3.2`.

Before:
<img width="178" alt="Selection-13 3 1" src="https://github.com/zotero/zotero-ios/assets/617841/a7666825-4c19-48a4-a4ca-ba799cad016e">
<img width="162" alt="Highlight-13 3 1" src="https://github.com/zotero/zotero-ios/assets/617841/144d8f11-1cbc-45ac-b318-c576f277220c">

After:
<img width="178" alt="Selection-13 3 2" src="https://github.com/zotero/zotero-ios/assets/617841/9fd815f5-f626-49ed-972b-cc17d4651f85">
<img width="166" alt="Highlight-13 3 2" src="https://github.com/zotero/zotero-ios/assets/617841/5c376cab-dea9-4f60-96ea-ef9cae189367">

I'll contact their support.

mvasilak:
In the meantime, should we revert to the version where the issue does not appear and release a new version? WDYT @dstillman  @michalrentka ?

dstillman:
Yes, it seems like we're going to get a lot of reports on this. We should revert if that won't cause other problems.

mvasilak:
PSPDFKit confirmed it's a regression on their part. As we have reverted to the older working version, we'll update the dependency once this is fixed.

michalrentka:
@mvasilak how is this looking? PSDPFKit 13.4.1 is out, does it still contain the issue?

mvasilak:
Still waiting, their internal ticket for this is `J#CORE-510`, and they are also going to inform me when it is fixed.

mvasilak:
Heard back from PSPDFKit support:
> I'm glad to tell you that the issue has been fixed and a patch will be available with our upcoming release.

Issue #623, opened 2023-02-25
Intra-PDF links should scroll to the exact page position


dstillman:
https://forums.zotero.org/discussion/98115/bug-report-ios-pdf-reader-line-level-linking-broken

(I'm not sure if this info is in the link or @mrtcode is doing something special in the desktop reader to scroll to the correct location.)

michalrentka:
@dstillman `PSPDFKit currently doesn't support executing links that zoom into a specific part of the page.` Unfortunately I can't adjust this behaviour myself and PSPDFKit needs to implement it. They created an internal feature request for it.

michalrentka:
PSPDFKit implemented this feature in 13.4.0 (https://pspdfkit.com/blog/2024/pspdfkit-13-4-ios-multiple-annotation-selection/#linking-to-regions-of-pages) but we're waiting for another update which fixes #880.

PR #919, opened 2024-05-07
Fix PDF document gaps for selection and highlight annotations

mvasilak:
@dstillman @michalrentka the latest release of PSDFKit (13.5.0) improves the highlight annotation gap issue (#880) , but doesn't resolve it completely. It still keeps some gaps between sentences. Should we go ahead and merge it, or stick with 13.3.1 until it's further resolved? (I've already replied in the original ticket we had with PSPDFKit)

Highlight 13.3.1:
highlight 13 3 1
Selection 13.3.1:
selection 13 3 1
Highlight 13.5.0:
highlight 13 5 0
Selection 13.5.0:
selection 13 5 0

### DIFF
--- a/Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
+++ b/Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
@@ -1466,14 +1466,49 @@ final class PDFReaderActionHandler: ViewModelActionHandler, BackgroundDbProcessi
                     continue
                 }
 
-                let splitAnnotations = splitIfNeeded(annotation: annotation)
+                var workingAnnotation = annotation
+                if annotation is HighlightAnnotation, let rects = annotation.rects, rects.count > 1 {
+                    // Check if there are gaps for sequential highlight rects on the same line, and if so transform the annotation to eliminate them.
+                    var mergedRects: [CGRect] = []
+                    for rect in rects {
+                        guard let previousRect = mergedRects.last, rect.minY == previousRect.minY, rect.height == previousRect.height else {
+                            mergedRects.append(rect)
+                            continue
+                        }
+                        let mergedRect = CGRect(x: previousRect.minX, y: previousRect.minY, width: rect.minX + rect.width - previousRect.minX, height: previousRect.height)
+                        mergedRects.removeLast()
+                        mergedRects.append(mergedRect)
+                    }
+                    if mergedRects.count < rects.count {
+                        DDLogInfo("PDFReaderActionHandler: did merge highlight annotation rects")
+                        let newAnnotation = HighlightAnnotation()
+                        newAnnotation.rects = mergedRects
+                        newAnnotation.boundingBox = AnnotationBoundingBoxCalculator.boundingBox(from: rects)
+                        newAnnotation.alpha = annotation.alpha
+                        newAnnotation.color = annotation.color
+                        newAnnotation.blendMode = annotation.blendMode
+                        newAnnotation.contents = annotation.contents
+                        newAnnotation.pageIndex = annotation.pageIndex
+                        toRemove.append(annotation)
+                        toAdd.append(newAnnotation)
+                        workingAnnotation = newAnnotation
+                    }
+                }
+
+                let splitAnnotations = splitIfNeeded(annotation: workingAnnotation)
 
                 guard splitAnnotations.count > 1 else {
-                    keptAsIs.append(annotation)
+                    if workingAnnotation == annotation {
+                        keptAsIs.append(annotation)
+                    }
                     continue
                 }
                 DDLogInfo("PDFReaderActionHandler: did split annotations into \(splitAnnotations.count)")
-                toRemove.append(annotation)
+                if workingAnnotation == annotation {
+                    toRemove.append(annotation)
+                } else {
+                    toAdd.removeLast()
+                }
                 toAdd.append(contentsOf: splitAnnotations)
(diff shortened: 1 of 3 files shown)


dstillman:
I would wait, unless there are other serious issues we're trying to fix. No reason to regress on this unnecessarily.

michalrentka:
> I would wait, unless there are other serious issues we're trying to fix. No reason to regress on this unnecessarily.

@dstillman I'll just mention #623, but I don't think it's worth updating just for that.

mvasilak:
@dstillman @michalrentka I got this update from PSPDFKit:

> I'm getting back to you after a good effort from a few teams to assess the text highlight concern.
> Based on our conversation this was the result of fixing an issue where the text highlighter was too aggressive and we believe now is more adequate to handle most PDFs.
> Please let us know your thoughts/concern.

I'll ask if this is something that can be parameterized or overridden, since we didn't have any issues reported by users, AFAIK, with their previous logic. I'm guessing their motive is different text columns/blocks, that are too close and are selected as one?

In the meantime, I'm also working on merging the highlight annotation `rects`, in our processing logic, that already makes other similar modifications. It's doable, as the two rects have the same `minY` and are sequential in the `rects` list, so it seems a sensible condition. The selection would still be split, but at least the highlight will be continuous. We can check if we can also remedy this afterwards.

Any thoughts?

dstillman:
As long as the new logic works, I don't think there's any need for a parameter — it doesn't seem like something that *should* be parameterized to me.

mvasilak:
@dstillman to be more clear, their new logic creates the gaps shown in the screenshots of the PR description, which they find acceptable. If we are ok with that as well, we can update as is.

dstillman:
Ah, got it. I understand that this is a hard problem in PDFs, but do any other PDF readers have trouble with that sample PDF?

mvasilak:
I tried with 3 different PDF files (which I've shared with PSPDFKit support), all two column articles with justified alignment.
In places where such gaps appear with PSPDFKit 13.5.0, the Preview app for macOS and the system PDF reader for iOS, don't have any issues.

dstillman:
OK. Might be worth checking Acrobat and PDF Expert as well. Not impossible that Preview (which I assume is the same as iOS) is uncommonly good here.

But if those all manage to avoid this and PSPDFKit doesn't, I guess parameterizing would be helpful, at least until they figure out how to improve this while addressing whatever other problem they were trying to fix. Might also be worth asking if they have examples of what they were actually fixing — even if we didn't get reports, if there are real-world problems addressed by this change, it would be interesting to know what they were. Because yes, it's entirely possible they're addressing problems on PDFs with complicated column layouts that our users will almost never encounter.

mvasilak:
I checked those, for both iOS and macOS, testing with the same text. Acrobat macOS does create 1 gap, **but only** for the highlight, not the selection! Furthermore, Acrobat iOS and PDF Expert macOS/iOS work as expected.
acrobat reader macOS highlight

I'll reply to PSPDFKit support with our valid concerns.

mvasilak:
A fix for the resulting highlight annotation is added in the PR, so we have an alternative, if the library logic remains as is. @michalrentka please take a look.

mvasilak:
I got some more feedback from PSPDFKit support, regarding the new logic:

> Previously PSPDFKit assumed any space was sized 10px x 10px - this was “good enough” for quite a while, but is obviously not correct and could lead to reported text bounding boxes being in the wrong location. We now calculate the space characters width correctly — which means if there is a bigger spacing represented by a space, we no longer combine the sentences.
 > 
> We don’t currently have a way to customise how our text parser groups text, but we’re internally discussing if we can improve this further. We’ll get back to you about this.

Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
```diff
@@ -1422,6 +1405,167 @@ final class PDFReaderActionHandler: ViewModelActionHandler, BackgroundDbProcessi
...
+                let (_, _, blendMode) = AnnotationColorGenerator.color(from: activeColor, isHighlight: (annotation is PSPDFKit.HighlightAnnotation), userInterfaceStyle: state.interfaceStyle)
+                annotation.blendMode = blendMode ?? .normal
+
+                // Either annotation is new (key not assigned) or the user used undo/redo and we check whether the annotation exists in DB
+                guard annotation.key == nil || state.annotation(for: .init(key: annotation.key!, type: .database)) == nil else { continue }
+                var workingAnnotation = annotation
+                var toRemove: [PSPDFKit.Annotation] = []
+                var toAdd: [PSPDFKit.Annotation] = []
+
+                if let annotation = annotation as? HighlightAnnotation, let rects = annotation.rects, rects.count > 1 {
```
michalrentka, line 1430: Looks good, I'd just separate thie merging logic into sub-function here, same as splitting.

michalrentka (approved):

mvasilak:
PSPDFKit created another ticket to try improving this issue, so I propse we put it on hold for a little more.

michalrentka:
> PSPDFKit created another ticket to try improving this issue, so I propse we put it on hold for a little more.

We already have a fix implemented though, are there any doubts? If they fix their stuff we can just disable our fix.

mvasilak:
@michalrentka good point, I'll create another PR that establishes some changes for proper undo recording for transformed annotations, and then I will rebase this one. In any case, a beta release will help us get more feedback.

michalrentka (approved):

PR #927, opened 2024-05-15
Regression fixed

michalrentka:
Fixed regression in #924. It didn't render annotations.

### DIFF
--- a/Zotero/Controllers/AnnotationPreviewController.swift
+++ b/Zotero/Controllers/AnnotationPreviewController.swift
@@ -203,7 +203,7 @@ extension AnnotationPreviewController {
 
         let newDocument = Document(url: fileURL)
 
-        if includeAnnotation, let annotation = document.annotations(at: pageIndex).filter({ $0.previewId == key }).first {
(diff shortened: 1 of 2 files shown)


Zotero/Controllers/AnnotationPreviewController.swift
```diff
@@ -199,17 +199,18 @@ extension AnnotationPreviewController {
...
     ) {
-        let document = Document(url: documentURL)
-        var skipAnnotations = document.annotations(at: pageIndex)
-        if includeAnnotation {
-            skipAnnotations = skipAnnotations.filter({ $0.previewId != key })
+        guard let fileURL = document.fileURL else { return }
+
+        let newDocument = Document(url: fileURL)
+
+        if includeAnnotation, let annotation = document.annotations(at: pageIndex).filter({ $0.previewId == key }).first {
```
mvasilak, line 206:
```suggestion
        if includeAnnotation, let annotation = document.annotations(at: pageIndex).first(where: { $0.previewId == key }) {
```

mvasilak (changes requested):

PR #934, opened 2024-05-24
Improve PDF reader action handler

mvasilak:
* Refactors code.
* Fixes bug when annotation is split.
* Pops undo stack and records proper undo command when an added annotation has been transformed.

### DIFF
--- a/Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
+++ b/Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
@@ -1121,28 +1121,6 @@ final class PDFReaderActionHandler: ViewModelActionHandler, BackgroundDbProcessi
 
     // MARK: - Annotation management
 
-    private func tool(from annotation: PSPDFKit.Annotation) -> PSPDFKit.Annotation.Tool? {
-        if annotation is PSPDFKit.HighlightAnnotation {
-            return .highlight
-        }
-        if annotation is PSPDFKit.NoteAnnotation {
-            return .note
-        }
-        if annotation is PSPDFKit.SquareAnnotation {
-            return .square
-        }
-        if annotation is PSPDFKit.InkAnnotation {
-            return .ink
-        }
-        if annotation is PSPDFKit.UnderlineAnnotation {
-            return .underline
-        }
-        if annotation is PSPDFKit.FreeTextAnnotation {
-            return .freeText
-        }
-        return nil
-    }
-
     private func tool(from annotationType: AnnotationType) -> PSPDFKit.Annotation.Tool {
         switch annotationType {
         case .note:
@@ -1401,11 +1379,16 @@ final class PDFReaderActionHandler: ViewModelActionHandler, BackgroundDbProcessi
     /// - parameter annotations: Annotations that were added to the document.
     /// - parameter viewModel: ViewModel.
     private func add(annotations: [PSPDFKit.Annotation], selectFirst: Bool, in viewModel: ViewModel<PDFReaderActionHandler>) {
-        guard let boundingBoxConverter = self.delegate else { return }
+        guard let boundingBoxConverter = delegate else { return }
 
         DDLogInfo("PDFReaderActionHandler: annotations added - \(annotations.map({ "\(type(of: $0));key=\($0.key ?? "nil");" }))")
 
-        let finalAnnotations = self.splitIfNeededAndProcess(annotations: annotations, state: viewModel.state)
+        let finalAnnotations = splitIfNeededAndProcess(
+            annotations: annotations,
+            state: viewModel.state,
+            boundingBoxConverter: boundingBoxConverter,
+            annotationPreviewController: annotationPreviewController
+        )
 
         guard !finalAnnotations.isEmpty else { return }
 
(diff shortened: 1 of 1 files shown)


Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
```diff
@@ -1396,32 +1374,170 @@ final class PDFReaderActionHandler: ViewModelActionHandler, BackgroundDbProcessi
...
 
         DDLogInfo("PDFReaderActionHandler: annotations added - \(annotations.map({ "\(type(of: $0));key=\($0.key ?? "nil");" }))")
 
-        let finalAnnotations = self.splitIfNeededAndProcess(annotations: annotations, state: viewModel.state)
+        let (keptAsIs, toRemove, toAdd) = transformIfNeeded(annotations: annotations, state: viewModel.state)
+        let finalAnnotations = keptAsIs + toAdd
+        if !toRemove.isEmpty || !toAdd.isEmpty {
+            // Originally added annotations are transformed, so we remove them by performing last undo.
+            // This also removes the undo command from the stack, allowing us to record the transformed addition.
+            if viewModel.state.document.undoController.undoManager.canUndo {
```
michalrentka, line 1387: Looks good, but I worry a little, does this happen synchronously? When annotation is added to document, is this called immediately after? Couldn't another annotation be added before this is called so you'd undo different annotation? Is there a way to check that we're undo-ing the annotation we need? Or can we just remove the undo action manually since we're removing annotations below anyway?

Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
```diff
@@ -1396,32 +1374,170 @@ final class PDFReaderActionHandler: ViewModelActionHandler, BackgroundDbProcessi
...
+                    annotation.customData = [AnnotationsConfig.keyKey: KeyGenerator.newKey]
+                }
+
+                return [annotation]
+
+                func createAnnotations(from splitRects: [[CGRect]], original: HighlightAnnotation) -> [HighlightAnnotation] {
+                    guard splitRects.count > 1 else { return [original] }
+                    return splitRects.map { rects -> HighlightAnnotation in
+                        let new = HighlightAnnotation()
+                        new.rects = rects
```
michalrentka, line 1491: I see we're assigning `user` above but not here or below.

Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
```diff
@@ -1737,74 +1752,68 @@ final class PDFReaderActionHandler: ViewModelActionHandler, BackgroundDbProcessi
...
         NotificationCenter.default.rx
             .notification(.PSPDFAnnotationsAdded)
             .observe(on: MainScheduler.instance)
-            .subscribe(onNext: { [weak self, weak viewModel] notification in
-                guard let self, let viewModel else { return }
-                self.processAnnotationObserving(notification: notification, viewModel: viewModel)
-            })
-            .disposed(by: self.pdfDisposeBag)
+            // Delay annotations added notifications, so they have completely recorded their undo command.
+            .delay(.microseconds(250), scheduler: MainScheduler.instance)
```
michalrentka, line 1772: This makes the worry above a bit worse :). What if the delay changes and we're removing wrong annotations?

michalrentka (changes requested):

Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
mvasilak: Did some improvements in https://github.com/zotero/zotero-ios/pull/934/commits/22f4389ed5c9d0d55e0f245b90900d7b76efbb22

It's immediatelly after the undo command is recorded. The sequence is as follows:
- add annotation programmatically in a recorded block
OR
add annotation via `AnnotationStateManager`, process notification with a debounce of `0` milliseconds so that the recording finishes
- `PDFReaderActionHandler` `add` method is called from the notification handler

AFAICS, we can either remove all undo actions, or call undo to remove the last one, and we cannot check programmatically what is in the undo/redo stacks. I researched other ways, like disabling automatic registration at document configuration and do all recordings programmatically, but the already drawn annotation when using the `AnnotationStateManager` causes a  lot of issues. 

We can ask PSPDFKit if they can add an option that passes starts detached recording, and passes the recorder with the notification, which will allow adding more commands to the recording block, and commiting afterwards. Still, there might be some issues I'm not aware of currently.

Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/934/commits/22f4389ed5c9d0d55e0f245b90900d7b76efbb22 I've set the delay to `0`, and it still works. This is what we actually want, dispatch to the next main run loop iteration, so that the recording block of the `AnnotationStateManager` is finished.

michalrentka: Removing this debounce (or well setting it to 0 milliseconds) defeats the comment's purpose a bit. Though text annotations are not enabled yet, so we can maybe keep it like this for now, but at least add a TODO there.

I'd probably also reach to PSDPFKit and explain our situation, maybe they could open the recording API a bit more so that we can search through the queue and remove only some steps manually.

mvasilak: My bad, I changed the wrong debounce :(
Let me fix it and test again.

Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/934/commits/22546ac6f7301f41043768d30b9eef62a5755033

Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/934/commits/22546ac6f7301f41043768d30b9eef62a5755033

Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/934/commits/5d5c17041c254a25430915a8b15114a413eedd4c

Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
```diff
@@ -1737,74 +1757,70 @@ final class PDFReaderActionHandler: ViewModelActionHandler, BackgroundDbProcessi
...
+                    let changes = PdfAnnotationChanges.stringValues(from: [.boundingBox, .paths])
+                    handler.change(annotation: annotation, with: changes, in: viewModel)
+                }
+
+            case .PSPDFAnnotationsAdded:
+                guard let annotations = notification.object as? [PSPDFKit.Annotation] else { return }
+                // TODO: Improve this if PSPDFKit allows more control for automatic command recording, e.g. by providing a detached recorder
+                // Delay handling of added notifications until next main run loop iteration, so they have completely recorded their undo command.
+                DispatchQueue.main.async {
+                    handler.add(annotations: annotations, in: viewModel)
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/934/commits/f114991312ee257e35ea5642ee3d51c8cf3d486f

Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
```diff
@@ -1396,32 +1374,174 @@ final class PDFReaderActionHandler: ViewModelActionHandler, BackgroundDbProcessi
...
 
         DDLogInfo("PDFReaderActionHandler: annotations added - \(annotations.map({ "\(type(of: $0));key=\($0.key ?? "nil");" }))")
 
-        let finalAnnotations = self.splitIfNeededAndProcess(annotations: annotations, state: viewModel.state)
+        let (keptAsIs, toRemove, toAdd) = transformIfNeeded(annotations: annotations, state: viewModel.state)
+        let finalAnnotations = keptAsIs + toAdd
+        for annotation in finalAnnotations {
+            if annotation.key == nil {
+                annotation.user = viewModel.state.displayName
+                annotation.customData = [AnnotationsConfig.keyKey: KeyGenerator.newKey]
```
michalrentka, line 1387: You shouldn't assign a new key to all "final" annotations. The ones that you re-do already have a key. So if you move it here I'd check for a key first and assign only to actually newly created ones.

mvasilak: Isn't the clause `if annotation.key == nil` adequate?

michalrentka (approved):

mvasilak:
I got this feedback from PSPDFKit regarding this solution:

> Currently it is not possible to acquire the undo recorder that records the action to add the annotations to the document, so you can't add any further commands on it. So what you're doing seems to be a viable workaround for now. I will write up an internal request for us to investigate how viable it is to add an alternative approach here.

So for the time being we can test this solution in our next beta.

michalrentka:
> I got this feedback from PSPDFKit regarding this solution:
> 
> > Currently it is not possible to acquire the undo recorder that records the action to add the annotations to the document, so you can't add any further commands on it. So what you're doing seems to be a viable workaround for now. I will write up an internal request for us to investigate how viable it is to add an alternative approach here.
> 
> So for the time being we can test this solution in our next beta.

Yeah that's unfortunate. This should work fine 99% of the time, we'll see if we get any reports.

PR #938, opened 2024-05-28
Fix restoration of deleted annotation

mvasilak:
When an annotation is deleted, if the action is undone, then the `annotation.rects` will be appended again to the `item.rects`. If you exit the document, and open it again, these rects will be draw with multiply blend mode, resulting in a darker annotation color.

It seems that returning after toggling the `item.deleted` property should be enough for this type of restoration.

### DIFF
--- a/Zotero/Controllers/Database/Requests/CreatePDFAnnotationsDbRequest.swift
+++ b/Zotero/Controllers/Database/Requests/CreatePDFAnnotationsDbRequest.swift
@@ -30,6 +30,7 @@ struct CreatePDFAnnotationsDbRequest: DbRequest {
     }
 
     private func create(annotation: PDFDocumentAnnotation, parent: RItem, in database: Realm) {
+        var fromRestore = false
         let item: RItem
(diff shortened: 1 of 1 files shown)


michalrentka:
What if `rects`/`paths` changed too though? Steps to test:
- turn airplane mode on iPad on
- delete ink annotation on iPad
- move ink annotation on desktop
- turn airplane mode on iPad off and sync

I believe this will result in restored incorrect annotation. So you shouldn't just `return` there, but you need to update `rects`/`paths`. That would mean that you can change `add(rects`/`add(paths` functions to remove original `rects`/`paths` if the item is being restored.

mvasilak:
When sync updates are processed, notifications are disabled, so this `DbRequest` won't be performed. In any case, check https://github.com/zotero/zotero-ios/pull/938/commits/0781efd758cd8b309aac7e677d7e944eea38ce4f for a safer approach.

michalrentka (approved):

PR #939, opened 2024-05-28
Fixed crash of remotely changed and updated document

### LINKED ISSUE #937: PDF Sync crash 1439304954
https://forums.zotero.org/discussion/114654/ios-crash-report-1439304954#latest

michalrentka:
Fixes #937

### DIFF
--- a/Zotero/Scenes/Detail/PDF/PDFCoordinator.swift
+++ b/Zotero/Scenes/Detail/PDF/PDFCoordinator.swift
@@ -36,7 +36,7 @@ protocol PdfReaderCoordinatorDelegate: AnyObject {
     func showFontSizePicker(sender: UIView, picked: @escaping (UInt) -> Void)
     func showDeleteAlertForAnnotation(sender: UIView, delete: @escaping () -> Void)
     func showTagPicker(libraryId: LibraryIdentifier, selected: Set<String>, userInterfaceStyle: UIUserInterfaceStyle?, picked: @escaping ([Tag]) -> Void)
-    func showDocumentChangedAlert(key: String, parentKey: String?, libraryId: LibraryIdentifier)
+    func showDocumentChangedAlert(key: String, parentKey: String?, libraryId: LibraryIdentifier, completed: @escaping () -> Void)
 }
 
 protocol PdfAnnotationsCoordinatorDelegate: AnyObject {
@@ -435,8 +435,10 @@ extension PDFCoordinator: PdfReaderCoordinatorDelegate {
         self.navigationController?.present(presentedController, animated: true)
     }
 
-    func showDocumentChangedAlert(key: String, parentKey: String?, libraryId: LibraryIdentifier) {
-        (parentCoordinator as? DetailCoordinator)?.showDocumentChangedAlert(key: key, parentKey: parentKey, libraryId: libraryId)
+    func showDocumentChangedAlert(key: String, parentKey: String?, libraryId: LibraryIdentifier, completed: @escaping () -> Void) {
+        let controller = UIAlertController(title: L10n.warning, message: L10n.Errors.Pdf.documentChanged, preferredStyle: .alert)
+        controller.addAction(UIAlertAction(title: L10n.ok, style: .cancel, handler: { _ in completed() }))
+        navigationController?.present(controller, animated: true)
     }
 }
 
--- a/Zotero/Scenes/Detail/PDF/PDFCoordinator.swift
+++ b/Zotero/Scenes/Detail/PDF/PDFCoordinator.swift
@@ -36,6 +36,7 @@ protocol PdfReaderCoordinatorDelegate: AnyObject {
     func showFontSizePicker(sender: UIView, picked: @escaping (UInt) -> Void)
     func showDeleteAlertForAnnotation(sender: UIView, delete: @escaping () -> Void)
     func showTagPicker(libraryId: LibraryIdentifier, selected: Set<String>, userInterfaceStyle: UIUserInterfaceStyle?, picked: @escaping ([Tag]) -> Void)
+    func showDocumentChangedAlert(key: String, parentKey: String?, libraryId: LibraryIdentifier)
 }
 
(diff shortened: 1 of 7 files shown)


michalrentka:
@dstillman can you check the error string and probably update it to something better?

Zotero/Scenes/Detail/PDF/Views/PDFReaderViewController.swift
```diff
@@ -421,6 +421,14 @@ class PDFReaderViewController: UIViewController {
     // MARK: - Actions
 
     private func update(state: PDFReaderState) {
+        if state.changes.contains(.md5) {
+            DispatchQueue.main.async {
+                self.coordinatorDelegate?.showDocumentChangedAlert(key: state.key, parentKey: state.parentKey, libraryId: state.library.identifier)
+            }
+            close()
```
mvasilak, line 428: It might be better if either `close()` has a completion block that it passes to `dismiss` and in that  `showDocumentChangedAlert()` is called,
or even if `showDocumentChangedAlert()` has an action block parameter, that then calls `close()`.
In the latter case, there is no need to have the `DetailCoordinator.navigationController` present, and it seems less forced than just dismissing the current screen.

mvasilak (commented):
Instead of just closing the reader, couldn't we offer an action to reload the document? That is, create a new `Document` (the `url` is the same) and set it to the document view controller child pdf view controller, which will trigger a reload. Of course other elements would need to be reloaded, such as page thumbnails, annotations, etc.

If that is feasible, I guess we can even make it a single action alert?

michalrentka:
> Instead of just closing the reader, couldn't we offer an action to reload the document? That is, create a new `Document` (the `url` is the same) and set it to the document view controller child pdf view controller, which will trigger a reload. Of course other elements would need to be reloaded, such as page thumbnails, annotations, etc.
> 
> If that is feasible, I guess we can even make it a single action alert?

It is feasible, but it's quite a lot of work for a thing that doesn't really happen normally. I mean it's possible, but people don't usually just manually edit the PDF file stored in Zotero and add/remove pages from it. So if that happens I think just closing the document with an explanation is fine. If there are complaints we can improve this, but I wouldn't overdo it.

Zotero/Scenes/Detail/PDF/Views/PDFReaderViewController.swift
michalrentka: Changed in bab66baa9ff2f21e76412a20c678189eb7f53680

Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
```diff
@@ -1629,13 +1629,21 @@ final class PDFReaderActionHandler: ViewModelActionHandler, BackgroundDbProcessi
...
             }
 
-            let isDark = viewModel.state.interfaceStyle == .dark
             let key = viewModel.state.key
+            let (item, liveAnnotations, storedPage) = try loadItemAnnotationsAndPage(for: key, libraryId: viewModel.state.library.identifier)
+
+            if let url = viewModel.state.document.fileURL, let md5 = md5(from: url), item.backendMd5 != md5 {
+                update(viewModel: viewModel) { state in
+                    state.changes = .md5
+                }
```
mvasilak, line 1638: This part is similar to `case .change` in nested function `observe` below. It can be refactored to a private method, or even better another nested function.

Zotero/Scenes/Detail/PDF/PDFCoordinator.swift
```diff
@@ -36,6 +36,7 @@ protocol PdfReaderCoordinatorDelegate: AnyObject {
     func showFontSizePicker(sender: UIView, picked: @escaping (UInt) -> Void)
     func showDeleteAlertForAnnotation(sender: UIView, delete: @escaping () -> Void)
     func showTagPicker(libraryId: LibraryIdentifier, selected: Set<String>, userInterfaceStyle: UIUserInterfaceStyle?, picked: @escaping ([Tag]) -> Void)
+    func showDocumentChangedAlert(key: String, parentKey: String?, libraryId: LibraryIdentifier, completed: @escaping () -> Void)
```
mvasilak, line 39: The first parameters seem not to be used at all.

Zotero/Scenes/Detail/PDF/Views/PDFReaderViewController.swift
```diff
@@ -421,6 +421,13 @@ class PDFReaderViewController: UIViewController {
     // MARK: - Actions
 
     private func update(state: PDFReaderState) {
+        if state.changes.contains(.md5) {
+            self.coordinatorDelegate?.showDocumentChangedAlert(key: state.key, parentKey: state.parentKey, libraryId: state.library.identifier) { [weak self] in
```
mvasilak, line 425: Remove `self`.

mvasilak (changes requested):

Zotero/Scenes/Detail/PDF/PDFCoordinator.swift
michalrentka: Removed in 9b5ebbbe2e66e8aa4763a6155777bfb0053b416e

Zotero/Scenes/Detail/PDF/Views/PDFReaderViewController.swift
michalrentka: Removed in 9b5ebbbe2e66e8aa4763a6155777bfb0053b416e

mvasilak (approved):

dstillman:
> "This document has been changed. You have to re-download it if you want to continue editing it."

@michalrentka: Does the alert appear before or after the reader closes?

michalrentka:
> > "This document has been changed. You have to re-download it if you want to continue editing it."
> 
> @michalrentka: Does the alert appear before or after the reader closes?

@dstillman currently before, but I can change it to after.

dstillman:
No, before is fine — just want to get the phrasing right.

PR #943, opened 2024-05-31
Debounce free text annotation rotation by key

### LINKED ISSUE #921: Merging Ink annotations sometimes keeps only the initial paths and removes the rest
https://forums.zotero.org/discussion/comment/462335#Comment_462335

Bugs:
- merging ink annotations keeps first annotation's paths and removes the rest
- annotations sometimes don't load up properly (empty locked annotations in sidebar)

mvasilak:
Debouncing all annotation changes, creates a race issue when merging ink annotations. As merging also removes annotations, the removal will cause a database update, that will disable notifications temporarily. The debounce delay may cause the annotation change notification to not be observed, if it falls in the disabled period. Debouncing after the notification is received, resolves the issue.
As a further improvement, the debouncing happens only for the offending annotation type and change.

Closes #921

### DIFF
--- a/Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
+++ b/Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
@@ -94,6 +94,7 @@ final class PDFReaderActionHandler: ViewModelActionHandler, BackgroundDbProcessi
     private var pdfDisposeBag: DisposeBag
     private var pageDebounceDisposeBag: DisposeBag?
     private var freeTextAnnotationRotationDebounceDisposeBagByKey: [String: DisposeBag]
+    private var debouncedFreeTextAnnotationAndChangesByKey: [String: ([String], PSPDFKit.FreeTextAnnotation)]
     weak var delegate: (PDFReaderContainerDelegate & AnnotationBoundingBoxConverter)?
 
     init(
@@ -117,6 +118,7 @@ final class PDFReaderActionHandler: ViewModelActionHandler, BackgroundDbProcessi
         self.backgroundQueue = DispatchQueue(label: "org.zotero.Zotero.PDFReaderActionHandler.queue", qos: .userInteractive)
         self.pdfDisposeBag = DisposeBag()
         freeTextAnnotationRotationDebounceDisposeBagByKey = [:]
+        debouncedFreeTextAnnotationAndChangesByKey = [:]
         self.disposeBag = DisposeBag()
     }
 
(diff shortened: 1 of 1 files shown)


Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
```diff
@@ -1780,7 +1782,26 @@ final class PDFReaderActionHandler: ViewModelActionHandler, BackgroundDbProcessi
...
+                        Single<Int>.timer(.milliseconds(100), scheduler: MainScheduler.instance)
+                            .subscribe(onSuccess: { [weak self, weak viewModel] _ in
+                                guard let self, let viewModel else { return }
+                                change(annotation: annotation, with: changes, in: viewModel)
+                                freeTextAnnotationRotationDebounceDisposeBagByKey[key] = nil
+                            })
+                            .disposed(by: disposeBag)
+                    } else {
+                        change(annotation: annotation, with: changes, in: viewModel)
+                        freeTextAnnotationRotationDebounceDisposeBagByKey[key] = nil
```
michalrentka, line 1800: Wouldn't this cancel previous rotation notification? You can leave `freeTextAnnotationRotationDebounceDisposeBagByKey` alone here, it'll be reset when it's debounced.

michalrentka (approved):

Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
michalrentka: It's change for the same annotation, but the change doesn't contain `rotation` change this time. The `func change(annotation: PSPDFKit.Annotation, with changes: [String], in viewModel: ViewModel<PDFReaderActionHandler>)` checks `changes` array to see which changes were applied and applies only those changes.

So if for example you got a `rotation` change and then suddenly a `color` change, rotation wouldn't be applied, only new color. I know that the 100 ms debounce won't really let this happen, but we may increase the debounce later while not being aware of this.

michalrentka: New commit should work fine, feel free to merge

mvasilak: I misunderstood that. Check https://github.com/zotero/zotero-ios/pull/943/commits/2328bf63dc8bf3e8019dd244e0f9877528364e51 for a solution that maintains proper order of execution.

PR #944, opened 2024-06-03
Pull to refresh for collections and libraries when split controller is collapsed

michalrentka:


Zotero/Extensions/Localizable.swift
```diff
@@ -538,8 +538,8 @@ internal enum L10n {
       internal static let cantDeleteAnnotations = L10n.tr("Localizable", "errors.pdf.cant_delete_annotations", fallback: "Can't delete annotations.")
       /// Can't update annotation.
       internal static let cantUpdateAnnotation = L10n.tr("Localizable", "errors.pdf.cant_update_annotation", fallback: "Can't update annotation.")
-      /// This document has been changed. You have to re-download it if you want to continue editing it.
-      internal static let documentChanged = L10n.tr("Localizable", "errors.pdf.document_changed", fallback: "This document has been changed. You have to re-download it if you want to continue editing it.")
+      /// This document has been changed on another device. Please reopen it to continue editing.
+      internal static let documentChanged = L10n.tr("Localizable", "errors.pdf.document_changed", fallback: "This document has been changed on another device. Please reopen it to continue editing.")
```
mvasilak, line 542: I've already commited this change to master, rebase before merging.

Zotero/Scenes/Master/Collections/Views/CollectionsViewController.swift
```diff
@@ -70,6 +78,19 @@ final class CollectionsViewController: UICollectionViewController {
                       .disposed(by: self.disposeBag)
     }
 
+    override func viewIsAppearing(_ animated: Bool) {
+        super.viewIsAppearing(animated)
+
+        if !isSplit {
```
mvasilak, line 84: This will add it even if in collapsed mode in iPad, but it will not remove it if it changes back to split mode. Probably it's ok though, no need to overthink it.

mvasilak (approved):

Zotero/Scenes/Master/Collections/Views/CollectionsViewController.swift
michalrentka: Yeah I thought about it but it's pretty hidden and I at least added it to `viewIsAppearing` so that it's removed at some point. Though even if you pull to refresh in split mode it's not the end of the world. I felt like observing the split/collapsed state change and updating based on that was unnecessary.

Zotero/Extensions/Localizable.swift
michalrentka: Done

PR #948, opened 2024-06-05
Add reset open item application shortcut

mvasilak:
Adds "Reset open item" static application shortcut, that can be used in the edge case where the app restores an item that immediately crashes. It can be selected by long tapping on the app icon.


dstillman:
This isn't ideal. Can we just detect a crash and reset state automatically?

michalrentka:
> This isn't ideal. Can we just detect a crash and reset state automatically? We shouldn't be exposing debug options on the app icon.

We discussed this. We can, but in some cases, users report that they just crash immediately instead of the Report ID alert popping up, which means that our crash reporter didn't catch the crash and it just crashes immediately again. So we decided this would be next best option, it's quite hidden, people probably won't even notice. I would just rename it to something like "Show all items" instead of "Reset open item"

dstillman:
Maybe just "My Library", which takes you to All Items in My Library?

michalrentka:
> Maybe just "My Library", which takes you to All Items in My Library? That will seem like a totally normal option.

Agreed, that's sounds fine to me.

michalrentka (approved):
Just change the title, otherwise looks good.

dstillman:
(If we wanted to actually make this a useful feature, we could even populate it with My Library + the names of the last 3 visited group libraries, if we're able to do that.)

dstillman:
Can we localize this string?

mvasilak:
Updated the shortcut title (localized), icon, and action to go to My Library/All Items. Adding actions for the last 3 visited group libraries is possible with dynamic application shortcuts, I'll open a separate issue for that.

michalrentka (approved):

PR #954, opened 2024-06-10
Fix annotation comment icon drawing

mvasilak:
Fixes issue mentioned in forum [post](https://forums.zotero.org/discussion/115061/ios-double-area-annotation-comment-icons)


michalrentka:
Does this also fix icons disappearing when annotation popup is opened? That looked weird too.

mvasilak:
No, it doesn't fix the selection disappearance, I can investigate this as well.

mvasilak:
@michalrentka the note icon disappears because for shape annotations (square, ink), when they are selected PSPDFKit redraws them in a separate `ShapeAnnotationView`, so that it can be resized or moved. However, this view doesn't take into account the overridden `SquareAnnotation` drawing methods. I'll ask PSPDFKit if there is a way to override this.

PR #956, opened 2024-06-12
Check for duplicates and fix them

michalrentka:


Zotero/Controllers/Controllers.swift
```diff
@@ -372,6 +372,8 @@ final class UserControllers {
                 try coordinator.perform(request: ResetSettingsVersionDbRequest())
                 controllers?.tagOrderNeedsSync = false
             }
+
+//            try coordinator.perform(request: CreateTestNote())
```
mvasilak, line 376: Forgotten code?

mvasilak (approved):

Zotero/Controllers/Controllers.swift
michalrentka: Oops, deleted

PR #966, opened 2024-06-24
Fix AttachmentDownloader URLSession memory leak

mvasilak:
If user signs out and signs in again, without terminating the app, their downloads are still handled by the leaked delegate, and cannot be finished.


michalrentka:
Can you also check other instances of `URLSession` usage? I see `URLSession` is initialized in `ExtensionViewModel` and `WebViewHandler`, so we should probably add invalidation there too?

michalrentka (approved):

mvasilak:
In these cases, there is no `URLSession` `delegate` set, therefore no potential leak.

PR #969, opened 2024-06-27
Improve transformed annotation removal

mvasilak:
There seems to be an edge case where removing a transformed annotation, throws undocumented PSPDFKit error `The removed annotation does not belong to the current document`, causing a crash. (Reported in forum [post](https://forums.zotero.org/discussion/115287/ios-crash-report-1677056178) and other submitted crashes.) Check in PR helps avoid that.

### DIFF
--- a/Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
+++ b/Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
@@ -1408,8 +1408,13 @@ final class PDFReaderActionHandler: ViewModelActionHandler, BackgroundDbProcessi
             }
             undoManager.enableUndoRegistration()
             undoController.recordCommand(named: nil, adding: finalAnnotations) {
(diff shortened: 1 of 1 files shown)


Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
```diff
@@ -1408,8 +1408,13 @@ final class PDFReaderActionHandler: ViewModelActionHandler, BackgroundDbProcessi
             }
             undoManager.enableUndoRegistration()
             undoController.recordCommand(named: nil, adding: finalAnnotations) {
-                // Remove may be superfluous, if those annotations are already removed by the undo, but it doesn't cause any issue.
-                document.remove(annotations: toRemove, options: [.suppressNotifications: true])
+                // Remove may be superfluous, if those annotations are already removed by the undo.
```
michalrentka, line 1411: That old comment didn't age well :)

michalrentka (approved):

PR #973, opened 2024-07-02
Fix Apple Pencil double-tap bug and add more tool switching functionality

slushydroplet:
1.) Fixes a bug where double tapping on the Apple Pencil could activate annotation mode without any UI change, mentioned here:

https://forums.zotero.org/discussion/105598/bug-ipad-douple-tapping-apple-pencil-activates-annotation-tools-outside-of-annotation-mode

https://forums.zotero.org/discussion/115677/ios-apple-pencil-double-tap-and-annotation-panel

2.) Adds more annotation tools and functionality to the Apple Pencil "Switch Between Current Tool and Last Used" double-tap mode. Currently, this mode ignores the eraser and also has no way to switch tools *off*. I've found that other PDF readers interpret this mode to include "no tool" as a switchable state (PDF Expert) so that users can double tap to activate and deactivate tools. Other PDF readers have a designated "hand" tool (Notability) that serves this purpose, but I don't think PSPDFKit supports this. The proposed functionality adds both the eraser and the unset tool states as switchable.

This second commit also makes some adjustments to distinguish between unset tool states happening in the course of normal work and those happening immediately after reopening a PDF or reactivating annotation mode. This is to ensure that a user's tool history is not overwritten in these situations since it is not a meaningful/deliberate use of the unset state.

This is unsolicited, so I understand if it isn't a priority to consider at this time. These changes just seemed too minor and too niche to ask someone else to work on. Thank you for all the work you do on this program 🙏

### DIFF
--- a/Zotero/Scenes/Detail/PDF/Views/PDFDocumentViewController.swift
+++ b/Zotero/Scenes/Detail/PDF/Views/PDFDocumentViewController.swift
@@ -830,6 +830,8 @@ extension PDFDocumentViewController: AnnotationStateManagerDelegate {
 
 extension PDFDocumentViewController: UIPencilInteractionDelegate {
     func pencilInteractionDidTap(_ interaction: UIPencilInteraction) {
+        guard self.parentDelegate?.isToolbarVisible == true else { return }
+    
         switch UIPencilInteraction.preferredTapAction {
         case .switchEraser:
             if let tool = self.pdfController?.annotationStateManager.state, tool != .eraser {
--- a/Zotero/Scenes/Detail/PDF/Views/PDFDocumentViewController.swift
+++ b/Zotero/Scenes/Detail/PDF/Views/PDFDocumentViewController.swift
@@ -37,8 +37,7 @@ final class PDFDocumentViewController: UIViewController {
     private let disposeBag: DisposeBag
     private let initialUIHidden: Bool
 
-    private static var toolHistory: [PSPDFKit.Annotation.Tool] = []
-    private static var disabledAnnotationTools: Bool = false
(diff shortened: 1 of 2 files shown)


mvasilak (changes requested):
Hey @slushydroplet, bug fix looks good, but can you improve the switching addition, by not adding another static variable (`disabledAnnotationTools`)? Either make it a local variable with a specific value upon initialization, or even better, use the existing static `toolHistory` e.g. as `toolHistory: [PSPDFKit.Annotation.Tool?]` to store `nil` to keep track of no tool selected (some minor modifications will be needed).

slushydroplet:
Thanks for the feedback @mvasilak – I adjusted `toolHistory` to include nil and made the modifications needed to make that work. It is definitely a more elegant approach!

The part about `disabledAnnotationTools` is trickier. The awkwardness here derives from (1) tools being added to `toolHistory` when they are switched away *from*, not *to*, and (2) how the unset state naturally occurs when reactivating annotation mode or switching PDFs. Together, they mean that users who customarily switch between two tools would necessarily overwrite one with nil every time they switch PDFs or move in/out of annotation mode.

To address this, I changed `toolHistory` to add tools at the point of selection, not departure. I removed `disabledAnnotationTools` as a static variable and replicated its function by adding an optional `noToolHistory` argument to the `toggle` function. With this, I can make it so that the function `disableAnnotationTools` doesn't improperly add a nil to the history when switching windows or turning off annotation mode. I think that fixes things, but let me know if I've erred or if things are unclear.

mvasilak (changes requested):

slushydroplet:
Thanks for the suggestions @mvasilak. I committed them, and they work as expected in tests.

mvasilak (approved):

PR #975, opened 2024-07-03
Performance improvements in PDF Viewer

michalrentka:
I cleaned up loading of annotations for pdf viewer so that we don't go through all annotations multiple times, which leads to some small-ish performance improvements. If we want to improve loading further we'd probably have to load annotations in batches instead of all at once.


mvasilak (approved):
Why was the rounding needed in the first place, and why is it not necessary now? Are just excess roundings removed?

michalrentka:
> Why was the rounding needed in the first place, and why is it not necessary now? Are just excess roundings removed?

API expects numbers to be rounded to 3 decimals and yes I removed excess roundings.

PR #979, opened 2024-07-10
Cache file MD5 and modification date by URL for app session

mvasilak:
There is a number of reported Hangs for version 1.0.39, when MD5 hashes are computed. These are re-computed every time a document appears as downloaded in a collection, or it is opened. Adding a simple session memory cache, can speed subsequent uses if the file has not been modified.

### DIFF
--- a/Zotero/Extensions/MD5+Url.swift
+++ b/Zotero/Extensions/MD5+Url.swift
@@ -53,7 +53,7 @@ func cachedMD5(from url: URL, using fileManager: FileManager) -> String? {
     } else {
         newModificationDate = .distantPast
     }
-    if let (cachedMd5, cachedModificationDate) = cachedMD5AndModificationDateByURL[url], newModificationDate <= cachedModificationDate {
+    if let (cachedMd5, cachedModificationDate) = cachedMD5AndModificationDateByURL[url], newModificationDate == cachedModificationDate {
         return cachedMd5
     }
     let md5 = md5(from: url)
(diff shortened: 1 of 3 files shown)


Zotero/Extensions/MD5+Url.swift
```diff
@@ -44,3 +44,23 @@ func md5(from url: URL) -> String? {
...
+
+var cachedMD5AndModificationDateByURL: [URL: (String, Date)] = [:]
+func cachedMD5(from url: URL, using fileManager: FileManager) -> String? {
+    let newModificationDate: Date
+    if let attributes = try? fileManager.attributesOfItem(atPath: url.path), let modificationDate = attributes[.modificationDate] as? Date {
+        newModificationDate = modificationDate
+    } else {
+        newModificationDate = .distantPast
+    }
+    if let (cachedMd5, cachedModificationDate) = cachedMD5AndModificationDateByURL[url], newModificationDate <= cachedModificationDate {
```
michalrentka, line 56: It shouldn't matter, but why `newModificationDate <= cachedModificationDate`? Why not equality only?

michalrentka (approved):

Zotero/Extensions/MD5+Url.swift
mvasilak: Hmmm, you are right, to be on the safe side, any change should be enough for recomputing the hash. Check https://github.com/zotero/zotero-ios/pull/979/commits/82c0443562bee4ba13578f5e89f18b5cee244472

Maybe we should add a size comparison as well?

michalrentka: I guess if the size changes the modification date should change as well, so it might be redundant. But it wouldn't hurt anyone.

PR #980, opened 2024-07-13
Fix group owner permissions

mvasilak:
Fixes missing permissions for owners, when only admins have metadata or file editing.

### DIFF
--- a/Zotero/Controllers/Database/Requests/StoreGroupDbRequest.swift
+++ b/Zotero/Controllers/Database/Requests/StoreGroupDbRequest.swift
@@ -56,7 +56,7 @@ struct StoreGroupDbRequest: DbRequest {
             canEditFiles = true
 
         default:
-            canEditFiles = false
+            canEditFiles = (response.data.owner == userId)
         }
 
         group.name = response.data.name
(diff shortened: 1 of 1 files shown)


Zotero/Controllers/Database/Requests/StoreGroupDbRequest.swift
```diff
@@ -50,7 +50,7 @@ struct StoreGroupDbRequest: DbRequest {
             canEditFiles = false
 
         case "admins":
-            canEditFiles = (response.data.admins ?? []).contains(userId)
+            canEditFiles = (response.data.admins ?? []).contains(userId) || (response.data.owner == userId)
```
michalrentka, line 53: @mvasilak Dan mentioned that `owners, like admins, always have libraryEditing, and they have fileEditing unless the group itself doesn't allow fileEditing`. So `owner == userId` shouldn't automatically enable file editing? I think we had it like this before and I changed it to this some time ago.

mvasilak: You are right, check https://github.com/zotero/zotero-ios/pull/980/commits/87123f6686dc87dd1029b54cbfa756239e116cfe

michalrentka (approved):

PR #983, opened 2024-07-17
Add collection remove downloads action

### LINKED ISSUE #981: Remove selected attachments
https://forums.zotero.org/discussion/95117/zotero-6-ios-batch-remove-downloaded-pdf#latest

mvasilak:
* Adds `Remove Downloads` action in collection context menu, when the collection has items. Similar to `Download Attachments` action, it doesn't explicitely check if there are any attachments already downloaded, it just checks for collection items count. This action is placed 2nd in the menu, after `Download Attachments`. For `Trash` it is placed 1st, as non-destructive.
* Adds `Remove downloads for selected items` action in items toolbar. To differentiate from `Move selected items to trash` the icon  `arrow.down.circle.dotted` is used.

Perhaps we should consider using the same icon, either a system or a custom one, for all actions that remove downloads.
Additionally, since we allow to remove downloads for selected items, maybe we should also allow to download attachments for selected items? If the toolbar gets too populated though, maybe we should consider a more button that presents a menu with the additional actions.

Closes #981

### DIFF
--- a/Zotero/Assets/en.lproj/Localizable.strings
+++ b/Zotero/Assets/en.lproj/Localizable.strings
@@ -533,6 +533,7 @@
 "accessibility.items.restore" = "Restore selected items";
 "accessibility.items.duplicate" = "Duplicate selected item";
 "accessibility.items.share" = "Share selected items";
+"accessibility.items.remove_downloads" = "Remove downloads for selected items";
 "accessibility.item_detail.download_and_open" = "Double tap to download and open";
 "accessibility.item_detail.open" = "Double tap to open";
 "accessibility.pdf.sidebar_open" = "Open sidebar";
--- a/Zotero/Assets/en.lproj/Localizable.strings
+++ b/Zotero/Assets/en.lproj/Localizable.strings
@@ -533,6 +533,7 @@
 "accessibility.items.restore" = "Restore selected items";
 "accessibility.items.duplicate" = "Duplicate selected item";
 "accessibility.items.share" = "Share selected items";
+"accessibility.items.download_attachments" = "Download attachments for selected items";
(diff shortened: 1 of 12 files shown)


michalrentka (approved):

michalrentka:
> * Adds `Remove Downloads` action in collection context menu, when the collection has items. Similar to `Download Attachments` action, it doesn't explicitely check if there are any attachments already downloaded, it just checks for collection items count. This action is placed 2nd in the menu, after `Download Attachments`. For `Trash` it is placed 1st, as non-destructive.
> * Adds `Remove downloads for selected items` action in items toolbar. To differentiate from `Move selected items to trash` the icon  `arrow.down.circle.dotted` is used.
> 
> Perhaps we should consider using the same icon, either a system or a custom one, for all actions that remove downloads. Additionally, since allow to remove downloads for selected items, maybe we should also allow to download attachments for selected items? If the toolbar gets too populated though, maybe we should consider a more button that presents a menu with the additional actions.
> 
> Closes #981

I agree having 2 icons for "delete item" and "delete attachment" would be great. 

I'd say download selected actions is unnecessary. It's simpler to just start tapping items you want downloaded. Most people will probably download the whole collection anyway. But removing attachments can benefit from this because currently you have to long press each attachment and then tap to delete it.

dstillman:
You're saying this adds an option to remove downloads for selected items without their being an option to download files for selected items? I don't think that makes sense. Managing storage is a much more advanced task than just wanting files available offline, and you can always purge all files or (after this) purge all files for a given collection. So I think we want to either add both as multi-select options or only the download option.

michalrentka (approved):

PR #987, opened 2024-07-18
Add highlight annotation transformation for overlapping lines

### LINKED ISSUE #984: Avoid overlapping highlighting with tighter line spacing
https://forums.zotero.org/discussion/comment/468295/#Comment_468295

I assume this is an issue for PSPDFKit, but the desktop one seems arguably more correct.

mvasilak:
Transforms highlight annotation that are added by a user in the device, if there are overlapping line rects, by trimming equidistantly.

Closes #984

### DIFF
--- a/Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
+++ b/Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
@@ -1547,7 +1547,8 @@ final class PDFReaderActionHandler: ViewModelActionHandler, BackgroundDbProcessi
                             trimmedRects.append(currentRect)
                             continue
                         }
-                        let trim = intersection.height / 2
+                        // Each rect is trimmed by half the intersection height, plus 0.25 to have a small gap between the lines.
+                        let trim = (intersection.height / 2) + 0.25
                         let previousTrimmedRect = CGRect(x: previousRect.minX, y: previousRect.minY + trim, width: previousRect.width, height: previousRect.height - trim)
                         let currentTrimmedRect = CGRect(x: currentRect.minX, y: currentRect.minY, width: currentRect.width, height: currentRect.height - trim)
                         trimmedRects.removeLast()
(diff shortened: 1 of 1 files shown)


michalrentka (approved):

Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
```diff
@@ -1547,7 +1547,8 @@ final class PDFReaderActionHandler: ViewModelActionHandler, BackgroundDbProcessi
                             trimmedRects.append(currentRect)
                             continue
                         }
-                        let trim = intersection.height / 2
+                        // Each rect is trimmed by half the intersection height, plus 0.25 to have a small gap between the lines.
+                        let trim = (intersection.height / 2) + 0.25
```
michalrentka, line 1551: I'm not against this hard coded value, but does it work for all/most documents? What about different font sizes or such?

michalrentka (approved):

Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
mvasilak: It's in the annotation's normalized PDF coordinates space, so it shouldn't be affected by font sizes. The goal is to have just a small gap, so that rects do not be viewed as a seamless block, but probably we don't want it to be in relation to the rects' height, as a larger part of the text glyphs ascenders/descenders may get out of the rect. In any case, I'll test with more documents first.

mvasilak: Fixed gap size works well for different font spaces. 

While testing, I found some edge cases, when the whole page is selected, due to the way PSPDFKit sorts the selection rects. Specificallly, it might place a rect on top, especially if it is centered on page, between two consecutive line rects that are on the page footer, resulting on skipping an overlap. Since this is a contrived selection to annotate, I'd argue that we shouldn't concern ourselves for now, unless we have reports from users.

michalrentka: Sounds good, feel free to merge.

PR #999, opened 2024-08-22
Fix reader action handler update from database

mvasilak:
* Uses correct indexes if there are invalid annotations
* Improves free text annotation validation

### DIFF
--- a/Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
+++ b/Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
@@ -2136,17 +2136,7 @@ final class PDFReaderActionHandler: ViewModelActionHandler, BackgroundDbProcessi
 
         DDLogInfo("PDFReaderActionHandler: database annotation changed")
 
-        // Keys for deletions and modifications are indexed by the previously observed items
-        var previousObservedKeys: [PDFReaderState.AnnotationKey] = []
-        for item in viewModel.state.databaseAnnotations {
-            previousObservedKeys.append(.init(key: item.key, type: .database))
-        }
-        // Keys for insertions are indexed by the currently observed items
-        var currentObservedKeys: [PDFReaderState.AnnotationKey] = []
-        for item in objects {
-            currentObservedKeys.append(.init(key: item.key, type: .database))
-        }
-
+        let databaseAnnotations = viewModel.state.databaseAnnotations!
         var texts = viewModel.state.texts
         var comments = viewModel.state.comments
         var selectKey: PDFReaderState.AnnotationKey?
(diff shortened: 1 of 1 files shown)


Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
```diff
@@ -2135,8 +2136,17 @@ final class PDFReaderActionHandler: ViewModelActionHandler, BackgroundDbProcessi
 
         DDLogInfo("PDFReaderActionHandler: database annotation changed")
 
-        // Get sorted database keys
-        var keys = (viewModel.state.snapshotKeys ?? viewModel.state.sortedKeys).filter({ $0.type == .database })
+        // Keys for deletions and modifications are indexed by the previously observed items
+        var previousObservedKeys: [PDFReaderState.AnnotationKey] = []
```
michalrentka, line 2140: Can we just directly use `viewModel.state.databaseAnnotations`? We do that on line 2225 anyway. No real need to extract keys separately.

Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
```diff
@@ -2135,8 +2136,17 @@ final class PDFReaderActionHandler: ViewModelActionHandler, BackgroundDbProcessi
...
+        for item in viewModel.state.databaseAnnotations {
+            previousObservedKeys.append(.init(key: item.key, type: .database))
+        }
+        // Keys for insertions are indexed by the currently observed items
+        var currentObservedKeys: [PDFReaderState.AnnotationKey] = []
```
michalrentka, line 2145: Does this even need to be created? I only see `currentObservedKeys.count` being used, so you could just do `objects.count` or store `let currentKeysCount = objects.count` and use that for clarity.

michalrentka (changes requested):

Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
mvasilak: Good point (I planned to improve that, after changing font size to use a float, but better if I do it in a separate PR)
Check https://github.com/zotero/zotero-ios/pull/999/commits/6e0b28fbb1a3b464b94c6cc4bbe698c5c366c8cb

Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/999/commits/6e0b28fbb1a3b464b94c6cc4bbe698c5c366c8cb

michalrentka (approved):

PR #1000, opened 2024-08-23
Change free text annotation font size to float

mvasilak:


Zotero/Models/AnnotationsConfig.swift
```diff
@@ -29,6 +29,13 @@ struct AnnotationsConfig {
     static let keyKey = "Zotero:Key"
     // Line width of image annotation in PDF document.
     static let imageAnnotationLineWidth: CGFloat = 2
+    // Free text annotation font size minimum, maximum, increment and rounding
+    static let freeTextAnnotationFontSizeMinimum: CGFloat = 1
+    static let freeTextAnnotationFontSizeMaximum: CGFloat = 200
+    static let freeTextAnnotationFontSizeIncrement: CGFloat = 0.5
+    static func roundFreeTextAnnotationFontSize(_ fontSize: CGFloat) -> CGFloat {
```
michalrentka, line 36: Just out of curiosity, why is this needed? Did we have some weird fractions stored or does server send other values?

Zotero/Controllers/Database/Requests/EditAnnotationFontSizeDbRequest.swift
```diff
@@ -13,7 +13,7 @@ import RealmSwift
 struct EditAnnotationFontSizeDbRequest: DbRequest {
     let key: String
     let libraryId: LibraryIdentifier
-    let size: UInt
+    let size: CGFloat
```
michalrentka, line 16: I can't add comment below to "original content", but is `field.value = "\(self.size)"` ok when we changed type to `CGFloat`? Wouldn't it be better to round the value here?

Zotero/Models/AnnotationsConfig.swift
mvasilak: If the user modifies the size using the resize handle, or even if they just start editing the text and do nothing, the size value changes. For conformity with the desktop client, that uses `0.5` steps, we round appropriately.

Zotero/Controllers/Database/Requests/EditAnnotationFontSizeDbRequest.swift
mvasilak: `"\(self.size)"` is formatted accordingly. The value passed to the request, is already rounded.

PR #1007, opened 2024-09-05
Note editor improvements

michalrentka:
Implemented note editor actions for annotations/citations. Small refactoring included.

### DIFF
--- a/Zotero/Controllers/ZoteroURIConverter.swift
+++ b/Zotero/Controllers/ZoteroURIConverter.swift
@@ -1,0 +1,28 @@
+//
+//  ZoteroURIConverter.swift
+//  Zotero
+//
+//  Created by Michal Rentka on 04.09.2024.
+//  Copyright © 2024 Corporation for Digital Scholarship. All rights reserved.
+//
+
+import Foundation
+
+final class ZoteroURIConverter {
+    func convert(uri: String) -> (String, LibraryIdentifier)? {
+        guard let components = URLComponents(string: uri) else { return nil }
+        let pathComponents = components.path.components(separatedBy: "/")
+        guard pathComponents.count == 5 else { return nil }
+        switch pathComponents[1] {
+        case "users":
+            return (pathComponents[4], .custom(.myLibrary))
+
+        case "groups":
+            guard let groupId = Int(pathComponents[2]) else { return nil }
+            return (pathComponents[4], .group(groupId))
+
+        default:
+            return nil
+        }
+    }
+}
--- a/Zotero/Scenes/AppCoordinator.swift
+++ b/Zotero/Scenes/AppCoordinator.swift
@@ -364,30 +364,7 @@ extension AppCoordinator: AppDelegateCoordinatorDelegate {
                 completion()
                 return
             }
-
-            let disposeBag = DisposeBag()
-
-            downloader.observable
-                .observe(on: MainScheduler.instance)
-                .subscribe(onNext: { [weak self] update in
-                    guard let self, update.libraryId == attachment.libraryId && update.key == attachment.key else { return }
-
-                    switch update.kind {
-                    case .ready:
-                        completion()
-                        downloadDisposeBag = nil
-
-                    case .cancelled, .failed:
-                        downloadDisposeBag = nil
-
-                    case .progress:
-                        break
-                    }
-                })
-                .disposed(by: disposeBag)
-
-            downloadDisposeBag = disposeBag
-            downloader.downloadIfNeeded(attachment: attachment, parentKey: parentKey)
+            downloader.downloadIfNeeded(attachment: attachment, parentKey: parentKey, completion: { _ in completion() })
         }
 
         func open(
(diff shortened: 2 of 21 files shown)


Zotero/Scenes/AppCoordinator.swift
```diff
@@ -364,30 +364,7 @@ extension AppCoordinator: AppDelegateCoordinatorDelegate {
...
-
-                    case .progress:
-                        break
-                    }
-                })
-                .disposed(by: disposeBag)
-
-            downloadDisposeBag = disposeBag
-            downloader.downloadIfNeeded(attachment: attachment, parentKey: parentKey)
+            downloader.downloadIfNeeded(attachment: attachment, parentKey: parentKey, completion: { _ in completion() })
```
mvasilak, line 367: In case of an error before, `completion` was not called, but probably it doesn't affect anything. We could consider passing the result there, if necessary.

Zotero/Controllers/ZoteroURIConverter.swift
```diff
@@ -0,0 +1,28 @@
...
+//  ZoteroURIConverter.swift
+//  Zotero
+//
+//  Created by Michal Rentka on 04.09.2024.
+//  Copyright © 2024 Corporation for Digital Scholarship. All rights reserved.
+//
+
+import Foundation
+
+final class ZoteroURIConverter {
```
mvasilak, line 11: There seems to be some overlap of concerns with `CustomURLController`. At some later point we can consider merging them.

Zotero/Scenes/General/NoteEditorCoordinator.swift
```diff
@@ -112,4 +121,41 @@ extension NoteEditorCoordinator: NoteEditorCoordinatorDelegate {
...
+        showPdf(key: citation.attachmentKey, libraryId: citation.libraryId, preview: (citation.locator, nil))
+    }
+
+    func showItemDetail(withCitation citation: CitationMetadata) {
+        guard let coordinator = (parentCoordinator as? DetailCoordinator) else { return }
+        coordinator.showItemDetail(for: .preview(key: citation.parentKey), libraryId: citation.libraryId, scrolledToKey: nil, animated: false)
+        navigationController?.dismiss(animated: true)
+    }
+
+    private func showPdf(key: String, libraryId: LibraryIdentifier, preview: (UInt, [CGRect]?)) {
```
mvasilak, line 139: I'm guessing `preview` wil be used later? However, it's a bit more clear to have two separate parameters, instead of a tuple, since the tuple is not optional.

mvasilak (approved):

Zotero/Scenes/AppCoordinator.swift
michalrentka: I know and it felt more like a bug before, so I added completion even in case of error.

Zotero/Controllers/ZoteroURIConverter.swift
michalrentka: Custom urls are similar (they have the same library/group part, but otherwise they are different, so I created different controller for it.

Examples:
URI: https://zotero.org/user/{user_id}/{key}
Custom url: zotero://select/library/items/{key}

mvasilak (approved):

Zotero/Scenes/Detail/PDF/Views/PDFDocumentViewController.swift
```diff
@@ -428,6 +433,43 @@ final class PDFDocumentViewController: UIViewController {
...
+    /// - parameter document: Active `Document` instance.
+    private func show(previewRects: [CGRect], pageIndex: PageIndex, document: PSPDFKit.Document) {
+        guard !previewRects.isEmpty, let pageView = self.pdfController?.pageViewForPage(at: pageIndex) else { return }
+
+        let convertedRects = previewRects.map({ pageView.convert($0, from: pageView.pdfCoordinateSpace) })
+        let view = AnnotationPreviewView(frames: convertedRects)
+        view.alpha = 0
+        pageView.contentView.addSubview(view)
+
+        UIView.animate(
```
mvasilak, line 448: Instead of using `asyncAfter` above and below, maybe the `animated` parameters `delay` and `options: .curveEaseInOut` can be used for simplicity?

Zotero/Scenes/Detail/PDF/Views/PDFDocumentViewController.swift
```diff
@@ -998,25 +1039,59 @@ extension PDFDocumentViewController: FreeTextInputDelegate {
...
+        backgroundColor = .clear
+        autoresizingMask = [.flexibleTopMargin, .flexibleLeftMargin, .flexibleBottomMargin, .flexibleRightMargin, .flexibleWidth, .flexibleHeight]
+        layer.borderColor = Asset.Colors.annotationHighlightSelection.color.cgColor
+        layer.borderWidth = 2.5
+        layer.cornerRadius = 2.5
+        layer.masksToBounds = true
+    }
+}
+
+final class AnnotationPreviewView: UIView {
```
mvasilak, line 1064: `AnnotationPreviewView` shares a lot of code with `SelectionView` and can probably be made a subclass of that one in the future.

mvasilak (approved):
Some proposed code improvements. Otherwise looks good.

Zotero/Scenes/Detail/PDF/Views/PDFDocumentViewController.swift
michalrentka: 6c880b9f408df98043278cb720159169778e2696

Zotero/Scenes/Detail/PDF/Views/PDFDocumentViewController.swift
michalrentka: The initial show preview function call still needs to be wrapped in `DispatchQueue.main.async` though, the initial state change is called too soon and without the delay the preview just doesn't show up - 32da1b737a859c087e1248a9bb1c65b4301665a6

PR #1020, opened 2024-10-21
Update dependencies

mvasilak:


Zotero/Controllers/IdentifierLookupController.swift
```diff
@@ -533,7 +533,7 @@ final class IdentifierLookupController {
                 let keys = lookupWebViewHandlersByLookupSettings.keys
                 for key in keys {
                     guard let webView = lookupWebViewHandlersByLookupSettings.removeValue(forKey: key)?.webViewHandler.webView else { continue }
-                    inMainThread {
+                    DispatchQueue.main.async {
```
michalrentka, line 536: Why change it? Does this have to be async even if it's already on main thread?

michalrentka (approved):

PR #1021, opened 2024-10-31
Improve item details UI performance

mvasilak:
Improves UI performance for item details. When there is a large number of sizable notes, e.g. ones created from annotations, scrolling becomes sluggish, and opening attachments from the details view causes the app to hang as the loading of the file is very slow.

### DIFF
--- a/Zotero/Scenes/Detail/ItemDetail/Views/ItemDetailViewController.swift
+++ b/Zotero/Scenes/Detail/ItemDetail/Views/ItemDetailViewController.swift
@@ -16,7 +16,9 @@ import RealmSwift
 import RxSwift
 
 private enum MainAttachmentButtonState {
-    case ready(String), downloading(String, CGFloat), error(String, Error)
+    case ready(String)
+    case downloading(String, CGFloat)
+    case error(String, Error)
 }
 
 final class ItemDetailViewController: UIViewController {
@@ -27,11 +29,26 @@ final class ItemDetailViewController: UIViewController {
     private let controllers: Controllers
     private let disposeBag: DisposeBag
 
-    private var collectionViewHandler: ItemDetailCollectionViewHandler!
+    lazy private var collectionViewHandler: ItemDetailCollectionViewHandler = {
+        let width = navigationController?.view.frame.width ?? view.frame.width
+        let collectionViewHandler = ItemDetailCollectionViewHandler(
+            collectionView: collectionView,
+            containerWidth: width,
+            viewModel: viewModel,
+            fileDownloader: controllers.userControllers?.fileDownloader
+        )
+        collectionViewHandler.delegate = self
+        collectionViewHandler.observer
+            .observe(on: MainScheduler.instance)
+            .subscribe(onNext: { [weak self] action in
+                self?.perform(collectionViewAction: action)
+            })
+            .disposed(by: disposeBag)
+        return collectionViewHandler
+    }()
     private var downloadingViaNavigationBar: Bool
-    private var didAppear: Bool
     var key: String {
-        return self.viewModel.state.key
+        return viewModel.state.key
     }
 
     weak var coordinatorDelegate: (DetailItemDetailCoordinatorDelegate & DetailNoteEditorCoordinatorDelegate)?
@@ -39,9 +56,8 @@ final class ItemDetailViewController: UIViewController {
     init(viewModel: ViewModel<ItemDetailActionHandler>, controllers: Controllers) {
         self.viewModel = viewModel
         self.controllers = controllers
-        self.downloadingViaNavigationBar = false
-        self.didAppear = false
-        self.disposeBag = DisposeBag()
+        downloadingViaNavigationBar = false
+        disposeBag = DisposeBag()
 
         super.init(nibName: "ItemDetailViewController", bundle: nil)
     }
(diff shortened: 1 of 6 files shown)


Zotero/Scenes/Detail/ItemDetail/Views/ItemDetailViewController.swift
```diff
@@ -53,36 +69,77 @@ final class ItemDetailViewController: UIViewController {
...
+
+            controllers.userControllers?.fileDownloader.observable
+                .observe(on: MainScheduler.instance)
+                .subscribe(onNext: { [weak self] update in
+                    guard let self else { return }
+                    viewModel.process(action: .updateDownload(update))
+
+                    guard viewModel.state.attachmentToOpen == update.key else { return }
+
+                    // 🍎 check logic of processes
```
michalrentka, line 111: Any reason to move from `// TODO` to `// 🍎`? :) Are these which you want to solve before closing this PR?

Zotero/Scenes/Detail/ItemDetail/Views/ItemDetailViewController.swift
```diff
@@ -53,36 +69,77 @@ final class ItemDetailViewController: UIViewController {
...
+                .disposed(by: self.disposeBag)
+        }
     }
 
-    override func viewDidAppear(_ animated: Bool) {
-        super.viewDidAppear(animated)
-        self.didAppear = true
+    override func viewIsAppearing(_ animated: Bool) {
+        super.viewIsAppearing(animated)
+        viewModel.process(action: .loadInitialData)
```
michalrentka, line 134: Any difference having it here rather than in `viewDidLoad`?

Zotero/Scenes/Detail/ItemDetail/Views/ItemDetailViewController.swift
```diff
@@ -53,36 +69,77 @@ final class ItemDetailViewController: UIViewController {
...
     }
 
     override func viewDidLayoutSubviews() {
         super.viewDidLayoutSubviews()
-
-        if !self.didAppear {
-            // Collapsed abstract is sometimes rendered incorrectly initially. The height of cell has proper height, but only 1 line is shown instead of 2.
-            self.collectionView.reloadData()
```
michalrentka, line 80: Has this been fixed? By you or iOS update? Can you check whether this still doesn't happen on older supported iOS?

Zotero/Scenes/Detail/ItemDetail/Views/ItemDetailViewController.swift
```diff
@@ -53,36 +69,77 @@ final class ItemDetailViewController: UIViewController {
...
-        }
-
-        guard let key = self.viewModel.state.preScrolledChildKey, self.collectionViewHandler.hasRows else { return }
-        self.collectionViewHandler.scrollTo(itemKey: key, animated: false)
-        self.viewModel.process(action: .clearPreScrolledItemKey)
+        // 🍎 move to is appearing? check where this `preScrolledChildKey` is set, only during start? we can probably lose clear action
```
michalrentka, line 139: This is used to focus on note/attachment when this view appears. It's set in state initially. Definitely movable to `viewIsAppearing`. The `.clearPreScrolledItemKey` probably can't be removed, otherwise on next `viewIsAppearing` it would re-focus on this key, which is not what we would want.

Zotero/Scenes/Detail/ItemDetail/Views/ItemDetailViewController.swift
```diff
@@ -142,63 +197,65 @@ final class ItemDetailViewController: UIViewController {
...
             }
             return
         }
 
         if state.changes.contains(.reloadedData) {
-            let wasHidden = self.collectionView.isHidden
-            self.collectionView.isHidden = state.isLoadingData
-            self.activityIndicator.isHidden = !state.isLoadingData
+            let wasHidden = collectionView.isHidden
+            print("🍎 changes = .reloadedData - wasHidden: \(wasHidden)")
```
michalrentka, line 252: Remove probably forgotten print

Zotero/Scenes/Detail/ItemDetail/Views/ItemDetailViewController.swift
```diff
@@ -142,63 +197,65 @@ final class ItemDetailViewController: UIViewController {
...
             return
         }
 
         if let error = state.error {
-            self.coordinatorDelegate?.show(error: error, viewModel: self.viewModel)
+            coordinatorDelegate?.show(error: error, viewModel: viewModel)
         }
 
         if state.changes.contains(.item) {
+            // 🍎 check why this is done
```
michalrentka, line 234: This state change calls `viewModel.process(action: .reloadData)` which causes another state change, which causes reentrancy issue in `RxSwift` (withing a subscription event you cause another subscription event). So I just `async` this so that it's called later. Didn't find a better way to deal with this.

michalrentka:
Just to confirm, the "performance improvement" was moving the slow parts to background? :) I didn't know we could do data source snapshot updates in background, doesn't it break anything? Otherwise if this helps then great.

michalrentka (changes requested):

mvasilak:
In this case, both this and the snapshot memory reduction are performance improvements, but the latter has a larger effect.

Using a background queue for diffable data source updates has been suggested since its introduction. The framework handles the snapshot application in the main queue, and all business logic can be safely done in the background.

michalrentka:
> Using a background queue for diffable data source updates has been suggested since its introduction. The framework handles the snapshot application in the main queue, and all business logic can be safely done in the background.

Good to know, we could definitely move all diffable data source interactions to background then. Feel free to create a ticket for it.

Zotero/Scenes/Detail/ItemDetail/Views/ItemDetailViewController.swift
mvasilak: No difference user-wise, it just simplifies the code by remving all needs for a `didAppear` varible.

Zotero/Scenes/Detail/ItemDetail/Views/ItemDetailViewController.swift
mvasilak: I somehow didn't delete all of my temporary comments & prints (I use an emoji for easier search). Sorry about that, fixed in https://github.com/zotero/zotero-ios/pull/1021/commits/a0358fce184823bf9b608541d5d6e69838c627bc

Zotero/Scenes/Detail/ItemDetail/Views/ItemDetailViewController.swift
mvasilak: `viewIsAppearing` is only called once, that's why I considered this simplification, but since the table may not have rows yet, I left it as is in the end.

Zotero/Scenes/Detail/ItemDetail/Views/ItemDetailViewController.swift
mvasilak: Seems ok as far as iOS 15.

michalrentka:
(Sorry, I added comments some time ago and forgot to submit them)

Zotero/Scenes/Detail/ItemDetail/Views/ItemDetailViewController.swift
mvasilak: Yes, my bad, I'll fix it.

mvasilak: Check https://github.com/zotero/zotero-ios/pull/1021/commits/4379c399244ba73fb0770afda77bc7d95f14e28b

Zotero/Scenes/Detail/ItemDetail/Views/ItemDetailViewController.swift
mvasilak: Fixed in https://github.com/zotero/zotero-ios/pull/1021/commits/4379c399244ba73fb0770afda77bc7d95f14e28b

michalrentka: Wait so this didn't work in `viewIsAppearing`? You could definitely move it there and get rid of `viewDidLayoutSubviews`. I just said you probably shouldn't remove the `. clearPreScrolledItemKey` action, otherwise this would be repeated on each appearance.

michalrentka (approved):

Zotero/Scenes/Detail/ItemDetail/Views/ItemDetailViewController.swift
mvasilak: `viewIsAppearing` is called before the collection view has fully reloaded.

PR #1028, opened 2024-11-20
Implemented custom idle timer logic

### LINKED ISSUE #1017: Remove auto locking setting from reader
Remove the "Allow device to sleep" setting from reader settings. Automatically disable sleeping for 25 minutes of no user interaction. After 25 minutes re-enable the system mechanism and let the device sleep.

michalrentka:
Closes #1017

### DIFF
--- a/Zotero/Controllers/IdleTimerController.swift
+++ b/Zotero/Controllers/IdleTimerController.swift
@@ -21,7 +21,18 @@ final class IdleTimerController {
 
     init() {
         disposeBag = DisposeBag()
+        observeOrientationChange()
         observeLowPowerMode()
+
+        func observeOrientationChange() {
+            NotificationCenter.default.rx
+                .notification(UIDevice.orientationDidChangeNotification)
+                .observe(on: MainScheduler.instance)
+                .subscribe(onNext: { [weak self] _ in
+                    self?.resetCustomTimer()
+                })
+                .disposed(by: disposeBag)
+        }
 
         func observeLowPowerMode() {
             NotificationCenter.default.rx
@@ -54,6 +65,7 @@ final class IdleTimerController {
             guard activeTimer == nil else { return }
             set(disabled: true)
             startTimer(controller: self)
+            UIDevice.current.beginGeneratingDeviceOrientationNotifications()
         }
 
         func startTimer(controller: IdleTimerController) {
@@ -91,6 +103,7 @@ final class IdleTimerController {
              */
             activeTimer?.resume()
             activeTimer = nil
+            UIDevice.current.endGeneratingDeviceOrientationNotifications()
         }
     }
 
--- a/Zotero/Controllers/IdleTimerController.swift
+++ b/Zotero/Controllers/IdleTimerController.swift
@@ -8,35 +8,96 @@
 
 import Foundation
 import UIKit
+
 import CocoaLumberjackSwift
+import RxSwift
 
 final class IdleTimerController {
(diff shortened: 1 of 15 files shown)


Zotero/Controllers/IdleTimerController.swift
```diff
@@ -8,35 +8,96 @@
...
     }
+
+    private func set(disabled: Bool) {
+        UIApplication.shared.isIdleTimerDisabled = disabled
+    }
+
+    private func forceStopIdleTimer() {
+        DDLogInfo("IdleTimerController: force stop timer")
+        activeProcesses = 0
+        stopCustomIdleTimer()
```
mvasilak, line 101: This call will fail in the first guard of `stopCustomIdleTimer`

Zotero/Controllers/IdleTimerController.swift
```diff
@@ -8,35 +8,96 @@
 
 import Foundation
 import UIKit
+
 import CocoaLumberjackSwift
+import RxSwift
 
 final class IdleTimerController {
+    private static let customIdleTimerTimemout = 1200
```
mvasilak, line 16: `Timemeout` has a typo (2nd m). Also, why not make it directly
```suggestion
    private static let customIdleTimerTimeout: DispatchTimeInterval = .seconds(1200)
```

Zotero/Controllers/IdleTimerController.swift
```diff
@@ -8,35 +8,96 @@
...
+            controller.activeTimer = timer
         }
     }
 
-    func enable() {
+    func stopCustomIdleTimer() {
         inMainThread { [weak self] in
-            guard let self = self else { return }
-            guard self.activeProcesses > 0 else {
+            guard let self, activeProcesses > 0 else {
```
mvasilak, line 72: If `activeProcesses` is `0`, log a warning but continue to stop the timer.

Zotero/EventObservableWindow.swift
```diff
@@ -0,0 +1,17 @@
...
+//  Created by Michal Rentka on 20.11.2024.
+//  Copyright © 2024 Corporation for Digital Scholarship. All rights reserved.
+//
+
+import UIKit
+
+final class EventObservableWindow: UIWindow {
+    override func sendEvent(_ event: UIEvent) {
+        super.sendEvent(event)
+        guard !(event.allTouches ?? []).isEmpty else { return }
```
mvasilak, line 14: Are other events that we should consider, that may be applicable here, like device rotation?

Zotero/Controllers/IdleTimerController.swift
michalrentka: Fixed in b81703099fea811fc4dda794b99a9b6c4a36cc17

Zotero/Controllers/IdleTimerController.swift
michalrentka: Fixed in ec21879cebda856ecb862bc78281538db6705d31

Zotero/Controllers/IdleTimerController.swift
michalrentka: Fixed in ec21879cebda856ecb862bc78281538db6705d31

mvasilak (approved):

Zotero/EventObservableWindow.swift
michalrentka: Added in d4b36edc9e401ac3c1b89c67f1705ab1898c2593

PR #1035, opened 2024-12-03
Adjust free text annotation size on editing end

mvasilak:
Should improve sync between different clients, for those text annotations that are created in iOS.

### DIFF
--- a/Zotero/Scenes/Detail/PDF/Views/FreeTextAnnotationView.swift
+++ b/Zotero/Scenes/Detail/PDF/Views/FreeTextAnnotationView.swift
@@ -1,5 +1,5 @@
 //
-//  CustomFreeTextAnnotationView.swift
+//  FreeTextAnnotationView.swift
 //  Zotero
 //
 //  Created by Michal Rentka on 02.08.2023.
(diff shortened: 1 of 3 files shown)


Zotero/Scenes/Detail/PDF/Views/FreeTextAnnotationView.swift
```diff
@@ -37,6 +37,22 @@ final class CustomFreeTextAnnotationView: FreeTextAnnotationView {
...
+        let previousTextBoundingBox = annotation.textBoundingBox
+        super.textViewDidEndEditing(textView)
+        annotation.sizeToFit()
+        var newTextBoundingBox = annotation.textBoundingBox
+        let minX = min(previousTextBoundingBox.minX, newTextBoundingBox.minX)
+        let maxX = max(previousTextBoundingBox.maxX, newTextBoundingBox.maxX)
+        let minY = min(previousTextBoundingBox.minY, newTextBoundingBox.minY)
+        let maxY = max(previousTextBoundingBox.maxY, newTextBoundingBox.maxY)
+        annotation.textBoundingBox = CGRect(x: minX, y: minY, width: maxX - minX, height: maxY - minY)
+        NotificationCenter.default.post(name: .PSPDFAnnotationChanged, object: annotation, userInfo: [PSPDFAnnotationChangedNotificationKeyPathKey: ["boundingBox", "fontSize"]])
```
michalrentka, line 52: Doesn't this cause unnecessary sync? Or is this quick enough to be "hidden" in debounce? I mean we probably can't do anything about it either way, but I saw it doesn't completely fix the issue anyway.

mvasilak: From my testing it's quick enough, but there could be edge cases where it requires an extra sync. The end result is certainly improved, however, there still seems to be some minor client ambiguity.

PR #1045, opened 2024-12-13
Fix transformed annotation user

mvasilak:
Fixes issue reported in forum [post](https://forums.zotero.org/discussion/118984/autolocking-annotations-on-ipad)

When an annotation is transformed, e.g. due to overlapping rects or rects that need to be split, the new annotations are added using `document.add`. If these annotations have an empty string for the `user` field, the field is converted to `nil`. This will cause the PSPDFKIT annotation to be added in the database with `createdBy = nil`, until the next sync happens, that will assign the proper user.

Guarantying a non-empty `user` field, alleviates the issue. Alternatively, we can consider adding the `username` or event better the `userId`, in the annotation `customData`, and not rely at all to `user`.

### DIFF
--- a/Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
+++ b/Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
@@ -1370,7 +1370,7 @@ final class PDFReaderActionHandler: ViewModelActionHandler, BackgroundDbProcessi
         for annotation in finalAnnotations {
             if annotation.key == nil {
                 // We use the displayName, but if this is empty we use the username, which is what would be presented anyway.
-                // Since a username cannot be empty, we guarranty an non-empty annotation.user field.
+                // Since a username cannot be empty, we guarantee an non-empty annotation.user field.
                 annotation.user = viewModel.state.displayName.isEmpty ? viewModel.state.username : viewModel.state.displayName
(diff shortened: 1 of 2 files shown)


Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
```diff
@@ -1369,7 +1369,9 @@ final class PDFReaderActionHandler: ViewModelActionHandler, BackgroundDbProcessi
         let finalAnnotations = keptAsIs + toAdd
         for annotation in finalAnnotations {
             if annotation.key == nil {
-                annotation.user = viewModel.state.displayName
+                // We use the displayName, but if this is empty we use the username, which is what would be presented anyway.
+                // Since a username cannot be empty, we guarranty an non-empty annotation.user field.
```
michalrentka, line 1373: `guarantee` (or `guaranty`), not `guarranty`

michalrentka (approved):

Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift
mvasilak: Oops, check https://github.com/zotero/zotero-ios/pull/1045/commits/50676c641f1c50bd1b589ba982c950c3ea01900e

PR #1050, opened 2025-01-09
Changing between highlight and underline annotations

### LINKED ISSUE #985: Allow annotations to be changed between highlights and underlines via sync
See https://github.com/zotero/zotero/issues/3744#issuecomment-2235951805

Any reason this would fail if it came in via a sync? For now we're going to delete and recreate to change between the two types, but we'll want to be able to switch to just changing the annotation type later.

(And once we have underlining merged, we'll want a UI option to change between highlights and underlines, but we can deal with that later. This is just to make sure there are no problems at the data layer.)

michalrentka:
Closes #985

### DIFF
--- a/Zotero.xcodeproj/project.pbxproj
+++ b/Zotero.xcodeproj/project.pbxproj
@@ -889,6 +889,7 @@
 		B3A17D1927FC33B800322CAD /* LowPowerModeController.swift in Sources */ = {isa = PBXBuildFile; fileRef = B3A17D1827FC33B800322CAD /* LowPowerModeController.swift */; };
 		B3A27ACD25BEE91A00DE0BB2 /* TranslationWebViewHandler.swift in Sources */ = {isa = PBXBuildFile; fileRef = B3A27ACC25BEE91A00DE0BB2 /* TranslationWebViewHandler.swift */; };
 		B3A27AD125BEE93400DE0BB2 /* FilenameFormatter.swift in Sources */ = {isa = PBXBuildFile; fileRef = B3A27AD025BEE93400DE0BB2 /* FilenameFormatter.swift */; };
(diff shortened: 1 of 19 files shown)


mvasilak (changes requested):
* Tapping the "Convert to ..." cell, and this changing without actually changing the annotation seems unintuitive. Maybe a cell with 2 radio buttons instead? Or make it an action in the `...` menu of the annotation instead of a cell, so it can be executed immediatelly.
* The colors for the highlight and the underline shouls be adopted for the annotation type. Converting an underline to a highlight results in a slightly darker color.
* Converting an annotation in the sidebar results in a jumpy animation and closes the reader.

michalrentka:
> * Tapping the "Convert to ..." cell, and this changing without actually changing the annotation seems unintuitive. Maybe a cell with 2 radio buttons instead? Or make it an action in the `...` menu of the annotation instead of a cell, so it can be executed immediatelly.
Updated based on #zotero discussion, changed to `UISegmentedControl` as you suggested.

> * The colors for the highlight and the underline shouls be adopted for the annotation type. Converting an underline to a highlight results in a slightly darker color.
I saw this before but I can't reproduce now. Anyway I do update the color of annotation when the type changes, so that shouldn't be the issue.

> * Converting an annotation in the sidebar results in a jumpy animation and closes the reader.
Fixed.

Check 52183050b2b6e5481305950a07387071f03c93bc

mvasilak (approved):
Can the cell change be made with a reconfigure instead of deletion and insertion, for a smoother animation?

michalrentka:
> Can the cell change be made with a reconfigure instead of deletion and insertion, for a smoother animation?

I tried to mitigate this, but since we're actually deleting one annotation and creating a new one with new key and Realm doesn't provide a good way to simply ignore a specific notification tied to those actions, I don't have a good way to change it. 
I could temporarily store the old and new key, wait for those 2 notifications and then just ignore original actions and create a refresh based on that data. But in case something else comes inbetween (it shouldn't, but yeah I wouldn't want to risk it), it'll crash. I'm not sure whether it's worth the work.

PR #1070, opened 2025-02-13
Added sepia mode to PDF reader

### LINKED ISSUE #1068: Add color themes to reader settings popup
As on desktop, to allow switching between light mode and dark mode regardless of system setting

We should also add support for Sepia, which exists in PDF Viewer and is now supported in the EPUB reader. (We may have to leave out "Snow", since — assuming they don't allow custom color schemes — there's no equivalent in the PDF reader.)

### LINKED ISSUE #1069: Add sepia mode to PDF reader
Didn't notice this is a duplicate of #1068, closing

michalrentka:
Closes #1068


Zotero/Scenes/Detail/PDF/Views/PDFDocumentViewController.swift
```diff
@@ -216,7 +216,7 @@ final class PDFDocumentViewController: UIViewController {
     }
 
     private func update(state: PDFReaderState, pdfController: PDFViewController) {
-        if state.changes.contains(.interfaceStyle) {
+        if state.changes.contains(.appearanceMode) {
```
mvasilak, line 219: Previous check should remain, otherwise while on Automatic it misses system change in appearance mode.
```suggestion
        if state.changes.contains(.interfaceStyle) || state.changes.contains(.appearanceMode) {
```

mvasilak (changes requested):
Shouldn't thumbnail & annotation previews also change when sepia is selected?

mvasilak (approved):

Issue #1046, opened 2024-12-18
PDF metadata retrieval


dstillman:
This should be possible with pdf-worker's `getRecognizerData` and identifier lookup.

web-library: https://github.com/zotero/web-library/issues/583
Android: https://github.com/zotero/zotero-android/issues/203

tnajdek:
I believe we should also rename the attachment item after metadata retrieval. Here is my comment on this topic in the zotero-android repo:

https://github.com/zotero/zotero-android/issues/225

PR #1071, opened 2025-02-13
Add PDF worker support

mvasilak:
* Adds support for PDF worker via `PDFWorkerWebViewHandler`.
* Adds `PDFWorkerController` that exposes PDF worker for extracting recognizer data and full text.
* Adds `RecognizerController` that extracts recognizer data, performs remote recognizer request, and finally performs identifier lookup to produce an item response, for selected local PDF file. (Logic for remote recognizer response processing is copied from https://github.com/zotero/zotero/blob/7b7a0b9e4692a0ca32984b763f67b7b10edf96fd/chrome/content/zotero/xpcom/recognizeDocument.js#L411)
* Uses `RecognizerController` in share extension, to create parent item for a shared local PDF file.
* Renames recognized PDF files according to https://github.com/zotero/zotero-ios/issues/1046#issuecomment-2706453811

TODO:
- [x] Use recognizer with `Retrieve Metadata` action for standalone attachments.
- [x] Use recognizer for standalone files added within the app.
- [x] Show in UI that there is metadata retrieval in progress for affected items.
- [x] Rename item and/or attachment file if needed after successful metadata retrieval.
- [x] Use PDF worker feature gate.

To be considered:
- [ ] Add popup window with all metadata retrievals' progress, that will be minimized in the items bottom toolbar.
- [ ] Allow canceling metadata retrievals.


mvasilak:
@michalrentka @dstillman this is ready for a first review / user testing. I did a very simple UI iteration, where items that are retrieving metadata, are indicating this by an animating subtitle, as it is empty for attachment type. If functionality is as desired, we can improve the UI as we see fit.

Zotero/Controllers/PDFWorkerController.swift
```diff
@@ -0,0 +1,213 @@
...
+    private let dispatchSpecificKey: DispatchSpecificKey<String>
+    private let accessQueueLabel: String
+    private let accessQueue: DispatchQueue
+    private let disposeBag: DisposeBag
+
+    internal weak var webViewProvider: WebViewProvider?
+
+    // Accessed only via accessQueue
+    private static let maxConcurrentPDFWorkers: Int = 1
+    private var queue: OrderedDictionary<PDFWork, (state: PDFWorkState, observable: PublishSubject<Update>)> = [:]
```
michalrentka, line 56: Since we have separate `queue` and `pdfWorkerWebViewHandlersByPDFWork` already, can `queue` be just an actual queue and `pdfWorkerWebViewHandlersByPDFWork` would be a dictionary containing `Observable`s/`PublishSubject`s and `PDFWorkerWebViewHandler`s which are currently in progress? It would simplify the queue to an array `[PDFWork]` and `inProgress` (or whatever name) would be `[PDFWork: (Observable<Update>, PDFWorkerWebViewHandler)]`.

When starting new work, you'd just remove last `PDFWork` from the queue, create an observable and web view handler and add them to the dictionary`. When the work is done/cancelled you'll just call the handler to cleanup the webview and then remove the dictionary entry.

Zotero/Controllers/RecognizerController.swift
```diff
@@ -0,0 +1,585 @@
...
+    var updates: Observable<Update> {
+        updatesSubject.asObservable()
+    }
+    private let disposeBag: DisposeBag
+
+    internal weak var webViewProvider: WebViewProvider?
+
+    // Accessed only via accessQueue
+    private static let maxConcurrentRecognizerTasks: Int = 1
+    private var queue: OrderedDictionary<RecognizerTask, (state: RecognizerTaskState, observable: PublishSubject<Update>)> = [:]
```
michalrentka, line 139: Maybe consider having separate "wait queue" and "in progress dictionary", same as above. But it's not as apparent here, so we might as well just keep this.

Zotero/Controllers/Web View Handling/PDFWorkerWebViewHandler.swift
```diff
@@ -0,0 +1,251 @@
...
+        case .fullText:
+            guard let data = (body as? [String: Any])?["fullText"] as? [String: Any] else { return }
+            observable.on(.next(.success(.recognizerData(data: data))))
+
+        case .log:
+            DDLogInfo("JSLOG: \(body)")
+        }
+    }
+
+    static let htmlString: String = """
```
michalrentka, line 169: Put this in a file somewhere, no need to hardcore whole html + javascript files here.

michalrentka (changes requested):
Didn't go through everything yet, but there will probably be bigger changes, so I'll have to go through most of it again anyway.

Zotero/Controllers/Controllers.swift
```diff
@@ -274,6 +274,8 @@ final class Controllers {
         controllers?.remoteFileDownloader.stop()
         // Cancel all background uploads
         controllers?.backgroundUploadObserver.cancelAllUploads()
+        // Cancel all PDF workers
+        controllers?.pdfWorkerController.cancellAllWorks()
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/368394cc5c26e270f7e18582b91f65d1860eb7bb

Zotero/Controllers/PDFWorkerController.swift
```diff
@@ -0,0 +1,213 @@
...
+            queue[work] = (state, observable)
+            completion(observable)
+
+            startWorkIfNeeded()
+        }
+    }
+
+    private func startWorkIfNeeded() {
+        guard pdfWorkerWebViewHandlersByPDFWork.count < Self.maxConcurrentPDFWorkers else { return }
+        let works = queue.keys
```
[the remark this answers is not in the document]
mvasilak: AFAICS, this or a similar syntax is not supported by `OrderedDictionary`, but since the for loop exits when it finds the desired item, the end result is the same.

Zotero/Controllers/PDFWorkerController.swift
```diff
@@ -0,0 +1,213 @@
...
+        } else {
+            accessQueue.async(flags: .barrier) {
+                cleanup(for: work, completion: completion)
+            }
+        }
+
+        func cleanup(for work: PDFWork, completion: @escaping (_ observable: PublishSubject<Update>?) -> Void) {
+            let observable = queue.removeValue(forKey: work).flatMap({ $0.observable })
+            DDLogInfo("PDFWorkerController: cleaned up for \(work)")
+            if let webView = pdfWorkerWebViewHandlersByPDFWork.removeValue(forKey: work)?.webViewHandler.webView {
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/2213f0199ecb441e168f8eace248c3d074a35d30

ZShare/ViewModels/ExtensionViewModel.swift
```diff
@@ -501,17 +516,59 @@ final class ExtensionViewModel {
...
+        func updateState(with attachmentState: State.AttachmentState) {
+            var state = self.state
+            state.processedAttachment = .file(file: tmpFile, filename: filename)
+            state.expectedAttachment = (filename, tmpFile)
+            state.attachmentState = attachmentState
+            self.state = state
+        }
+
+        func recognize(file: FileData) {
+            recognizerController.queue(task: RecognizerController.RecognizerTask(file: file, kind: .simple)) { [weak self] observable in
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/af7941dd2eff2dd790b24910b575e42399bf3a76

Zotero/Controllers/PDFWorkerController.swift
```diff
@@ -0,0 +1,213 @@
...
+        case inProgress
+    }
+
+    // MARK: Properties
+    private let dispatchSpecificKey: DispatchSpecificKey<String>
+    private let accessQueueLabel: String
+    private let accessQueue: DispatchQueue
+    private let disposeBag: DisposeBag
+
+    internal weak var webViewProvider: WebViewProvider?
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/af7941dd2eff2dd790b24910b575e42399bf3a76

Zotero/Controllers/PDFWorkerController.swift
```diff
@@ -0,0 +1,213 @@
...
+                }
+            }
+            guard let pdfWorkerWebViewHandler else {
+                DDLogError("PDFWorkerController: can't create PDFWorkerWebViewHandler instance")
+                cleanupPDFWorker(for: work) { observable in
+                    observable?.on(.next(Update(work: work, kind: .failed)))
+                }
+                return
+            }
+            pdfWorkerWebViewHandlersByPDFWork[work] = pdfWorkerWebViewHandler
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/af7941dd2eff2dd790b24910b575e42399bf3a76

Zotero/Controllers/PDFWorkerController.swift
```diff
@@ -0,0 +1,213 @@
...
+            case .recognizer:
+                pdfWorkerWebViewHandler.recognize(file: work.file)
+
+            case .fullText:
+                pdfWorkerWebViewHandler.getFullText(file: work.file)
+            }
+
+            func setupObserver(for pdfWorkerWebViewHandler: PDFWorkerWebViewHandler) {
+                pdfWorkerWebViewHandler.observable
+                    .subscribe(onNext: { process(result: $0) })
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/af7941dd2eff2dd790b24910b575e42399bf3a76

Zotero/Controllers/RecognizerController.swift
```diff
@@ -0,0 +1,585 @@
...
+import Foundation
+import WebKit
+import OrderedCollections
+
+import CocoaLumberjackSwift
+import RxSwift
+
+final class RecognizerController {
+    // MARK: Types
+    enum RemoteRecognizerIdentifier {
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/7c81b9a49dacfd9090892aa33261b15ebeb169dc

Zotero/Controllers/PDFWorkerController.swift
```diff
@@ -0,0 +1,213 @@
...
+    init() {
+        dispatchSpecificKey = DispatchSpecificKey<String>()
+        accessQueueLabel = "org.zotero.PDFWorkerController.accessQueue"
+        accessQueue = DispatchQueue(label: accessQueueLabel, qos: .userInteractive, attributes: .concurrent)
+        accessQueue.setSpecific(key: dispatchSpecificKey, value: accessQueueLabel)
+        disposeBag = DisposeBag()
+    }
+
+    // MARK: Actions
+    func queue(work: PDFWork, completion: @escaping (_ observable: PublishSubject<Update>?) -> Void) {
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/4172e5d021e6c8962b1aaebc14d5cc6e651c455b for

Zotero/Controllers/PDFWorkerController.swift
```diff
@@ -0,0 +1,213 @@
...
+    init() {
+        dispatchSpecificKey = DispatchSpecificKey<String>()
+        accessQueueLabel = "org.zotero.PDFWorkerController.accessQueue"
+        accessQueue = DispatchQueue(label: accessQueueLabel, qos: .userInteractive, attributes: .concurrent)
+        accessQueue.setSpecific(key: dispatchSpecificKey, value: accessQueueLabel)
+        disposeBag = DisposeBag()
+    }
+
+    // MARK: Actions
+    func queue(work: PDFWork, completion: @escaping (_ observable: Observable<Update>?) -> Void) {
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/4172e5d021e6c8962b1aaebc14d5cc6e651c455b

Zotero/Controllers/RecognizerController.swift
```diff
@@ -0,0 +1,585 @@
...
+        accessQueueLabel = "org.zotero.RecognizerController.accessQueue"
+        accessQueue = DispatchQueue(label: accessQueueLabel, qos: .userInteractive, attributes: .concurrent)
+        accessQueue.setSpecific(key: dispatchSpecificKey, value: accessQueueLabel)
+        backgroundQueue = DispatchQueue(label: "org.zotero.RecognizerController.backgroundQueue", qos: .userInitiated)
+        updatesSubject = PublishSubject()
+        disposeBag = DisposeBag()
+    }
+
+    // MARK: Actions
+    func queue(task: RecognizerTask, completion: ((_ observable: Observable<Update>?) -> Void)? = nil) {
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/4172e5d021e6c8962b1aaebc14d5cc6e651c455b

Zotero/Controllers/PDFWorkerController.swift
```diff
@@ -0,0 +1,213 @@
...
+            guard let self else {
+                completion(nil)
+                return
+            }
+            if let (_, observable) = queue[work] {
+                completion(observable.asObservable())
+                return
+            }
+            let state: PDFWorkState = .enqueued
+            let observable: PublishSubject<Update> = PublishSubject()
```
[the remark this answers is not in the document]
michalrentka: I understand, but my point was you could use `Observable` directly, no need to use `PublishSubject` here. Anyway I see you already changed that 👍

Zotero/Controllers/PDFWorkerController.swift
```diff
@@ -0,0 +1,207 @@
...
+            guard let self else { return Disposables.create() }
+            accessQueue.async(flags: .barrier) { [weak self] in
+                guard let self else { return }
+                if let subject = queue[work]?.1 {
+                    subject.subscribe(subscriber)
+                        .disposed(by: disposeBag)
+                    return
+                }
+                let state: PDFWorkState = .enqueued
+                let subject: PublishSubject<Update> = PublishSubject()
```
michalrentka, line 79: Wait, why create a `PublishSubject` here at all? Just store the `subscriber` block and call that on state change. The `queue` should store the `subscriber` and not `PublishSubject`.

Zotero/Controllers/RecognizerController.swift
```diff
@@ -0,0 +1,585 @@
...
+            emmitUpdate(for: task, observable: observable, kind: .enqueued)
+            startRecognitionIfNeeded()
+        }
+    }
+
+    private func emmitUpdate(for task: RecognizerTask, observable: PublishSubject<Update>, kind: Update.Kind) {
+        let update = Update(task: task, kind: kind)
+        if case .createParentForItem(let libraryId, let key) = task.kind {
+            var libraryLatestUpdates = latestUpdates[libraryId, default: [:]]
+            libraryLatestUpdates[key] = kind
```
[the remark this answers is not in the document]
michalrentka: Actually ignore this, since you're returning `Observable` in `func queue(task: Task) -> Observable<Update>` you should just store the subscriber in `queue`, same as in `PDFWorkerController` and store state separately as you're doing already. No need for `BehaviorSubject`.

Zotero/Controllers/PDFWorkerController.swift
mvasilak: I'd argue that it's properly valid to have multiple works, each one with their own set of multiple subscribers. As it is also valid of course to have a single subject that publishes updates for all works, and then let the subscribers filter what they want. Both approaches have their own merits and I just opted to offer both, as I was fleshing out the end uses of the controllers.

In any case, I'll simplify it as you suggest, since we mostly start work in one place, and have subscribers in another.

michalrentka: > I'd argue that it's properly valid to have multiple works, each one with their own set of multiple subscribers. 

That's valid of course. I just didn't like the implementation. At first you had a closure which contained the `Observable`, now you create an `Observable` which creates a `PublishSubject` internally, subscribes to it and forwards it's messages. Both solutions feel clunky. 

I didn't know there are multiple subscribers, so I suggested simply returning the `Observable` instead of all that. Now that I know, another simpler way would be just to synchronously return created `PublishSubject` and asynchronously store it as needed. No need for a block or observable object, you can just immediately subscribe to the `PublishSubject` (converted to `Observable`)

mvasilak: We use the `accessQueue` to access existing subjects, or store new ones. If we create and return a subject in sync, we might find in the async check that one already exists for this work/task. In that case we could just link it to the existing one by subscribing to it. Would that be ok?

michalrentka: Can't we simply do this?
```
func queue(work: PDFWork) -> Observable<Update> {
        if let observable = accessQueue.sync(execute: { queue[work]?.1.asObservable() }) {
            return observable
        }
        let subject: PublishSubject<Update> = PublishSubject()
        accessQueue.async(flags: .barrier) { [weak self] in
            queue[work] = (.enqueued, subject)
            startWorkIfNeeded()
        }
        return subject.asObservable()
    }
```

mvasilak: I think we could still have a race issue, where two callers reach the subject creation, and therefore the last one that goes to the async block will overwrite the other. 

As I mentioned, our purposes are covered by a single subject so I'll refactor to what you suggested before, to be on the safe side. If we ever need to have something more complex, we can revisit it then.

mvasilak: I spoke a bit too fast, having just a single subejct for `PDFWorkerController` would complicate things a lot (for `RecognizerController` not too much). So what I suggest is this:
```
    func queue(work: PDFWork) -> Observable<Update> {
        let subject = PublishSubject<Update>()
        accessQueue.async(flags: .barrier) { [weak self] in
            if let existingSubject = queue[work]?.1 {
                existingSubject.subscribe(subject)
                    .disposed(by: disposeBag)
                return
            }
            let state: PDFWorkState = .enqueued
            queue[work] = (state, subject)

            startWorkIfNeeded()
        }
        return subject.asObservable()
    }
```
It returns a subject as observable at once, and in the edge case where there are more subscribers for the same work, it's just subscribes to the existing subject, acting as a simple relay. WDYT?

Zotero/Controllers/Web View Handling/PDFWorkerWebViewHandler.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/89bcec05fc56f567b7e7b566e7744c5eb9de59c2

Zotero/Controllers/RecognizerController.swift
```diff
@@ -0,0 +1,585 @@
...
+    private func enqueueNextIdentifierLookup(for task: RecognizerTask) {
+        if DispatchQueue.getSpecific(key: dispatchSpecificKey) == accessQueueLabel {
+            _enqueueNextIdentifierLookup(for: task)
+        } else {
+            accessQueue.async(flags: .barrier) {
+                _enqueueNextIdentifierLookup(for: task)
+            }
+        }
+
+        func _enqueueNextIdentifierLookup(for task: RecognizerTask) {
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/3a80543ec1269cb0247ecfb8acc9bcc6cb257725

Zotero/Controllers/PDFWorkerController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/6bb03ba5e81e962b6971061ed64e2a9d3438ba95
It's easier to just maintain separate dictionaries for `PublishSubject`, `PDFWorkerWebViewHandler`, and `PDFWorkState`, but the end result is the same.

Zotero/Controllers/RecognizerController.swift
```diff
@@ -0,0 +1,585 @@
...
+            if let (_, observable) = queue[task] {
+                completion?(observable.asObservable())
+                return
+            }
+            let state: RecognizerTaskState = .enqueued
+            let observable: PublishSubject<Update> = PublishSubject()
+            queue[task] = (state, observable)
+            completion?(observable.asObservable())
+            observable.subscribe(onNext: { [weak self] update in
+                self?.updatesSubject.on(.next(update))
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/266e2af98ac90709b36958fa3baee2477eaddd86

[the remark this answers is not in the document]
mvasilak: Actually reverted in https://github.com/zotero/zotero-ios/pull/1071/commits/ec93630ff821797662751ab8f267d9e79b4c213c 
Having to emmit two the same update from two separate subjects would complicate the cleanup logic, while now updates of a task are transparently relayed to the general controller updates.

Zotero/Controllers/RecognizerController.swift
```diff
@@ -0,0 +1,585 @@
...
+
+    struct Update {
+        enum Kind {
+            case failed(Error)
+            case cancelled
+            case enqueued
+            case recognitionInProgress
+            case remoteRecognitionInProgress(data: [String: Any])
+            case identifierLookupInProgress(response: RemoteRecognizerResponse, identifier: String)
+            case translated(itemResponse: ItemResponse)
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/8b825ea98e2d183a51a9d48ea014eea6c008f7e6

Zotero/Controllers/RecognizerController.swift
```diff
@@ -0,0 +1,548 @@
...
+        let (task, subject) = queue.removeFirst()
+        start(task: task, subject: subject)
+
+        func start(task: Task, subject: PublishSubject<Update>) {
+            subjectsByTask[task] = subject
+            statesByTask[task] = .recognitionInProgress
+            emmitUpdate(for: task, subject: subject, kind: .inProgress)
+
+            pdfWorkerController.queue(work: PDFWorkerController.PDFWork(file: task.file, kind: .recognizer))
+                .subscribe(onNext: { update in
```
michalrentka, line 183: `[weak self]` and pass it to `process` sub-function

Zotero/Controllers/RecognizerController.swift
```diff
@@ -0,0 +1,548 @@
...
+            guard let subject = subjectsByTask[task] else {
+                startRecognitionIfNeeded()
+                return
+            }
+            statesByTask[task] = .remoteRecognitionInProgress(data: data)
+
+            apiClient.send(request: RecognizerRequest(parameters: data)).subscribe(
+                onSuccess: { (response: (RemoteRecognizerResponse, HTTPURLResponse)) in
+                    DDLogInfo("RecognizerController: \(task) - remote recognizer response received")
+                    process(response: response.0)
```
michalrentka, line 230: `[weak self]` and pass to `process` sub function

Zotero/Controllers/RecognizerController.swift
```diff
@@ -0,0 +1,548 @@
...
+            guard case .identifiersLookupInProgress(let response, _, let pendingIdentifiers) = state else { return nil }
+            return (response, pendingIdentifiers)
+        }
+    ) {
+        if DispatchQueue.getSpecific(key: dispatchSpecificKey) == accessQueueLabel {
+            _enqueueNextIdentifierLookup(for: task)
+        } else {
+            accessQueue.async(flags: .barrier) { [weak self] in
+                guard let self else { return }
+                _enqueueNextIdentifierLookup(for: task)
```
michalrentka, line 280: Pass `self` to `_enqueueNextIdentifierLookup`, it's unused now.

Zotero/Controllers/RecognizerController.swift
```diff
@@ -0,0 +1,548 @@
...
+                    DDLogWarn("RecognizerController: \(task) - can't create LookupWebViewHandler instance")
+                    return nil
+                }
+                lookupWebViewHandlersByTask[task] = lookupWebViewHandler
+                setupObserver(for: lookupWebViewHandler)
+                return lookupWebViewHandler
+
+                func setupObserver(for lookupWebViewHandler: LookupWebViewHandler) {
+                    lookupWebViewHandler.observable
+                        .subscribe(onNext: { process(result: $0) })
```
michalrentka, line 349: `[weak self]` and pass to `process` sub-function

Zotero/Controllers/RecognizerController.swift
```diff
@@ -0,0 +1,548 @@
...
+        return Maybe.create { [weak self] maybe in
+            guard let self else {
+                maybe(.completed)
+                return Disposables.create()
+            }
+            if DispatchQueue.getSpecific(key: dispatchSpecificKey) == accessQueueLabel {
+                cleanup(for: task, maybe: maybe)
+            } else {
+                accessQueue.async(flags: .barrier) {
+                    cleanup(for: task, maybe: maybe)
```
michalrentka, line 516: `[weak self]` and pass to `cleanup` sub-function (also above on 513)

Zotero/Controllers/Web View Handling/PDFWorkerWebViewHandler.swift
```diff
@@ -0,0 +1,178 @@
...
+                }
+            })
+            .disposed(by: disposeBag)
+        }
+    }
+
+    func recognize(file: FileData) {
+        let filePath = file.createUrl().path
+        performCall {
+            performRecognize(for: filePath)
```
michalrentka, line 125: `[weak self]` and pass to `performRecognize` sub-function

Zotero/Controllers/Web View Handling/PDFWorkerWebViewHandler.swift
```diff
@@ -0,0 +1,178 @@
...
+                    self?.observable.on(.next(.failure(error)))
+                })
+                .disposed(by: disposeBag)
+        }
+    }
+
+    func getFullText(file: FileData) {
+        let filePath = file.createUrl().path
+        performCall {
+            performGetFullText(for: filePath)
```
michalrentka, line 144: `[weak self]` and pass to `performGetFullText` sub-function

Zotero/Controllers/PDFWorkerController.swift
```diff
@@ -0,0 +1,203 @@
...
+
+            case .fullText:
+                pdfWorkerWebViewHandler.getFullText(file: work.file)
+            }
+
+            func setupObserver(for pdfWorkerWebViewHandler: PDFWorkerWebViewHandler) {
+                pdfWorkerWebViewHandler.observable
+                    .subscribe(onNext: { [weak self] in
+                        guard let self else { return }
+                        process(result: $0)
```
michalrentka, line 128: `self` is unused, pass to `process` sub-function

Zotero/Controllers/PDFWorkerController.swift
```diff
@@ -0,0 +1,203 @@
...
+        return Maybe.create { [weak self] maybe in
+            guard let self else {
+                maybe(.completed)
+                return Disposables.create()
+            }
+            if DispatchQueue.getSpecific(key: dispatchSpecificKey) == accessQueueLabel {
+                cleanup(for: work, maybe: maybe)
+            } else {
+                accessQueue.async(flags: .barrier) {
+                    cleanup(for: work, maybe: maybe)
```
michalrentka, line 182: `[weak self]` and pass to `cleanup` sub-function (same as above on 179)

Zotero/Controllers/PDFWorkerController.swift
```diff
@@ -0,0 +1,203 @@
...
+            // Then cancel actual works, and send cancelled event for each queued work.
+            let works = subjectsByPDFWork.keys + Array(queue.keys)
+            for work in works {
+                cancel(work: work)
+            }
+        }
+    }
+
+    private func cleanupPDFWorker(for work: PDFWork) -> Maybe<PublishSubject<Update>> {
+        return Maybe.create { [weak self] maybe in
```
michalrentka, line 173: Same as in `RecognizerController`, `Maybe` is not really utilised, basic Swift block should be sufficient and simpler to use.

Zotero/Controllers/Web View Handling/PDFWorkerWebViewHandler.swift
```diff
@@ -0,0 +1,178 @@
...
+
+    func recognize(file: FileData) {
+        let filePath = file.createUrl().path
+        performCall {
+            performRecognize(for: filePath)
+        }
+
+        func performRecognize(for path: String) {
+            DDLogInfo("PDFWorkerWebViewHandler: call recognize js")
+            return webViewHandler.call(javascript: "recognize('\(path)');")
```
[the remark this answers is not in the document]
mvasilak: Testing with such file names, seems to work ok.

Zotero/Controllers/Web View Handling/PDFWorkerWebViewHandler.swift
```diff
@@ -0,0 +1,178 @@
...
+
+    func getFullText(file: FileData) {
+        let filePath = file.createUrl().path
+        performCall {
+            performGetFullText(for: filePath)
+        }
+
+        func performGetFullText(for path: String) {
+            DDLogInfo("PDFWorkerWebViewHandler: call getFullText js")
+            return webViewHandler.call(javascript: "getFullText('\(path)');")
```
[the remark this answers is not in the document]
mvasilak: Ditto.

Zotero/Controllers/Web View Handling/WebViewHandler.swift
```diff
@@ -70,6 +78,15 @@ final class WebViewHandler: NSObject {
         return createWebLoadedSingle()
     }
 
+    func loadHTMLString(_ string: String, baseURL: URL?) -> Single<()> {
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/61a67e8c1f27f36906e0e99e43e51108d644b926

Zotero/Controllers/Web View Handling/PDFWorkerWebViewHandler.swift
```diff
@@ -0,0 +1,178 @@
...
+        case .recognizerData:
+            guard let data = (body as? [String: Any])?["recognizerData"] as? [String: Any] else { return }
+            observable.on(.next(.success(.recognizerData(data: data))))
+
+        case .fullText:
+            guard let data = (body as? [String: Any])?["fullText"] as? [String: Any] else { return }
+            observable.on(.next(.success(.recognizerData(data: data))))
+
+        case .log:
+            DDLogInfo("JSLOG: \(body)")
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/61a67e8c1f27f36906e0e99e43e51108d644b926

Zotero/Controllers/Web View Handling/PDFWorkerWebViewHandler.swift
```diff
@@ -0,0 +1,178 @@
...
+        case initialized
+        case inProgress
+        case failed(Swift.Error)
+    }
+
+    let webViewHandler: WebViewHandler
+    private let disposeBag: DisposeBag
+    let observable: PublishSubject<Result<PDFWorkerData, Swift.Error>>
+
+    private var isLoading: BehaviorRelay<InitializationResult>
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/61a67e8c1f27f36906e0e99e43e51108d644b926

Zotero/Controllers/PDFWorkerController.swift
```diff
@@ -0,0 +1,203 @@
...
+
+    weak var webViewProvider: WebViewProvider?
+
+    // Accessed only via accessQueue
+    private static let maxConcurrentPDFWorkers: Int = 1
+    // Using an OrderedDictionary instead of an Array, so we can O(1) when cancelling a work that is still queued.
+    private var queue: OrderedDictionary<PDFWork, PublishSubject<Update>> = [:]
+    private var subjectsByPDFWork: [PDFWork: PublishSubject<Update>] = [:]
+    private var pdfWorkerWebViewHandlersByPDFWork: [PDFWork: PDFWorkerWebViewHandler] = [:]
+    private var statesByPDFWork: [PDFWork: PDFWorkState] = [:]
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/61a67e8c1f27f36906e0e99e43e51108d644b926

Zotero/Controllers/PDFWorkerController.swift
```diff
@@ -0,0 +1,203 @@
...
+    private let accessQueueLabel: String
+    private let accessQueue: DispatchQueue
+    private let disposeBag: DisposeBag
+
+    weak var webViewProvider: WebViewProvider?
+
+    // Accessed only via accessQueue
+    private static let maxConcurrentPDFWorkers: Int = 1
+    // Using an OrderedDictionary instead of an Array, so we can O(1) when cancelling a work that is still queued.
+    private var queue: OrderedDictionary<PDFWork, PublishSubject<Update>> = [:]
```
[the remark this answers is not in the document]
mvasilak: I considered OrderedSet<PDFWork> but its remove is not O(1). Check https://github.com/zotero/zotero-ios/pull/1071/commits/61a67e8c1f27f36906e0e99e43e51108d644b926

Zotero/Controllers/Web View Handling/PDFWorkerWebViewHandler.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/61a67e8c1f27f36906e0e99e43e51108d644b926

Zotero/Controllers/PDFWorkerController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/61a67e8c1f27f36906e0e99e43e51108d644b926

Zotero/Controllers/PDFWorkerController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/61a67e8c1f27f36906e0e99e43e51108d644b926

Zotero/Controllers/RecognizerController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/ab73b171265c79aeca339e902cdec23032efb901

Zotero/Controllers/RecognizerController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/ab73b171265c79aeca339e902cdec23032efb901

Zotero/Controllers/RecognizerController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/ab73b171265c79aeca339e902cdec23032efb901

Zotero/Controllers/RecognizerController.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/ab73b171265c79aeca339e902cdec23032efb901

Zotero/Controllers/RecognizerController.swift
```diff
@@ -0,0 +1,548 @@
...
+            lookupWebViewHandlersByTask = [:]
+            // Then cancel actual tasks, and send cancelled event for each queued task.
+            let tasks = subjectsByTask.keys + Array(queue.keys)
+            for task in tasks {
+                cancel(task: task)
+            }
+        }
+    }
+
+    private func cleanupTask(for task: Task) -> Maybe<PublishSubject<Update>> {
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/e577f79a1f614d9ceaa24f203a9601d668442aaf

Zotero/Controllers/Web View Handling/PDFWorkerWebViewHandler.swift
```diff
@@ -0,0 +1,178 @@
...
+            func loadIndex() -> Single<()> {
+                guard let indexUrl = Bundle.main.url(forResource: "worker", withExtension: "html") else {
+                    return .error(Error.cantFindFile)
+                }
+                return webViewHandler.load(fileUrl: indexUrl)
+            }
+        }
+    }
+
+    private func performCall(completion: @escaping () -> Void) {
```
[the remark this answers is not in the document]
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/fdc895b70508096a1c352b5571c6713c721d07f0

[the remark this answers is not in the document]
michalrentka: Not what I had in mind. You just switched Swift block to `Single` and then just subscribed to that `Single` immediately where in success block you progress to some other function. 

So what was
```
performCall() { [weak self] in
    performRecognize(for: filePath, self: self)
}
```

is now
```
performCall().subscribe(onSuccess: { [weak self] in
     performRecognize(for: filePath, self: self)
 }, onFailure: { [weak self] error in
     self?.observable.on(.next(.failure(error)))
 })
 .disposed(by: disposeBag)
```
which is actually worse.

What I wanted this to turn into is that you can get rid of `performCall` and you can chain `initializationState.flatMap({ filterState($0) }).flatMap({ webViewHandler.call(javascript: "recognize('\(path)');") }).subscribe().disposed(by: disposeBag)` directly. No need to subscribe to individual `Single`/`Observable` instance to call the next function in success block, when you can chain them.

Zotero/Controllers/Web View Handling/PDFWorkerWebViewHandler.swift
```diff
@@ -0,0 +1,182 @@
...
+                guard let indexUrl = Bundle.main.url(forResource: "worker", withExtension: "html") else {
+                    return .error(Error.cantFindFile)
+                }
+                return webViewHandler.load(fileUrl: indexUrl)
+            }
+        }
+    }
+
+    private func performCall() -> Single<()> {
+        switch initializationState.value {
```
michalrentka, line 84: This whole switch can be dropped. `BehaviorRelay` reports current state on subscription, so you can just do what you do in `.inProgress` branch directly.

michalrentka (approved):

Zotero/Controllers/Web View Handling/PDFWorkerWebViewHandler.swift
michalrentka: In some previous code review we decided we wouldn't call those attributes `self` to avoid confusion. So rename `self`s in functions to `controller`/`handler`/`instance` or whatever else. Same for all other related instances.

Zotero/Controllers/Web View Handling/PDFWorkerWebViewHandler.swift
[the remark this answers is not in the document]
michalrentka: You can also drop those `.subscribe(on: MainScheduler.instance).observe(on: MainScheduler.instance)`, it doesn't matter on which queue you call `self?.observable.on(...)`, it should be safe and most likely will be on main thread because it's called back by webview handler.

Zotero/Controllers/PDFWorkerController.swift
```diff
@@ -0,0 +1,183 @@
...
+            pdfWorkerWebViewHandlersByPDFWork = [:]
+            // Then cancel actual works, and send cancelled event for each queued work.
+            let works = subjectsByPDFWork.keys + queue
+            for work in works {
+                cancel(work: work)
+            }
+        }
+    }
+
+    private func cleanupPDFWorker(for work: PDFWork, completion: ((_ subject: PublishSubject<Update>?) -> Void)?) {
```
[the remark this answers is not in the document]
mvasilak: The type is optional so `@escaping` is implied.

Zotero/Controllers/Web View Handling/PDFWorkerWebViewHandler.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/ca751d7b948e08fdf1bc2d83cb92bcc83d4723b6

Zotero/Controllers/Web View Handling/PDFWorkerWebViewHandler.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1071/commits/9e171883f963d8f21f8d14e26d25ce2ed2cccc46

Zotero/Controllers/RecognizerController.swift
```diff
@@ -0,0 +1,521 @@
...
+                return lookupWebViewHandler
+
+                func setupObserver(for lookupWebViewHandler: LookupWebViewHandler) {
+                    lookupWebViewHandler.observable.subscribe(onNext: { [weak self] result in
+                        guard let self else { return }
+                        switch result {
+                        case .success(let data):
+                            switch data {
+                            case .identifiers(let identifiers):
+                                if identifiers.isEmpty {
```
michalrentka, line 342: What happens when `identifiers` are not empty? Don't they need to be handled?

michalrentka (approved):
I think this is good to submit now. Although ideally I'd like to see some changes in `RecognizerController`, but they would require some further refactoring which might be unnecessary time kill at this point.

Anyway if I understand `RecognizerController` correctly, the order of events there is one big chain, which could be summed up as: `enqueue pdf worker` -> `start remote recognition` -> `lookup identifiers` -> `process identified items` (create new items).

So theoretically it could be simply written as something like:
```
pdfWorkerController.queue(work: PDFWorkerController.PDFWork(file: task.file, kind: .recognizer))
.filter({
    // filter out in-progress states
})
.flatMap({
    // Report errors if failed/cancelled, continue with extracted data
})
.flatMap({
    return apiClient.send(request: RecognizerRequest(parameters: data))
})
.flatMap({
    // extract identifiers
})
.flatMap({
    // This would require a new controller which would loop through all identifiers and report when they are all finished
    // It could be `Observable` and report multiple `onNext` events so that each event is handled separately if needed, but I guess it would be even better to gather all data first and then have one database write with multiple items
    identifierController.lookupIdentifiers()
})
.subscribe(
    // Process lookup results - create/report items
    // Report errors
)
.disposed(by: disposeBag)
```

It would be neater and much easier to understand quickly what's going on. Of course there would probably have to be some more `filter`s/`flatMap`s for special cases, but I think it would still be even less verbose than it is currently, where you have to subscribe to and process each individual step.

Anyway this is not required at this point, just a thought for the future.

Zotero/Controllers/RecognizerController.swift
mvasilak: We are going to get the next update from the `lookupWebViewHandler.observable`, in the same subscription, either for `.item` or `.failure`.

PR #1072, opened 2025-02-17
Improve attachment move to/from parent actions

mvasilak:
Modifies these item actions to be in sync with the desktop implementations:
* Move item to an existing parent, by drag and drop.
* Move item to manually created parent. This action can be canceled. 
* Remove item atttachment from its parent, back to a standalone item.

### DIFF
--- a/Zotero/Scenes/Detail/DetailCoordinator.swift
+++ b/Zotero/Scenes/Detail/DetailCoordinator.swift
@@ -414,16 +414,16 @@ extension DetailCoordinator: DetailItemsCoordinatorDelegate {
 
         controller.addAction(UIAlertAction(title: L10n.Items.new, style: .default, handler: { [weak self, weak viewModel] _ in
             guard let self, let viewModel else { return }
-            let collectionKey: String?
+            let collectionsSource: ItemDetailState.DetailType.CollectionsSource?
             switch viewModel.state.collection.identifier {
             case .collection(let key):
-                collectionKey = key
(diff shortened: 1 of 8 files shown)


Zotero/Scenes/Detail/DetailCoordinator.swift
```diff
@@ -423,7 +423,7 @@ extension DetailCoordinator: DetailItemsCoordinatorDelegate {
                 collectionKey = nil
             }
             showTypePicker(selected: "") { [weak self] type in
-                self?.showItemDetail(for: .creation(type: type, child: nil, collectionKey: collectionKey), libraryId: viewModel.state.library.identifier, scrolledToKey: nil, animated: true)
+                self?.showItemDetail(for: .creation(type: type, child: nil, collectionsSource: nil), libraryId: viewModel.state.library.identifier, scrolledToKey: nil, animated: true)
```
michalrentka, line 426: Why is the "collection source" changed to `nil` from previously used `collectionKey`? Also if `collectionKey` is not needed, you can remove code above this.

michalrentka (approved):

Zotero/Scenes/Detail/DetailCoordinator.swift
mvasilak: It's an omission, good catch. Check https://github.com/zotero/zotero-ios/pull/1072/commits/b4c09a9d9952a70ad861a00e6b08a515b737bad4

michalrentka (approved):

PR #1091, opened 2025-03-17
Enable NSAllowsLocalNetworking to allow access local WebDAV Server

mr-liusg:
Enable NSAllowsLocalNetworking to allow access local WebDAV Server

### DIFF
--- a/Zotero/Info.plist
+++ b/Zotero/Info.plist
@@ -57,6 +57,10 @@
 				<true/>
 			</dict>
 		</dict>
+		<dict>
+		<key>NSAllowsLocalNetworking</key>
+		<true/>
+		</dict>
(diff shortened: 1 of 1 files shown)


michalrentka:
@mr-liusg according to documentation (https://developer.apple.com/documentation/bundleresources/information-property-list/nsapptransportsecurity/nsallowslocalnetworking) the `NSAllowsLocalNetworking` is allowed by default on iOS 10+. Did you test whether your change fixes your issue? If not, can you describe your use case so that we can try testing and possibly fixing it?

mr-liusg:
@michalrentka You are right. It seems that there is a misunderstanding of LocalNetwork. I will reopen a new PR

PR #1092, opened 2025-03-19
Allow Insecure HTTP for Reserved IP addresses (not covered by NSAllowsLocalNetworking)

mr-liusg:
Allow Insecure HTTP for Reserved IP addresses (not covered by NSAllowsLocalNetworking)
When using private networking software, this option is required to allow intranet HTTP traffic.

### DIFF
--- a/Zotero/Info.plist
+++ b/Zotero/Info.plist
@@ -56,6 +56,11 @@
 				<key>NSIncludesSubdomains</key>
 				<true/>
 			</dict>
+			<key>100.64.0.0/10</key>
+			<dict>
+				<key>NSExceptionAllowsInsecureHTTPLoads</key>
+				<true/>
+			</dict>
 		</dict>
 	</dict>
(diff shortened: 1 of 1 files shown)


michalrentka:
@mr-liusg did you test whether your change fixes the issue? How did you come to this solution?

mr-liusg:
> @mr-liusg did you test whether your change fixes the issue? How did you come to this solution?

Yes. I've resolved this issue in my test environment.

The problem I was having was that I was using Tailscale for home network networking and when I expected to use my home NAS, it had a WebDAV address similar to https://100.65.11.105:5244/dav , limited by Apple's ATS.
After adding this configuration item, everything worked fine.
Tailscale assigns IPs in the reserved address 100.64.0.0/10 by default, and this CIDR will not be used on the internet either.
I investigated the ATS configurations of other similar apps, and many of them just disable the ATS outright or don't use URLSession.
I think it would be better to add this specific CIDR for security reasons.

PR #1096, opened 2025-04-03
Update citation process

mvasilak:
Adds and uses zotero/utilities as submodule, instead of maintaining the files in the codebase. 
Updates zotero/utilities to the latest version and update citeproc setup accordingly.
Updates citation controller accordingly (just the necessary changes, ongoing refactoring update will be rebased and make further improvements).

Fixes issue mentioned in https://forums.zotero.org/discussion/123070/ios-citation-function-error-and-workaround


Zotero/Controllers/Citation/citation/index.js
```diff
@@ -1,4 +1,14 @@
+function log(message) {
+  webkit.messageHandlers.logHandler.postMessage(message);
+}
+let collator = new Intl.Collator(['en-US'], {
+    numeric: true,
+    sensitivity: 'base'
+});
 var Zotero = {};
+Zotero.debug = (s) => log(s);
+Zotero.locale = "en-US";
```
michalrentka, line 10: Does this do something? Maybe we could pass device locale?

michalrentka (approved):

Zotero/Controllers/Citation/citation/index.js
mvasilak: Bot `Zotero.debug` & `Zotero.locale` need to be defined, for `date.js` to work as expected.
The former is called especially on the cases where date field is not standard, such as the "year-year" range, where the date validation fails. I'll check if we can use the device locale instead.

mvasilak: Check https://github.com/zotero/zotero-ios/pull/1096/commits/66f658c402d6181ebe0f627531e62d76fdf8f65d

michalrentka (approved):

PR #1110, opened 2025-04-25
Added background audio support

### LINKED ISSUE #1109: Add background audio support
https://forums.zotero.org/discussion/123825/audio-on-ios-and-ipados-can-t-play-in-background

michalrentka:
Closes #1109

### DIFF
--- a/Zotero/Scenes/Detail/DetailCoordinator.swift
+++ b/Zotero/Scenes/Detail/DetailCoordinator.swift
@@ -6,6 +6,7 @@
 //  Copyright © 2020 Corporation for Digital Scholarship. All rights reserved.
 //
 
+import AVFoundation
 import AVKit
 import MobileCoreServices
 import UIKit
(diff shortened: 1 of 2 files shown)


Zotero/Scenes/Detail/DetailCoordinator.swift
```diff
@@ -1059,3 +1077,18 @@ extension DetailCoordinator: DetailCitationCoordinatorDelegate {
 }
 
 extension DetailCoordinator: DetailCopyBibliographyCoordinatorDelegate { }
+
+// swiftlint:disable private_over_fileprivate
```
mvasilak, line 1081: If we consider disabling this globally, you can remove it from `.swiftlint.yml`

mvasilak (approved):

Zotero/Scenes/Detail/DetailCoordinator.swift
michalrentka: I still agree with the rule. Generally `fileprivate` shouldn't really be used, pretty much only in this quite specific case, where you can just disable it. So I'd keep it.

PR #1112, opened 2025-04-30
Use prebuilt npm

mvasilak:
Using cached builds for `reader` and `pdf-worker` so far. 
`note-editor` hasn't cause any issues when buildging with Xcode cloud, but will also change this when iOS builds become available in our cache.


michalrentka (approved):

dstillman:
OK, new folders are `document-worker`, `note-editor`, and `reader`.

Issue #1116, opened 2025-05-12
Retry attachment download tasks that have failed with a Retry-After header


mvasilak:
From this forum [post](https://forums.zotero.org/discussion/124137/ios-debug-log-d1883589021/)

Moving to URLSession for attachment downloads doesn't retry anymore download tasks when needed, e.g. for response status code `429` (Too Many Requests), with a `Retry-After` header. Add retrial of active attachment downloads where needed.

PR #1117, opened 2025-05-12
Handle status response 429 as an error for download tasks

mvasilak:
Fixes handling of attachment download response when status code is `429` (Too Many Requests).

From forum [post](https://forums.zotero.org/discussion/124137/ios-debug-log-d1883589021/)

Check also #1116


dstillman:
What's the idea here? Generally speaking, retrying automatically with backoff, which we appear to be doing, and following Retry-After if present, is correct. Are you saying you want to show a 429 as an immediate error to the user? Why? I could maybe see doing that because in the iOS app this is by definition a manual request, but we should be clear about the reasoning.

mvasilak:
@dstillman since we moved the attachment downloads directly to URLSession, for background purposes, we unfortunately didn't replicate previous logic exactly, as URLSession download tasks don't do retrial automatically. Therefore we have to implement this anew.

dstillman:
Ah, sorry, I misread the debug output — the ones backing off are for a different URL.

michalrentka (approved):
Good so far :)

mvasilak:
Closes #https://github.com/zotero/zotero-ios/issues/1116 as well

michalrentka (approved):

PR #1127, opened 2025-06-10
Improve opening URLs

mvasilak:
* Fixes several edge cases when opening a URL for item selection or presentation in the reader
* Adds support for opening URLS for HTML snapshot & ePub items

- [x] Fix collapsed split view controller edge cases (iPhone & iPad in compact horizontal size class)

(Also streamlines some HTML/ePub cases that will be used to rebase multiple open items)

### DIFF
--- a/Zotero/Scenes/AppCoordinator.swift
+++ b/Zotero/Scenes/AppCoordinator.swift
@@ -17,7 +17,8 @@ import RxSwift
 protocol AppDelegateCoordinatorDelegate: AnyObject {
     func showMainScreen(isLoggedIn: Bool, options: UIScene.ConnectionOptions, session: UISceneSession, animated: Bool)
     func didRotate(to size: CGSize)
-    func show(customUrl: CustomURLController.Kind, animated: Bool)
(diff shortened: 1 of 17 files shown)


Zotero/Scenes/AppCoordinator.swift
```diff
@@ -250,18 +241,19 @@ extension AppCoordinator: AppDelegateCoordinatorDelegate {
                         }
                         return presentation
 
-                        func loadPDFPresentation(key: String, libraryId: LibraryIdentifier, coordinator: DbCoordinator) throws -> ItemPresentation? {
+                        func loadItemPresentation(kind: OpenItem.Kind, coordinator: DbCoordinator) throws -> ItemPresentation? {
```
michalrentka, line 244: This seems to do mostly the same work that `CustomURLController` does in `func openItem(path: String, queryItems: [URLQueryItem]) -> Kind?`. Could it be unified?

michalrentka (approved):

michalrentka (approved):

PR #1131, opened 2025-06-18
Speech for PDF documents

michalrentka:


michalrentka:
Thanks for that @yexingsha, looks good! I just wonder whether it's worth "hiding" the speech interface behind another button. Wouldn't it be simpler to just immediately show the speech interface? If you don't want speech it's easy to ignore and if you want it you can just tap twice instead of three times to start it.

Zotero/Models/FeatureGates.swift
```diff
@@ -42,6 +44,10 @@ struct FeatureGates: OptionSet {
 #if FEATURE_GATE_DOWNLOAD_FILES_AT_SYNC
         gates.insert(.downloadFilesAtSync)
 #endif
+
+#if FEATURE_GATE_SPEECH
+        gates.insert(.speech)
```
mvasilak, line 49: Speech requires PDF worker as well.
```suggestion
        gates.insert(.speech)
        gates.insert(.pdfWorker)
```

Zotero/Models/Defaults.swift
```diff
@@ -6,6 +6,7 @@
 //  Copyright © 2019 Corporation for Digital Scholarship. All rights reserved.
 //
 
+import AVFAudio
```
mvasilak, line 9: Doesn't seem to be needed here.
```suggestion
import AVFAudio
```

Zotero/Controllers/SpeechManager.swift
```diff
@@ -0,0 +1,402 @@
...
+    private struct PageData {
+        // Voice used for this page
+        let voice: AVSpeechSynthesisVoice
+        // Text to read
+        let text: String
+
+        init(text: String, voice: AVSpeechSynthesisVoice) {
+            self.voice = voice
+            self.text = text
+        }
```
mvasilak, line 67: Can be simplified as such:
```suggestion
        // Text to read
        let text: String
        // Voice used for this page
        let voice: AVSpeechSynthesisVoice
```

Zotero/Scenes/General/Views/AccessibilityViewHandler.swift
```diff
@@ -0,0 +1,222 @@
...
+        self.libraryId = libraryId
+        self.viewController = viewController
+        self.documentContainer = documentContainer
+        self.dbStorage = dbStorage
+        let language = try? dbStorage.perform(request: ReadSpeechLanguageDbRequest(key: key, libraryId: libraryId), on: .main)
+        speechManager = SpeechManager(delegate: delegate, speechRateModifier: Defaults.shared.speechRateModifier, voiceLanguage: language)
+    }
+
+    func createAccessibilityButton(isSelected: Bool, isFilled: Bool, isEnabled: Bool = true) -> UIBarButtonItem {
+        let button = CheckboxButton(
```
mvasilak, line 65: Checkbox button doesn't seem the right choice here. Since a modal is always shown, a simple button suffices.

Zotero/Models/Defaults.swift
michalrentka: 9c74e684f65d01c615c7e788de8685ba1d9edd3c

yexingsha:
Let's give it a try then. I'm just worried that switching between popover on the top left and controls on the bottom right is going to feel jarring, but maybe it's not that big of an issue.

A side note: since the play/pause symbols are a bit smaller than the rest, it needs to use a bigger font size. I'm using `Title1`, so that it stands out a bit more as the primary icon.

michalrentka:
> @yexingsha so I tried putting it bottom right and I matched the size of the page preview overlay. I think it looks fine when there's enough space, but when the page overlay covers whole width, our overlay needs to move and it looks a bit weird. Any ideas where to move it or what to do with it?

@yexingsha any ideas? I honestly don't know which one is better, all positions have some downsides.

yexingsha:
I've been focusing on the iOS 26 redesign, so I'm wondering if we will push this before or after that? Since following the iPadOS 26 design would put this at the center of the top toolbar:

reader

reader-sidebar

If we want to put it closer to the screen edge (which iPad native apps seem to have totally abandoned), we can put it right next to the reading mode button.

The one uncertain case is when the annotation toolbar is at the very top in full-screen mode. The solution to that is going to depend on whether we style the annotation toolbar as a liquid glass top toolbar. If we keep the current structure, then the read aloud controls will be below that and centered; if we style it as a liquid glass top toolbar, then I think it makes more sense to put the annotation tools at the center, the undo/redo/close/move buttons at the right, and the read aloud controls at the right.

dstillman:
(This is definitely post-26.)

mvasilak (approved):

PR #1134, opened 2025-06-21
Improve detail coordinator change

mvasilak:
Fixes edge case where hidden detail coordinator change would cause an unintended deinititalization of the presented view controller's coordinator.


Zotero/Scenes/Main/Views/MainViewController.swift
```diff
@@ -111,16 +111,26 @@ final class MainViewController: UISplitViewController {
...
             collection: collection,
             searchItemKeys: searchItemKeys,
             navigationController: navigationController,
             itemsTagFilterDelegate: tagFilterController,
             controllers: controllers
         )
-        coordinator.start(animated: false)
-        detailCoordinator = coordinator
+        newDetailCoordinator.start(animated: false)
+        if let detailCoordinator, presentedViewController != nil {
```
michalrentka, line 123: Wouldn't this allow for a detail coordinator to include children coordinators of another library/group? Could that be an issue somewhere along the line?

PR #1145, opened 2025-07-06
Improve drag and drop items

mvasilak:
* Dropping items in other apps copies their citation in HTML or plain text format.
* Allows drag of multiple items from the same library. (
* Fixes drop on tags that applies tag on dropped items.

### DIFF
--- a/Zotero/Controllers/Citation/CitationController.swift
+++ b/Zotero/Controllers/Citation/CitationController.swift
@@ -24,6 +24,7 @@ private struct StyleData {
     }
 }
 
+@preconcurrency
 class CitationController: NSObject {
     struct Session: Hashable {
         let id = UUID()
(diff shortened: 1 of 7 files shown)


Zotero/Scenes/Detail/Items/Views/BaseItemsViewController.swift
```diff
@@ -25,6 +25,7 @@ class BaseItemsViewController: UIViewController {
 
     private static let itemBatchingLimit = 150
     unowned let controllers: Controllers
+    unowned let userControllers: UserControllers
```
michalrentka, line 28: Is `BaseItemsViewController` deinitialized before `UserControllers` when user logs out? Otherwise it could lead to crashes if `UserControllers` is deinitialized first.

michalrentka (approved):

Zotero/Scenes/Detail/Items/Views/BaseItemsViewController.swift
mvasilak: You are correct, I reset this commit for the time being and force pushed.

PR #1147, opened 2025-07-13
Fix collections action handler crash

mvasilak:
Potential fix for issue reported in this forum [post](https://forums.zotero.org/discussion/125361/ios-crash-report-666393121/).

The crash with the error `swift_unknownObjectUnownedLoadStrong` when accessing`dbStorage` in the nested fuction `updateTrashCount`. Both calls of this are in `Results<RItem>` observation blocks. Converting the struct to class allows to also weakly capture `self` and also pass explicitely to nested functions after guarding that it is valid.

If we still get similar crashes though, perhaps due to internal iOS changes that cause these edge cases, we should consider refactoring `BackgroundDbProcessingActionHandler` `dbStorage` to `DbStorage?` so that it can be `weak` when implementing the protocol.

### DIFF
--- a/Zotero/Scenes/Master/Collections/ViewModels/CollectionsActionHandler.swift
+++ b/Zotero/Scenes/Master/Collections/ViewModels/CollectionsActionHandler.swift
@@ -80,8 +80,9 @@ final class CollectionsActionHandler: ViewModelActionHandler, BackgroundDbProces
             guard let self, let viewModel else { return }
             do {
                 let items = try dbStorage.perform(request: ReadAllAttachmentsFromCollectionDbRequest(collectionId: collectionId, libraryId: viewModel.state.library.identifier), on: backgroundQueue)
-                let attachments = items.compactMap({ [weak self] item -> (Attachment, String?)? in
-                    guard let self, let attachment = AttachmentCreator.attachment(for: item, fileStorage: fileStorage, urlDetector: nil) else { return nil }
+                let fileStorage = self.fileStorage
+                let attachments = items.compactMap({ item -> (Attachment, String?)? in
+                    guard let attachment = AttachmentCreator.attachment(for: item, fileStorage: fileStorage, urlDetector: nil) else { return nil }
 
                     switch attachment.type {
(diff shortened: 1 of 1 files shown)


Zotero/Scenes/Master/Collections/ViewModels/CollectionsActionHandler.swift
```diff
@@ -32,88 +32,83 @@ struct CollectionsActionHandler: ViewModelActionHandler, BackgroundDbProcessingA
...
-            guard let viewModel = viewModel else { return }
-            self._downloadAttachments(in: collectionId, viewModel: viewModel)
-        }
-    }
+        backgroundQueue.async { [weak self, weak viewModel] in
+            guard let self, let viewModel else { return }
+            do {
+                let items = try dbStorage.perform(request: ReadAllAttachmentsFromCollectionDbRequest(collectionId: collectionId, libraryId: viewModel.state.library.identifier), on: backgroundQueue)
+                let attachments = items.compactMap({ [weak self] item -> (Attachment, String?)? in
+                    guard let self, let attachment = AttachmentCreator.attachment(for: item, fileStorage: fileStorage, urlDetector: nil) else { return nil }
```
michalrentka, line 84: Does `compactMap` need `[weak self]`? It's converted to `Array` so it should be run pretty much immediately?

michalrentka (approved):

Zotero/Scenes/Master/Collections/ViewModels/CollectionsActionHandler.swift
mvasilak: The `compactMap` `transform` closure is `@escaping`, it is not called inline. In this case calling `fileStorage` causes `self` to be captured.

mvasilak: Check this small improvement though https://github.com/zotero/zotero-ios/pull/1147/commits/a10a5fed874b7e77a90f1bf464e1e4f7f8278395

PR #1149, opened 2025-07-17
WIP - Adapt UI to iOS 26

mvasilak:
* Makes edit items toolbar flexible space not hide shared background.
from:
iPadOS 26 - item selection toolbar
to:
iPadOS 26 - item selection toolbar - fixed

* Implements #1171 
* Ignores appearance in iOS 26 where needed 
* Customizes bar buttons for iOS 26 
* Adapts annotation toolbar button for iOS 26
* Makes view controllers presented as page sheets or inside `ContainerViewController` to popovers for iOS 26
* Fixes layout for popovers in iOS 26
* Changes Libraries, Collections, Items UI for iOS 26
* Removes Item contextual action titles for iOS 26
* Fixes split view controller issues for iOS 26
* Fixes focus issues for iOS 26
* Makes UI split layout adaptive wheren needed for iOS 26


michalrentka (approved):

Zotero/Scenes/Main/Views/MainViewController.swift
```diff
@@ -131,7 +131,13 @@ final class MainViewController: UISplitViewController {
...
 
     private func showItems(for collection: Collection, in libraryId: LibraryIdentifier, searchItemKeys: [String]?) {
-        let navigationController = UINavigationController()
+        // Reuse the detail navigation controller that is already on screen instead of creating a new one for every collection.
+        // In iOS 26 the split view controller hoists the secondary column's toolbar items into its shared bottom toolbar. Swapping in a brand-new navigation controller
+        // for each collection leaves the previous (now deallocated) controller's items behind in that shared toolbar, so the filter/sort buttons pile up with every selection.
+        // Reusing the same navigation controller keeps the secondary column identity stable, so its toolbar items are replaced rather than accumulated. The coordinator holds the
+        // navigation controller weakly, so once it is actually gone (e.g. popped during back navigation in collapsed layout) this is nil and a fresh one is created and shown again.
+        let existingNavigationController = detailCoordinator?.navigationController
+        let navigationController = existingNavigationController ?? UINavigationController()
```
mvasilak: @michalrentka do you have any example of this issue?
I tried to reproduce it, logging even items toolbar and its items init/deinit, and I coudldn't so far. Furthermore, it messes up iOS 18 navigation, as well as iOS 26 navigation when split view is in collapsed (either phone or pad), by skipping the detail set you have to tap twice to go to a collection.

So far, it seems to me that this and the code below should be reverted.

michalrentka: Try switching between libraries, the whole split view becomes blank and you can't do anything anymore. This fixed the issue on iOS 27, but I didn't check older iOSes yet.

mvasilak: I reproduced it, still there were cases where it still happened with the fix, especially after changing window size to become collapsed and expanded again. Also, debugging I couldn't find the cause to be the toolbar items. Check https://github.com/zotero/zotero-ios/pull/1149/commits/636ce5a1a669c506c5a53cf9aea7c2ef30e01b7a for another fix that points to the transition of the split view controller as the cause. Could be an iOS 27 issue that might be fixed in later betas.

PR #1174, opened 2025-09-12
Fix support for file editing

mvasilak:
* Fixes permission check for presenting add actions.
* Ignores attachments when adding by identifier in a library that doesn't allow file editing.
* Ignores attachments in share extension, when picked collection belongs to a library that doesn't allow file editing.
* Removes convert to standalone attachment action from URL attachments' context menu.
* Use first linked URL, if any, for item accessory for items without any file attachments.

### DIFF
--- a/Zotero/Controllers/AttachmentCreator.swift
+++ b/Zotero/Controllers/AttachmentCreator.swift
@@ -24,7 +24,6 @@ struct AttachmentCreator {
 
     static func mainAttachment(for item: RItem, fileStorage: FileStorage, urlDetector: UrlDetector?) -> Attachment? {
         if item.rawType == ItemTypes.attachment {
-            // If item is attachment, create `Attachment` and ignore linked attachments.
             if let attachment = attachment(for: item, fileStorage: fileStorage, urlDetector: urlDetector) {
                 switch attachment.type {
                 case .url:
@@ -40,9 +39,14 @@ struct AttachmentCreator {
             return nil
         }
 
-        return bestFileAttachmentIfAny(for: item) ?? firstLinkedURLAttachmentIfAny(for: item, urlDetector: urlDetector)
+        if let fileAttachment = fileAttachmentData(for: item).sorted(by: { lData, rData in
+            mainAttachmentsAreInIncreasingOrder(lData: (lData.1, lData.3, lData.4), rData: (rData.1, rData.3, rData.4))
+        }).first.flatMap({ fileAttachmentDataToAttachment($0) }) {
+            return fileAttachment
+        }
+        return firstLinkedURLAttachmentIfAny(for: item, urlDetector: urlDetector)
 
-        func bestFileAttachmentIfAny(for item: RItem) -> Attachment? {
+        func fileAttachmentData(for item: RItem) -> [(Int, String, LinkMode, Bool, Date)] {
             let itemUrl = item.fields.first(where: { $0.key == FieldKeys.Item.url })?.value
             var data: [(Int, String, LinkMode, Bool, Date)] = []
             for (idx, child) in item.children.enumerated() {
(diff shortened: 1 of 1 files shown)


ZShare/ViewModels/ExtensionViewModel.swift
```diff
@@ -1049,7 +1059,12 @@ final class ExtensionViewModel {
 
             case .itemWithAttachment(let item, let attachmentData, let attachmentFile):
                 let newTags = (Defaults.shared.shareExtensionIncludeTags ? item.tags + tags : tags)
-                attachment = .itemWithAttachment(item: item.copy(libraryId: libraryId, collectionKeys: collectionKeys, tags: newTags), attachment: attachmentData, attachmentFile: attachmentFile)
+                if state.collectionPickerState.libraryIsFilesEditable {
+                    attachment = .itemWithAttachment(item: item.copy(libraryId: libraryId, collectionKeys: collectionKeys, tags: newTags), attachment: attachmentData, attachmentFile: attachmentFile)
+                } else {
+                    // Picked library doesn't allow file editing, convert attachment to item only.
+                    attachment = .item(item.copy(libraryId: libraryId, collectionKeys: collectionKeys, tags: newTags))
```
michalrentka, line 1066: What if the user picks another collection afterwards? Won't the data be just lost?

Zotero/Controllers/AttachmentCreator.swift
```diff
@@ -22,10 +22,10 @@ struct AttachmentCreator {
 
     private static let mainAttachmentContentTypes: Set<String> = ["text/html", "application/pdf", "image/png", "image/jpeg", "image/gif", "text/plain", "application/epub+zip"]
 
-    static func mainAttachment(for item: RItem, fileStorage: FileStorage) -> Attachment? {
+    static func mainAttachment(for item: RItem, fileStorage: FileStorage, urlDetector: UrlDetector?) -> Attachment? {
         if item.rawType == ItemTypes.attachment {
             // If item is attachment, create `Attachment` and ignore linked attachments.
```
michalrentka, line 27: I guess the comment is not right anymore, since you provide the url detector the linked attachemnt is not ignored now?

Zotero/Controllers/AttachmentCreator.swift
```diff
@@ -40,20 +40,49 @@ struct AttachmentCreator {
...
 
-        var attachmentData = attachmentData(for: item)
-
-        guard !attachmentData.isEmpty else { return nil }
-
-        attachmentData.sort { lData, rData in
-            mainAttachmentsAreInIncreasingOrder(lData: (lData.1, lData.3, lData.4), rData: (rData.1, rData.3, rData.4))
+        return bestFileAttachmentIfAny(for: item) ?? firstLinkedURLAttachmentIfAny(for: item, urlDetector: urlDetector)
+
+        func bestFileAttachmentIfAny(for item: RItem) -> Attachment? {
```
michalrentka, line 45: This looks a bit messier than it was, while it's still pretty much just moved around without many changes. I'd keep the separation between filtering out items (previously in `attachmentData(for:)` and sorting based on priority, which is still in place (`mainAttachmentsAreInIncreasingOrder`). Maybe even add another one to create `Attachment` from the main attachment data.

Zotero/Scenes/Detail/DetailCoordinator.swift
```diff
@@ -441,15 +441,13 @@ extension DetailCoordinator: DetailItemsCoordinatorDelegate {
         let controller = UIAlertController(title: nil, message: nil, preferredStyle: .actionSheet)
         controller.popoverPresentationController?.sourceItem = button
 
-        if viewModel.state.library.metadataAndFilesEditable {
-            controller.addAction(UIAlertAction(title: L10n.Items.lookup, style: .default, handler: { [weak self] _ in
-                self?.showLookup(startWith: .manual(restoreLookupState: false))
-            }))
+        controller.addAction(UIAlertAction(title: L10n.Items.lookup, style: .default, handler: { [weak self] _ in
```
michalrentka, line 444: Should lookup be available if metadata editing is not enabled?

Zotero/Scenes/Detail/DetailCoordinator.swift
```diff
@@ -471,7 +469,7 @@ extension DetailCoordinator: DetailItemsCoordinatorDelegate {
             showNote(library: viewModel.state.library, kind: .standaloneCreation(collection: viewModel.state.collection), saveCallback: nil)
         }))
 
-        if viewModel.state.library.metadataAndFilesEditable {
+        if viewModel.state.library.filesEditable {
```
michalrentka, line 472: If metadata editing is disabled, you can't create an item for the new file attachment? So why go from `metadataAndFilesEditable` to just `filesEditable`?

ZShare/ViewModels/ExtensionViewModel.swift
mvasilak: This is done when submitting, so there is no data loss.

Zotero/Controllers/AttachmentCreator.swift
mvasilak: Removed comment, check https://github.com/zotero/zotero-ios/pull/1174/commits/5a42917d72f728fc8d1ac28317e990e1cc9125d4

michalrentka (approved):

PR #1199, opened 2025-10-29
Improve trash

mvasilak:
- Shows empty trash icon if trash is empty and showing item count is enabled.
- Removes `Convert to Standalone Attachment` & `Move to Trash` actions from attachments in item details of items in trash (parity with desktop).


Zotero/Scenes/Detail/ItemDetail/Models/ItemDetailState.swift
```diff
@@ -250,6 +250,7 @@ struct ItemDetailState: ViewModelState {
     var changes: Changes
     var isEditing: Bool
     var isSaving: Bool
+    var isTrash: Bool
```
michalrentka: I'd maybe move this to `Data` which already has `isAttachment` just to be consistent.

PR #1209, opened 2025-11-20
Fix bugs

mvasilak:
* Potential fix for `ItemsActionHandler` apply settings crash (reported in [1](https://forums.zotero.org/discussion/128095/ios-crash-report-1844923468))
* Potential fix for `CollectionsActionHandler` update trash count crash (reported in [2](https://forums.zotero.org/discussion/128120/ios-crash-report-1320653110))
* Potential fix for `ItemsViewController` observation update crash when both deletions and modifications take place (reported in [3](https://forums.zotero.org/discussion/128174/ios-crash-report-1601815099/))

### DIFF
--- a/Zotero/Scenes/Detail/Items/Views/ItemsViewController.swift
+++ b/Zotero/Scenes/Detail/Items/Views/ItemsViewController.swift
@@ -371,9 +371,10 @@ final class ItemsViewController: BaseItemsViewController {
                 updateTagFilter(filters: viewModel.state.filters, collectionId: collection.identifier, libraryId: library.identifier)
 
             case .update(let results, let deletions, let insertions, let modifications):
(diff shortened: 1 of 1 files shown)


Zotero/Scenes/Master/Collections/Models/CollectionsState.swift
```diff
@@ -46,6 +46,8 @@ struct CollectionsState: ViewModelState {
     var trashItemsCountToken: NotificationToken?
     var trashCollectionsCountToken: NotificationToken?
     var itemsChangesToken: NotificationToken?
+    var cachedTrashItemsCount: Int
```
michalrentka: Do we need these new vars in both `CollectionsState` and `CollectionsActionHandler`?

PR #1212, opened 2025-11-27
Improved html/epub apple pencil usage

michalrentka:


michalrentka:
@mvasilak I updated the reader, can you re-check whether there are improvements?

mvasilak:
@michalrentka @AbeJellinek it seems to me now that selection with a finger is a bit worse, as there are cases that I start over text and I get navigation gesture detection. Pencil interactions seem the same to me. 

Furthermore, changing setting `UIPencilInteraction.prefersPencilOnlyDrawing` value in `setPenExclusive` doesn't seem to affect the UX, I can still highlight/underline with a finger in both cases.

michalrentka:
> > `setPenExclusive` doesn't seem to affect the UX, I can still highlight/underline with a finger in both cases.
> 
> Yeah, I wasn't completely sure how to handle this because it turns out we can't detect whether the pencil is connected, and nothing in this PR is calling `setPenConnected()`, which the current implementation needs. I seem to remember we discussed treating it as connected until some timeout period after the last pencil interaction (a half an hour?). @michalrentka, is that correct? If so, I can implement.

Sorry @AbeJellinek I missed this comment. 

Yes, we can't really detect if the pencil dies / disconnects or is just out of range. I'd probably give it shorter timeout, like 5 - 10 minutes. So if a pencil touch happens while annotating, finger navigation is enabled. If the user doesn't touch the document with a pencil for ~5 minutes, annotations with fingers are enabled.

Or we can just do longer timeout and if the pencil dies, the user will have to disable and re-enable annotations to enable annotating with fingers, which might honestly be better, because if you're actually using the pencil, it'll probably be near the iPad and charged most of the time.

AbeJellinek:
OK, see zotero/reader@9619295cbcecf8732c54d74e39cbb0657ba6bb15 and let me know if that's what you were thinking.

mvasilak:
@AbeJellinek @michalrentka I tried these tests with the PencilPro, on a build with the latest reader. First in the browser (where applicable) and then in the reader, for comparison.

- Start dragging from a point with words and lift to finish selection, first with a finger, and then with the pencil. Finger selection is better and the same in both browser and reader, with selection appearing from the start. Pencil selection appears only when you lift in both cases, and in the reader a swipe action is shown starting but dismissed. It seem to work most of the times, but there are times that the swipe action concludes while you select and you end up in another page with the selection there.

https://github.com/user-attachments/assets/eb75fe11-61a2-4011-b721-b4a5fd08c709

- With the pencil, double tap to select a word, then drag to customise selection, and finally lift. In the reader it works half of the time, and the other half it registers the event to move to/from focus mode.

https://github.com/user-attachments/assets/1041ad46-c468-4547-b400-0c394dd9fc27

- When the highlight tool is active, pencil seems more precise than finger, but there are still edge cases where a swipe action might conclude, messing up results.

AbeJellinek:
Thanks, I'll take a look.

Apple Pencil should be able to switch pages (and toggle the UI chrome) in addition to selecting/annotating unless `penExclusive` is true, though, right? Just need to do a better job of differentiating the intention.

mvasilak:
@AbeJellinek I checked both cases for `penExclusive`, and to be honest I don't see any differences, unless I didn't look for the exact difference. The findings are:
- I can still select, or highlight with the highlight tool active with both pencil and finger.
- Navigation can be done with both. Pencil always seem to be more bothersome to swipe forward. Swipe backward with the pencil works most of the time, but still worse than with the finger. Tapping the page edge to navigate with the pencil works fine.
- With both pencil and finger, if an annotation popup is dismissed by tapping on an edge, dismisses the popup and also navigates, which is not ideal UX.

Let me know if you want me to check something more specific.

AbeJellinek:
OK, thanks, that's helpful. What's the expected behavior in each of those cases with `penExclusive`?

And when you say "annotation popup," do you mean the popup shown by the app? Or the reader-internal popup in a non-`view` build?

AbeJellinek:
I guess my first question should be more specific - with `penExclusive` true, should the pencil *only* be able to highlight? No switching pages? Or selecting text (and if so, what happens if the highlight/underline tool isn't selected)?

mvasilak:
@AbeJellinek 
- `penExclusive == true` means that if the pen is used for highlight/underline, then finger is not allowed for highlight/underline anymore in this session. The reason this is not so from the beginning, is that the user should still be able to annotate if they don't have their pencil, regardless of this preference. Navigation, text selection, and annotation selection should always be possible.
- By "annotation popup" I do mean the one shown by the app. So I'm not sure if this falls under the reader, or the app should let the reader know to ignore events until the popup is dismissed.

PR #1213, opened 2025-12-03
Add debug reader item action for debug builds

mvasilak:
Adds debug reader item action when opening an Epub or HTML snapshot.
Allows downloading of a specific reader build to use, by providing either its commit hash, or an `http` or `https` URL to a zip archive.
It is available only in debug builds. 
Reader builds for hash commits are cached locally.


mvasilak:
@michalrentka ready for another look.

michalrentka (approved):

PR #1215, opened 2025-12-07
Update reader settings preferred content size when needed

### LINKED ISSUE #1204: Bottom of reader settings cut off at larger Dynamic Type sizes?
https://forums.zotero.org/discussion/comment/502327/#Comment_502327

I didn't try to reproduce, but this doesn't look right.

mvasilak:
Closes #1204

### DIFF
--- a/Zotero/Scenes/General/Views/ReaderSettingsViewController.swift
+++ b/Zotero/Scenes/General/Views/ReaderSettingsViewController.swift
@@ -67,7 +67,7 @@ final class ReaderSettingsViewController: UICollectionViewController {
 
         func updatePreferredContentSizeIfNeeded() {
             let contentHeight = max(collectionView.collectionViewLayout.collectionViewContentSize.height, minimumPreferredContentSize.height)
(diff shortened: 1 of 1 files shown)


Zotero/Scenes/General/Views/ReaderSettingsViewController.swift
```diff
@@ -56,6 +61,20 @@ final class ReaderSettingsViewController: UICollectionViewController {
             .disposed(by: disposeBag)
     }
 
+    override func viewDidLayoutSubviews() {
+        super.viewDidLayoutSubviews()
+        updatePreferredContentSizeIfNeeded()
+
+        func updatePreferredContentSizeIfNeeded() {
+            let contentHeight = max(collectionView.collectionViewLayout.collectionViewContentSize.height, minimumPreferredContentSize.height)
+            let contentWidth = max(collectionView.collectionViewLayout.collectionViewContentSize.width, minimumPreferredContentSize.width)
```
michalrentka: I'd keep the width fixed, collection height should be calculated based on the fixed width and the preferred width will always be bigger anyway.

michalrentka (approved):

PR #1224, opened 2025-12-30
Create JS Context for PDF worker

mvasilak:
Creates a JS Context to load the PDF worker in, that will run solely in a background queue.


michalrentka (approved):

mvasilak:
@michalrentka ready for another quick look, updated to properly use the latest pdf-worker, with some extra tests.

michalrentka (approved):

mrtcode:
@mvasilak Now that [document-worker](https://github.com/zotero/document-worker/) is merged and has fully working JSContext [tests](https://github.com/zotero/document-worker/tree/master/test/runtime/jscontext), it would be great if you could test `getStructuredDocumentText` because it does model inference using ONNX WASM, and we need to be sure all that actually works on iOS and JSContext and how fast that works, especially with larger PDFs.

PR #1235, opened 2026-02-10
Add custom PDF annotation provider

mvasilak:
Adds custom PDF document annotation provider, that improves PDF reader performance through several aspects.
* Adds custom PDF document annotation provider that combines unsupported document annotations from file, with supported document annotations that are cached in the database, and regular synced Zotero annotations already residing in the database. Database cache for document annotations is primed each time the PDF file changes.
* Fixes filtering for document annotations if they have a non-default color (parity with desktop).
* Fixes appearance change of document annotations if they have a non-default color.
* Makes side bar view controller lazy loading.
* Improves document annotation preview deletion logic.
* Zotero database annotations are also loaded lazily per page, replacing previous iterations over all Zotero annotations during document load.
* Filtering, metadata change, appearance change, also happen in the provider, lazily only for loaded pages.
* PDF reader and sidebar PDF annotations concerns are decoupled as much as is possible.
* Improves PDF reader action handler annotations observer updates, by using explicit keypaths, to ignore changes to parent item, that would trigger an update that would iterate unnecessarilly over all annotations as modified.


Zotero/Controllers/Database/Requests/ReadAnnotationsDbRequest.swift
```diff
@@ -8,42 +8,142 @@
...
+            .process(in: database)
+        var keys: [PDFReaderAnnotationKey] = []
+        for item in items {
+            if !validate || isValidDatabaseAnnotation(item: item) {
+                keys.append(PDFReaderAnnotationKey(key: item.key, sortIndex: item.annotationSortIndex, type: .database))
+            }
+        }
+        return keys
+
+        func isValidDatabaseAnnotation(item: RItem) -> Bool {
```
michalrentka: I'm a bit confused by the whole validation thing. Did you find that we store invalid annotations? Are there any places where we don't need validation? Shouldn't we just delete invalid annotations?

Zotero/Models/Database/RDocumentAnnotationCacheInfo.swift
```diff
@@ -0,0 +1,25 @@
...
+//
+//  Created by Miltiadis Vasilakis on 26/01/26.
+//  Copyright © 2026 Corporation for Digital Scholarship. All rights reserved.
+//
+
+import Foundation
+
+import RealmSwift
+
+final class RDocumentAnnotationCacheInfo: Object, LibraryScoped {
```
michalrentka: Why did you decide to create a copy of `RDocumentAnnotation` with additional data? Why not just use `EmbeddedObject`

Something like:
```
final class RDocumentAnnotationCacheInfo: EmbeddedObject {
    @Persisted var md5: String
    @Persisted var pageCount: Int
    @Persisted var annotationCount: Int
    @Persisted var uniqueBaseColors: List<String>
    @Persisted var updatedAt: Date
}
```

It would be stored alongside original `RDocumentAnnotation`, it would be created and deleted together with the item and you can query that data directly. Also, why not just include these properties into original `RDocumentAnnotation`?. This approach seems to copy data unnecessarily.

Zotero/Scenes/Detail/PDF/ViewModels/PDFAnnotationsActionHandler.swift
```diff
@@ -0,0 +1,427 @@
...
+                    return lhs.sortIndex < rhs.sortIndex
+                }
+                if lhs.key != rhs.key {
+                    return lhs.key < rhs.key
+                }
+                return lhs.type == .database && rhs.type == .document
+            })
+            return keys
+
+            func isValid(databaseAnnotation: PDFDatabaseAnnotation) -> Bool {
```
michalrentka, line 204: Same validation seems to happen in ReadAnnotationsDbRequest, do we need to re-validate here? Can the validation process be extracted and re-used in both places?

Zotero/Scenes/Detail/PDF/Models/PDFReaderAnnotationProvider.swift
```diff
@@ -0,0 +1,788 @@
...
+    private let dbQueueSpecificKey: DispatchSpecificKey<Void>
+    private let attachmentKey: String
+    private let libraryId: LibraryIdentifier
+    private let userId: Int
+    private let username: String
+    private let documentPageCount: PageCount
+    private var metadataEditable: Bool
+    private unowned let boundingBoxConverter: AnnotationBoundingBoxConverter
+    var displayName: String {
+        pdfReaderAnnotationProviderDelegate?.displayName ?? ""
```
michalrentka, line 65: Why is `displayName` provided through `pdfReaderAnnotationProviderDelegate` and `username`, `userid` are stored locally? Can't we store displayName as well? Or use delegate for all?

Zotero/Scenes/Detail/PDF/Models/PDFReaderAnnotationProvider.swift
```diff
@@ -0,0 +1,788 @@
...
+            loadDatabaseCacheIfNeededForPage(at: pageIndex)
+
+            // Annotations are fetched from `super`, apart from cached document and database annotations, as more may have been added before it was accessed here.
+            let cachePageAnnotations = super.annotationsForPage(at: pageIndex)
+
+            // Since this is the first time the annotations are loaded, apply current filter, if not empty.
+            if !currentFilterContext.isEmpty {
+                applyCurrentFilterForPage(at: pageIndex, notify: true)
+            }
+            // TODO: Check if current apperance needs to be applied to the file page annotations.
```
michalrentka: DId you? 😊

Zotero/Controllers/Database/Requests/ReadAnnotationsDbRequest.swift
```diff
@@ -8,42 +8,142 @@
...
+                DDLogInfo("ReadAnnotationKeysDbRequest: invalid sort index (\(sortIndex)) for \(item.key)")
+                return false
+            }
+
+            return true
+        }
+    }
+}
+
+struct ReadCombinedAnnotationKeysDbRequest: DbResponseRequest {
```
michalrentka: Is this used anywhere? I can't find any mention of this. Also `ReadAnnotationKeysDbRequest` seems to only be used here, so if this is not needed also delete that.

mvasilak: Cleaned up in https://github.com/zotero/zotero-ios/pull/1235/commits/766dd916f62f0f4b76e9111117cff0a721cd2a75

Zotero/Models/Database/RDocumentAnnotationCacheInfo.swift
mvasilak: `RDocumentAnnotation` entries represent individual cached document annotations. `RDocumentAnnotationCacheInfo` represents metadata for a document's annotations cache, not a copy. Perhaps I need to rename it to `RDocumentAnnotationsCacheInfo` or some other name to make it more clear. We can even consider improving the model by moving the document annotations in a list under the cache info, rather than linking to them as standalone objects, if it doesn't affect the current queries.

Zotero/Controllers/Database/Requests/ReadAnnotationsDbRequest.swift
mvasilak: Cleaned up in https://github.com/zotero/zotero-ios/pull/1235/commits/766dd916f62f0f4b76e9111117cff0a721cd2a75

Zotero/Scenes/Detail/PDF/ViewModels/PDFAnnotationsActionHandler.swift
mvasilak: Duplicate cleaned up in https://github.com/zotero/zotero-ios/pull/1235/commits/766dd916f62f0f4b76e9111117cff0a721cd2a75

Zotero/Scenes/Detail/PDF/Models/PDFReaderAnnotationProvider.swift
mvasilak: `username` & `userId` are constants in the reader state, available at the time of initialization of the provider. `displayName` is read from `Defaults.shared.displayName`, which can be modified externally during a reader's lifetime, hence needs to be dynamic.

Zotero/Scenes/Detail/PDF/Models/PDFReaderAnnotationProvider.swift
mvasilak: I checked now, and it doesn't need to be applied 😅 Removed the todo in https://github.com/zotero/zotero-ios/pull/1235/commits/01dec20647628cd4cec65f11483585860318f6b2

Zotero/Controllers/Database/Requests/ReadAnnotationsDbRequest.swift
michalrentka: The validation still happens in `PDFAnnotationsActionHandler`, so this kind of still stands. Do we know we're storing invalid annotations anywhere? If so, do we maybe want to delete the invalid ones?
Or is this more like a future compatibility check (similar to schema fields check)? In that case, do we maybe want to report some error to the user?

Zotero/Models/Database/RDocumentAnnotationCacheInfo.swift
michalrentka: In that case it's ok, I misunderstood it to be some kind of further per-annotation metadata. You can rename to `RDocumentAnnotationsCacheInfo`, it's slightly confusing, but I don't have a better name for it either.

Zotero/Controllers/Database/Requests/ReadAnnotationsDbRequest.swift
mvasilak: Unless I misunderstood, this validation comes from existing logic in https://github.com/zotero/zotero-ios/blob/01043339973910a6d7232b63c37191e1b7523144/Zotero/Scenes/Detail/PDF/ViewModels/PDFReaderActionHandler.swift#L2099
However it might not be relevant anymore, so I'll check if we still need it, and if we do if we can surface the error, rather than have a simple debug log.

michalrentka: I guess it was an issue when I implemented it, it was changing during development. We can definitely keep it in, but if we do find some incompatible annotation we could report it to the user so that they report to forums.

mvasilak: These checks were initally introduced in https://github.com/zotero/zotero-ios/pull/714 and were also extended in future PRs.
They seem to be defensive, for cases where an annotation was synced with insufficient data, perhaps due to an older version, or some other client/plugin discrepancy corrupted it? With this PR it is move outside the PDF loading workflow, and only if the annotations sidebar is opened this check takes place. However, I think we can remove them completely, and if we need we can make a database migration check. What do you think?

michalrentka: Ideally it shouldn't be needed, but I guess even migration errors or sync errors can happen, so I wouldn't get rid of it, as it could appear as weird random bugs all over pdf reader (text annotation doesn't render properly, annotations missing from pages, incorrect data in sidebar, etc.). 

At the very least change from `DDLogInfo` to `DDLogError` in cases where it's actually an error (so where it returns false - invalid). If at some point in future we end up with reports of missing annotations, we'll improve to show the error to the user.

Zotero/Models/Database/RDocumentAnnotationCacheInfo.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1235/commits/f80ef0ee57c9eb5317cee66c69f33f3b271c3ec1

michalrentka (approved):
Eh I forgot to submit my comments :). Anyway it's mostly fine now, maybe change those `DDLogInfo`s to `DDLogError`s and it's ok to merge.

Zotero/Controllers/Database/Requests/ReadAnnotationsDbRequest.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1235/commits/2eb93002f8cfc3951ccebbccf45694a8839148a4

PR #1250, opened 2026-03-24
Recently read

michalrentka:
Closes #1246


mvasilak (changes requested):
* `SyncController` tests need to be updated as well.
* When a document closes, shouldn't it submit last read without delay?
* Swipe action to remove from a collection is not available in `Recently Read`, but it does make sense to be and correspond to `Remove from Recently Read`, perhaps with a different icon.
* Using the same icon for `Remove from Recently Read` & `Move to Trash` is confusing. Perhaps it shouldn't even be a destructive action?
* Whenever an item's `lastRead` is set to `nil`, shouldn't the `effectiveLastRead` of the item be updated to `nil`, and its parent be informed?

mvasilak (approved):

PR #1260, opened 2026-04-07
Changed item sorting view to expand sort options when view is full screen

### LINKED ISSUE #1129: Show Sort By options inline
Since we have the space, can we just show the Sort By options in a vertical list and not require an extra tap for the drop-down?

michalrentka:
Closes #1129


mvasilak (changes requested):
* When the sort view appears in split view, the items toolbar disappears. 
* If the app size changes with this screen open, layout is not always correct.
* In the smallest possible window size the screen doesn't present all options.

In my opinion, navigation bar can be ommited, order should be on top, and sort types could always be visible. As space is more than enough there is no need for an extra click. Look at https://github.com/zotero/zotero-ios/pull/1149/commits/acf34fdd524f3a62b7b0b00650b9e569768d7ba3 for such an approach (not merged).

However, I think it might be even simpler to use an actual menu, per the iOS 26 guidelines, if possible.

michalrentka:
@yexingsha I also converted the plus button to UIMenu so that all picker elements on that screen have the same UI. But we could actually improve this one with icons. On desktop client we already have icons for "Add Document / Manually", "Add File", "New Standalone Note" and "Add by Identifier". Could you send them to me? And can you create one for "Scan Barcode"? It could be a barcode with a + in a circle, similar to other ones.

yexingsha:
Actually, I'm not sure we want to convert the plus button to UIMenu. Sort and Filter provide persisting options that modify the current items view, whereas the plus button starts an action that takes you out of the current items view. So I think an action sheet is more appropriate for the plus button, as per Apple's [HIG](https://developer.apple.com/design/human-interface-guidelines/action-sheets#iOS-iPadOS).

As to whether to add icons to the Sort and Filter menus: we definitely don't need icons for the Sort By types, since a lot of them have no universally understood visual representation, and so icons would just be visual noise. And I don't think there are good SF symbols for ascending and descending? I could make custom icons for them (vertical arrow + AZ) if you think it's necessary. We can use `arrow.down.circle.fill` for the download filter.

michalrentka:
> Actually, I'm not sure we want to convert the plus button to UIMenu. Sort and Filter provide persisting options that modify the current items view, whereas the plus button starts an action that takes you out of the current items view. So I think an action sheet is more appropriate for the plus button, as per Apple's [HIG](https://developer.apple.com/design/human-interface-guidelines/action-sheets#iOS-iPadOS).

You're right @yexingsha, but also, the action sheet looks quite dated and I don't really see it being used in Apple's own apps anywhere. I was able to find cases, where even Apple uses UIMenu instead of the action sheet. Find My is similar case to ours, you can tap "+" button to add a device and you get a UIMenu to pick what kind of device:
IMG_0769

But also in calculator you have the button to pick what calculator type you want to change to and that's UIMenu too:
IMG_0770


I don't have strong feelings about this, but it seems like even Apple is moving away from action sheets in favour of menus, which look more modern on their systems. Other apps seem to prefer menus too, either custom or system ones. But if you feel like keeping the action sheet I can revert of course.

> As to whether to add icons to the Sort and Filter menus: we definitely don't need icons for the Sort By types, since a lot of them have no universally understood visual representation, and so icons would just be visual noise. And I don't think there are good SF symbols for ascending and descending? I could make custom icons for them (vertical arrow + AZ) if you think it's necessary.

Definitely not necessary. Having icons only for ordering and not for sort type would probably look worse, so we can probably skip icons here. Or have a generic icon for sort types, but that would be weird.

> We can use `arrow.down.circle.fill` for the download filter.

The empty icon looks better I think, here's comparison:

filled
empty

michalrentka:
@yexingsha do you want me to revert the + button to action sheet or do we add icons to the + menu?

yexingsha:
Let's revert the + button to action sheet.

mvasilak:
@michalrentka there is a regression when the app is in compact horizontal size, as the filter menu doesn't show the tags filter. I reverted it for now, but I think we may need to open a new PR with the extra changes, we cannot re-open this merged PR.

PR #1264, opened 2026-04-09
Converted item detail abstract to UITextView so that the abstract is selectable

### LINKED ISSUE #1121: Make item details abstract text selectable
https://forums.zotero.org/discussion/124056/abstract-translation-in-ios-ipad

Labels don't allow text selection. Using instead a non-editable text view subclass that also accommodates collapse/expansion would allow users to more easily do so.

michalrentka:
Closes #1121


michalrentka:
@dstillman fyi, I made abstract selectable, so that you can long-press it and copy/translate/share parts of it. Currently you can only select and copy the whole thing. The little downside is that you have to tap the "Show more" / "Show less" button directly instead of tapping the whole cell to expand / compact abstract.

Zotero/Scenes/Detail/ItemDetail/Views/ItemDetailCollectionViewHandler.swift
```diff
@@ -925,10 +933,6 @@ extension ItemDetailCollectionViewHandler: UICollectionViewDelegate {
         case .addTag:
             observer.on(.next(.openTagPicker))
 
-        case .abstract:
```
mvasilak, line 928: `.abstract` case still needs to be handled by the swich.

mvasilak (changes requested):
This chage loses the recently added tap and hold menu that only had the copy action. So quick copy of the whole abstract now needs to expand and copy the whole text. Perhaps a copy button e.g. on the top right side of the cell might be a good addition.
Better yet, if possible, in collapsed state show label with copy action in context menu, as before, and in expand state adopt the text view added by this PR.

Additionally, show more and show less, can be selected along the text. It might be better to make them buttons, that are placed on top of where they would have been placed as text. This would also remove the small animation when tapping he expand link in the text view, which is a bit distracting.

Zotero/Scenes/Detail/ItemDetail/Views/ItemDetailCollectionViewHandler.swift
michalrentka: Fixed 501e5b34295844954d2066d49ffeab4828d530a7

michalrentka:
> This chage loses the recently added tap and hold menu that only had the copy action. So quick copy of the whole abstract now needs to expand and copy the whole text. Perhaps a copy button e.g. on the top right side of the cell might be a good addition. Better yet, if possible, in collapsed state show label with copy action in context menu, as before, and in expand state adopt the text view added by this PR.

Good idea, I re-added "Copy" context menu when it's collapsed and kept the new functionality when expanded.
 
> Additionally, show more and show less, can be selected along the text. It might be better to make them buttons, that are placed on top of where they would have been placed as text. This would also remove the small animation when tapping he expand link in the text view, which is a bit distracting.

Replacing with an actual button would be a bit troublesome, I exclude the link text from selection in delegate call `textViewDidChangeSelection`, so that you won't be able to select it.

501e5b34295844954d2066d49ffeab4828d530a7

mvasilak (approved):

PR #1265, opened 2026-04-10
Improve keyboard layout handling

mvasilak:
Modernizes all logic that needs to update layout when keyboard appears/disappears. It uses keyboard layout guide that works as expected when keyboard is floating and/or when app is windowed in iPadOS 26.

Fixes issue reported in https://forums.zotero.org/discussion/130599/ipados-scribble-not-working


Zotero/Scenes/Detail/Lookup/Views/ManualLookupViewController.swift
```diff
@@ -210,11 +204,12 @@ class ManualLookupViewController: UIViewController {
         }
 
         let isPhone = UIDevice.current.userInterfaceIdiom == .phone
-        self.phoneBottomConstraint.isActive = isPhone
         self.padBottomConstraint.isActive = !isPhone
         self.textView.heightAnchor.constraint(equalToConstant: 80).isActive = true
+        if isPhone {
```
michalrentka: Is `isPhone` ok here? What about small iPad window?

michalrentka (approved):

Zotero/Scenes/Detail/Lookup/Views/ManualLookupViewController.swift
mvasilak: It had the same behavior as split view or slide over, which is not ideal, as compact horizontal size causes the form sheet to expand to the bottom, and then the layout of the scan button is wrong. Check https://github.com/zotero/zotero-ios/pull/1265/commits/ef9697bbafccc95cfbd04ce7ea470dc8aab106e7 for a better approach for all the cases where the app horizontal size class has changed to compact, including iPadOS 26 windowing.

michalrentka (approved):

PR #1267, opened 2026-04-13
Properly cancel document search on debounce, increase debounce

### LINKED ISSUE #1266: Debounce PDF search?
https://forums.zotero.org/discussion/130843/ios-android-text-search-speed-improvement-proposal

Not sure if we have control over this. I haven't tested in a large-enough document to reproduce.

michalrentka:
@dstillman I increased to debounce to 250 ms. Also I now properly cancel document search when new search starts, it might help with performance on big documents.

Closes #1266


mvasilak (approved):

dstillman:
Sounds good.

PR #1268, opened 2026-04-13
Updated strings, excluded strings from Localizable.strings which don't need to be translated

michalrentka:
@dstillman I created new files Messages.strings/.stringsdict and moved some messages from Localizable to Messages. I re-added the strings you removed and hardcoded. Feel free to glance over it and let me know whether you want me to move some other error messages around.

Closes #1221


Zotero/Extensions/Messages.swift
mvasilak, line 1: This file should be added to `.gitignore` under the `# SwiftGen` section.

mvasilak (changes requested):

Zotero/Extensions/Messages.swift
michalrentka: Fixed in 1554268ced843e7492c376acbf416e5e78593e14

mvasilak (approved):

Zotero/Assets/en.lproj/Localizable.strings
```diff
@@ -423,93 +414,28 @@
...
 "errors.translators.bundle_loading" = "Could not update translators. Try again?";
-"errors.styles.addition" = "Could not add style “%@”";
-"errors.styles.deletion" = "Could not delete style “%@”";
-"errors.styles.loading" = "Could not load styles";
 "errors.styles_search.loading" = "Could not load styles. Try again?";
-"errors.citation.generate_citation" = "Could not generate citation";
-"errors.citation.generate_bibliography" = "Could not generate bibliography";
-"errors.citation.invalid_types" = "Invalid item types selected";
-"errors.citation.missing_style" = "No citation style selected. Go to Settings → Quick Copy and select a new style.";
-"errors.citation.open_settings" = "Open Settings";
```
dstillman, line 504: Some of these should remain translated. Anything that can happen in normal usage (e.g., because a selected style was removed) should stay translated. It's just message for unexpected internal errors that hopefully never happen that we don't need to bother translating.

Zotero/Assets/en.lproj/Localizable.strings
```diff
@@ -423,93 +414,28 @@
...
 "errors.lookup.no_identifiers_with_lookup_data" = "Zotero could not find any new identifiers in your input, or they are already being added. Please verify your input and try again.";
-"errors.pdf.merge_too_big_title" = "Unable to merge annotations";
-"errors.pdf.merge_too_big" = "The combined annotation would be too large.";
-"errors.pdf.cant_update_annotation" = "Can’t update annotation";
-"errors.pdf.cant_add_annotations" = "Can’t add annotations";
-"errors.pdf.cant_delete_annotations" = "Can’t delete annotations";
-"errors.pdf.incompatible_document" = "This document is not supported.";
-"errors.pdf.empty_document" = "This document is empty.";
-"errors.pdf.page_index_not_int" = "Incorrect format of page stored for this document";
-"errors.pdf.document_changed" = "This document has been changed on another device. Please reopen it to continue editing.";
```
dstillman, line 515: At least some of these should remain as well.

Zotero/Assets/en.lproj/Messages.strings
```diff
@@ -0,0 +1,90 @@
...
+"errors.shareext.cant_load_schema" = "An error occurred. Please open the Zotero app, sync, and try again.";
+"errors.shareext.download_failed" = "Could not download file";
+"errors.shareext.items_not_found" = "No items found on page";
+"errors.shareext.translation_failed" = "Saving failed";
+"errors.shareext.response_missing_data" = "An error occurred. Please try again.";
+"errors.shareext.failed_additional" = "You can still save this page as a webpage item.";
+"errors.shareext.file_not_pdf" = "Unable to save PDF";
+"errors.shareext.api_error" = "Error uploading item. The item was saved to your local library.";
+"errors.shareext.webdav_error" = "Error uploading attachment to WebDAV server";
+"errors.shareext.webdav_not_verified" = "WebDAV verification error";
```
dstillman: I'm confused. Lots of these seem like normal user-facing errors?

dstillman: To be clear, it's not that any error message shouldn't be translated. This is specifically for internal error messages that users should never see unless something went very wrong. We don't need to make localizers translate those (among other things because it's hard to even explain what it's talking about).

michalrentka:
@dstillman now I'm a bit confused too :). In the #1221 issue you mentioned:
> Anything that's an internal error, or that people couldn't possibly fix themselves and really just needs to exist so people have an error message to post to the forums.
> Actually, if strings might get shown to users at all, better than hard-coding them would just be to move them to a separate .strings file that we don't upload to Transifex

So I thought generic errors like `"errors.citation.generate_bibliography" = "Could not generate bibliography";`, `"errors.pdf.empty_document" = "This document is empty.";` or `"errors.pdf.incompatible_document" = "This document is not supported.";` don't need to be translated, because the user can't really do anything about them, but also they shouldn't happen normally.

I'm moving some other strings back to localized, but what do we do about generic messages like "Could not save note", "Unable to load libraries", "Could not remove item" etc.? They are there mostly because the database returned some error and shouldn't really happen. But the user will see them after the action failed to perform.

Zotero/Assets/en.lproj/Localizable.strings
michalrentka: My bad, these should be localized. I'd keep `"errors.citation.generate_bibliography"` removed, it's a generic error that you can't really do anything about?

Zotero/Assets/en.lproj/Localizable.strings
michalrentka: `"errors.pdf.incompatible_document"` is a bit misleading, it's used in html/epub reader actually (I'll rename it) when the document is not html/epub, but we do check that before we try to open the reader, so this is just an additional check to be 100% sure we're trying to open the right document. This shouldn't really happen.

`"errors.pdf.empty_document"` I guess it can be localized, but you can't really do anything about it? The document is "broken"

`"errors.pdf.document_changed"` should stay localized, sorry.

Zotero/Assets/en.lproj/Messages.strings
michalrentka: `"errors.shareext.file_not_pdf"` is actually not even used anymore, removed.
`"errors.shareext.api_error"` do we want this translated? It's a generic HTTP error handling, outside of the ones that we handle specifically (like 413 or 403). 
`"errors.shareext.webdav_error"` same as above, it's generic webdav error, the webdav can be misconfigured or some other issue with it.
 `"errors.shareext.webdav_not_verified"` adding back to localized

michalrentka:
@dstillman does it look better now?

PR #1272, opened 2026-04-16
Improve session reset workflow when api token is missing on keychain

mvasilak:
When user migrates to another device, it's not guarrantied that the encrypted keychain will be copied. In such cases a zotero user will be logged out and shown a debug log to share. This separates this edge case, so that the user want be prompted to share a debug log.

@dstillman @michalrentka maybe we should show an explicit alert as well?

Relevenat user report for instance https://forums.zotero.org/discussion/130933/ios-debug-log-d1756733406

### DIFF
--- a/Zotero/Controllers/Controllers.swift
+++ b/Zotero/Controllers/Controllers.swift
@@ -142,12 +142,18 @@ final class Controllers {
         do {
             // Try to initialize session
             try sessionController.initializeSession()
-            // Start with initialized session
-            update(with: sessionController.sessionData, isLogin: false, debugLogging: debugLogging)
-            // Start observing further session changes
-            startObservingSession()
-            completion(true)
+            finish(with: sessionController.sessionData, completed: true)
         } catch let error {
+            if let sessionError = error as? SessionController.Error,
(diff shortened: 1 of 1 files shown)


michalrentka (approved):
Yeah I've been looking at this too. I've seen multiple debug logs with the same issue, but nobody else mentioned transferring to new device. We should show proper error message.

dstillman:
I think we can just clear the state and log them out. No need for an explicit error. It's totally standard for some apps to be logged out after a device transfer.

(I'm not totally clear on what affects that. Presumably the specifics of the transfer method, whether they were using iCloud Backup / iCloud Keychain / etc.? We should make sure our credentials are transferred at least for the default case — e.g., we're not setting them as `ThisDeviceOnly`.)

mvasilak:
We use the default settings in the library that abstracts keychain, which make use of `kSecAttrAccessibleWhenUnlocked`. Per the docs:
> Items with this attribute migrate to a new device when using encrypted backups.

dstillman:
OK, yeah, so then everything else would theoretically be logged out too. No need for a special message.

PR #1273, opened 2026-04-17
Fixed broken dictionary localization

michalrentka:
Fixes crash in https://forums.zotero.org/discussion/130972/ios-crash-report-1093553924


michalrentka:
I see claude changed the order in localized dict, the only change is in "settings.storage.files_size_and_unit"

mvasilak (changes requested):
If only a few things should change, it's better if the order doesn't change. This is too noisy.

michalrentka:
> If only a few things should change, it's better if the order doesn't change. This is too noisy.

Fixed by hand, but Xcode still decided to change spacing in most files :)...

mvasilak (approved):

PR #1275, opened 2026-04-18
Fix last read date deletion crash

mvasilak:
Potential fix for issue reported in https://forums.zotero.org/discussion/131004/ios-crash-report-1969201833

### DIFF
--- a/Zotero/Controllers/Database/Requests/PerformDeletionsDbRequest.swift
+++ b/Zotero/Controllers/Database/Requests/PerformDeletionsDbRequest.swift
@@ -149,16 +149,36 @@ struct PerformLastReadDeletionsDbRequest: DbRequest {
     var needsWrite: Bool { return true }
 
     func process(in database: Realm) throws {
-        let objects = database.objects(RLastReadDate.self).filter(.keys(keys, in: libraryId))
-        for object in objects {
-            guard !object.isInvalidated else { continue }
-            if object.isChanged {
-                // If remotely deleted lastRead is changed locally, we want to keep the lastRead, so we mark that
-                // this lastRead is new and it will be reinserted by sync
-                object.markAsChanged(in: database)
-            } else {
-                object.willRemove(in: database)
-                database.delete(object)
+        switch libraryId {
+        case .custom(.myLibrary):
+            let objects = database.objects(RItem.self).filter(.keys(keys, in: libraryId))
+            for object in objects {
+                guard !object.isInvalidated else { continue }
+
+                // My Library stores last-read directly on the item, so only preserve it when the field itself
+                // has an unsynced local change. Other local item changes should not block a remote last-read clear.
+                if object.changedFields.contains(.lastRead) {
+                    continue
+                }
+
+                if object.lastRead != nil {
+                    object.lastRead = nil
+                    object.updateEffectiveLastRead()
+                }
+            }
+
+        case .group:
+            let objects = database.objects(RLastReadDate.self).filter(.keys(keys, in: libraryId))
+            for object in objects {
+                guard !object.isInvalidated else { continue }
+                if object.isChanged {
+                    // If remotely deleted lastRead is changed locally, we want to keep the lastRead, so we mark that
+                    // this lastRead is new and it will be reinserted by sync
+                    object.markAsChanged(in: database)
+                } else {
+                    object.willRemove(in: database)
+                    database.delete(object)
+                }
             }
         }
     }
--- a/Zotero/Controllers/Database/Requests/PerformDeletionsDbRequest.swift
+++ b/Zotero/Controllers/Database/Requests/PerformDeletionsDbRequest.swift
@@ -8,7 +8,6 @@
 
 import Foundation
 
-import CocoaLumberjackSwift
 import RealmSwift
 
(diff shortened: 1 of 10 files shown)


Zotero/Controllers/Database/Requests/PerformDeletionsDbRequest.swift
```diff
@@ -149,16 +149,36 @@ struct PerformLastReadDeletionsDbRequest: DbRequest {
...
-            guard !object.isInvalidated else { continue }
-            if object.isChanged {
-                // If remotely deleted lastRead is changed locally, we want to keep the lastRead, so we mark that
-                // this lastRead is new and it will be reinserted by sync
-                object.markAsChanged(in: database)
-            } else {
-                object.willRemove(in: database)
-                database.delete(object)
+        switch libraryId {
+        case .custom(.myLibrary):
```
michalrentka, line 157: `PerformLastReadDeletionsDbRequest` shouldn't be called for custom library. Doing anything with the item here is wrong. Item changes sync in different place and all item changes should be handled there. 

If `RLastReadDate` object was created for custom library we need to investigate why. That shouldn't happen.

michalrentka (changes requested):

Zotero/Controllers/Database/Requests/PerformDeletionsDbRequest.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1275/commits/8cf411b5df35a3e1fa768f6c76a1ec8daeb432e8

Zotero/Controllers/Database/Requests/PerformDeletionsDbRequest.swift
```diff
@@ -149,16 +150,22 @@ struct PerformLastReadDeletionsDbRequest: DbRequest {
...
+            DDLogWarn("PerformLastReadDeletionsDbRequest: Ignoring deletion for My Library with keys \(keys)")
```
michalrentka: It would be a bit nicer to actually throw an error here and then the caller of this request can decide how to handle the error. Currently PerformLastReadDeletionsDbRequest is only called from PerformDeletionsSyncAction which handles this case properly (which is a bit redundant to check for the library in multiple places), but in case this request is called from other place in future, this error could end up being ignored.

Zotero/Controllers/Database/Requests/PerformDeletionsDbRequest.swift
michalrentka: `This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository.` :)

Zotero/Controllers/Sync/SyncActions/PerformDeletionsSyncAction.swift
```diff
@@ -82,12 +84,21 @@ struct PerformDeletionsSyncAction: SyncAction {
                             groupedIndices[libraryId, default: []].append(key)
                         }
                         for (libraryId, keys) in groupedIndices {
-                            try dbStorage.perform(request: PerformLastReadDeletionsDbRequest(libraryId: libraryId, keys: keys), on: queue)
+                            switch libraryId {
```
michalrentka: As mentioned above, checking for the library here as well is not really needed. You could do 
```
do {
    try dbStorage.perform(request: PerformLastReadDeletionsDbRequest(libraryId: libraryId, keys: keys), on: queue)
} catch let error {
    if let error is PerformLastReadDeletionsDbRequest.Error.myLibrary {
        unexpectedMyLibraryLastReadDeletions.append(contentsOf: keys)
    } else {
        throw error
    }
}
```

michalrentka:
Added some comments how to slightly improve the code. Though the issue lies in a different place, looks like the faulty setting is `lastReadAloudPosition_u_R2NCC4YU`, not `lastRead`. So the issue lies in detecting the setting type :). But we can keep the code, it won't hurt to have proper checks here too.

But please make sure to fix the actual issue too.

michalrentka (changes requested):

Zotero/Controllers/Database/Requests/PerformDeletionsDbRequest.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1275/commits/da2d5abd73150ba83f50e4a0762efe7e4ce3da2d

Zotero/Controllers/Database/Requests/PerformDeletionsDbRequest.swift
mvasilak: Somehow the link to the commit on its own was copied, and not the commit as part of the PR 🤷‍♂️

Zotero/Controllers/Sync/SyncActions/PerformDeletionsSyncAction.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1275/commits/da2d5abd73150ba83f50e4a0762efe7e4ce3da2d

michalrentka (approved):

PR #1281, opened 2026-04-29
Improve annotation default page label computation

mvasilak:
* Adds tests for annotation default page label computation.
* Improves performance by using database request (more significant for attachments with a very large number of annotations).

### DIFF
--- a/Zotero/Scenes/Detail/PDF/Models/DefaultAnnotationPageLabel.swift
+++ b/Zotero/Scenes/Detail/PDF/Models/DefaultAnnotationPageLabel.swift
@@ -1,5 +1,5 @@
 //
-//  DefaultPageLabel.swift
+//  DefaultAnnotationPageLabel.swift
 //  Zotero
 //
 //  Created by Miltiadis Vasilakis on 29/4/26.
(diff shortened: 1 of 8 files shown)


Zotero/Scenes/Detail/PDF/Models/DefaultAnnotationPageLabel.swift
```diff
@@ -0,0 +1,60 @@
...
+        do {
+            return try dbStorage.perform(request: ReadDefaultAnnotationPageLabelDbRequest(attachmentKey: attachmentKey, libraryId: libraryId), on: queue)
+        } catch {
+            DDLogError("DefaultAnnotationPageLabel: failed to read default annotation page label - \(error)")
+            return .commonPageOffset(offset: 1)
+        }
+    }
+
+    static func from(databaseAnnotations: Results<RItem>) -> Self {
+        var uniquePageLabelsCountByPage: [Int: [String: Int]] = [:]
```
michalrentka: This logic seems to be the same/very similar to `ReadDefaultAnnotationPageLabelDbRequest`. Could the logic be re-used?

michalrentka (approved):

Zotero/Scenes/Detail/PDF/Models/DefaultAnnotationPageLabel.swift
mvasilak: This code is not used anymore, so it is removed in https://github.com/zotero/zotero-ios/pull/1281/commits/4812386ce57523ac7a552087990d97cf4c0637a6

PR #1283, opened 2026-05-03
Fix lookup with turnstile challenge

mvasilak:
Adds hidden turnstile chalenge resolution.
Based on https://github.com/zotero/zotero/pull/5890/

### DIFF
--- a/Zotero/Controllers/Web View Handling/WebViewHandler.swift
+++ b/Zotero/Controllers/Web View Handling/WebViewHandler.swift
@@ -29,28 +29,18 @@ class WebViewHandler: NSObject {
         case failed(Swift.Error)
     }
 
-    /// Per-host descriptor for a bot-protection challenge that can be cleared in a hidden web view.
-    /// Subclasses return one from `browserChallenge(for:statusCode:headers:responseText:)` to opt-in.
     struct BrowserChallenge {
         struct Cookie {
             let host: String
             let name: String
         }
 
-        /// Substring used to dedupe attempts within a translation/lookup operation.
         let match: String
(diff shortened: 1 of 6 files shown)


Zotero/Controllers/Web View Handling/WebViewHandler.swift
```diff
@@ -203,12 +272,53 @@ class WebViewHandler: NSObject {
...
+            tearDownActiveChallenges()
+            guard let webView else { return }
             webView.configuration.userContentController.removeAllScriptMessageHandlers()
             webView.removeFromSuperview()
         }
     }
 
+    // MARK: - Browser Challenges
+
+    /// Decide whether a non-success response is a bot challenge that can be cleared in a hidden web view.
```
michalrentka: Comments in this browser challenge section are a bit unrelated to actual `WebViewhandler`. They provide additional info for current issue, but if you read them in a few months, they don't really make sense here. In this case you're just returning nil, you don't "decide whether a non-success response is a bot challenge". "default returns nil so `sendRequest` behaves as before." is not a great comment either, because what is "before" when you don't have the context of this PR? Sure, you can look up the PR when checking git blame, but these comments just look confusing.

michalrentka (approved):

Zotero/Controllers/Web View Handling/WebViewHandler.swift
mvasilak: You are right, the generated comments are too verbose and in many cases unnecessary. Check https://github.com/zotero/zotero-ios/pull/1283/commits/ac143c0625e2c6c374ec09436d10f105c8d5af37

PR #1284, opened 2026-05-05
Update document worker

mvasilak:
Updates document-worker submodule. Also updates local wrapper and tests accordingly.


Zotero/Controllers/Web View Handling/DocumentWorkerJSHandler.swift
```diff
@@ -16,6 +16,11 @@ enum DocumentWorkerData {
 }
 
 final class DocumentWorkerJSHandler {
+    enum Action: String {
+        case recognize = "pdf.getRecognizerData"
```
michalrentka, line 20: Is it actually just for PDF? If yes, rename the action.

michalrentka (approved):

Zotero/Controllers/Web View Handling/DocumentWorkerJSHandler.swift
mvasilak: It is indeed. I'm preparing a PR that extends the document worker to allow structured document text extraction, where a lot of refactoring takes place. I'll add this improvement in that PR.

PR #1292, opened 2026-05-25
Extend document worker

mvasilak:
Adds support in DocumentWorkerController for structured document text extraction.
* Updates document-worker to https://github.com/zotero/document-worker/commit/be8a4b07bda8ebdb3d6a29c9db289e465d0ad204
* Updates all DocumentWorkerController relevant fixtures and tests.
* Improves file MD5 cache (since it is now necessary by `document-worker` for SDT extraction), by storing it in a hidden sidecar file along the attachment file. The cache maintains the memory part, making it two-layer. When an attachment file is modified, either the encompassing directory is removed, that also includes any hidden sidecar files, or those are removed explicitly where needed.
* Caches full text in a sidecar directory per attachment file, using a local cache version, to be incremented when integration changes. Path is `<attachment folder>/.zotero-derived/<attachment source hash>/text/<cache version>`. It contains a `manifest.json` file and for each page a `<page number>.json` file that contains the cached page text.
* Caches packed binary SDT in a sidecar directory per attachment file, using a local cache version, to be incremented when integration changes. Path is `<attachment folder>/.zotero-derived/<attachment source hash>/sdt/<cache version>/document.sdt`.
* Reimplements a native version of SDT pack in `SDTPack` class.
* Uses native ONNX Runtime 1.26.0 for SDT extraction. This may result in slightly altered SDT from current document-worker WKWebView WASM implementation that uses ORT 1.18.0 Dependency is not available through SPM, so it is added via a custom script to avoid adding cocoapods in the project. Run manually `./scripts/fetch_onnxruntime.sh ` before first compilation.

Also adds the following debugging tools, enabled during compilation by a feature gate: 
* Adds an attachment context menu action to generate its SDT.
* A recorder UI that shows completed and running works.
* Allows sharing cached work results for comparison with other implementations.
* Allows cancelling a running work.
* Allows purging cached work results.
* Allows toggling native/WASM SDT extraction


Zotero/Controllers/Web View Handling/DocumentWorkerJSHandler.swift
```diff
@@ -45,8 +36,17 @@ final class DocumentWorkerJSHandler {
     private var pending: [Int: (Result<Any, Swift.Error>) -> Void]
     private var loadError: Swift.Error?
     private var cachedWorkData: Data?
+    private let usesNativeONNXForStructuredDocumentText: Bool
+#if MAINAPP
```
michalrentka, line 40: I guess we can limit the whole document worker to main app? I don't think we'll be running it in the extension? If yes, at least limit all ONNX-related code to MAINAPP, not just the bridge. Possibly update init to have separate parameters (onnx included for main app).

Zotero/Controllers/DocumentWorkerController.swift
```diff
@@ -45,16 +86,23 @@ final class DocumentWorkerController {
 
         let id = UUID()
         let file: FileData
-        let shouldCacheData: Bool
+        let shouldCacheInput: Bool
```
michalrentka: Aren't `isOneOff` and `shouldCacheInput` mutually exclusive? Do we need both?

michalrentka (approved):
`DocumentWorkerController` - small cleanup could separate cache-related work for full text and sdt extraction. Functions like `cachedFullTextData` / `cachedStructuredDocumentTextData` or `writeFullTextCacheIfNeeded` / `writeStructuredDocumentTextCacheIfNeeded` could be replaced by `fullTextCache.cachedData` / `sdtCache.cachedData` or `fullTextCache.writeIfNeeded` / `sdtCache.writeIfNeeded`. Possibly some other small parts I missed.

Zotero/Controllers/DocumentWorkerController.swift
mvasilak: Good catch, made it an enum for more clarity. Check https://github.com/zotero/zotero-ios/pull/1292/commits/a068bdc9437e5694e62cb6788ed0c7743da0bdcc

Zotero/Controllers/Web View Handling/DocumentWorkerJSHandler.swift
mvasilak: Share extension does use document worker for the recognizer. Limited however the exposure of ONNX related code to main app only. Check https://github.com/zotero/zotero-ios/pull/1292/changes/e2fde6cc43f63633c9bea5fdd3fd21ed19873be0

mvasilak:
@michalrentka refactored `DocumentWorkController` for further clarity. Ready for another look.

michalrentka (approved):

PR #1293, opened 2026-05-28
Page Indicator and Chapter Highlight for Epub/Html reader

michalrentka:


mvasilak (commented):
* Page indicator in HTML/epub reader should be centered in the document, not the reader winow, because when the sidebar appears it doesn't shift.
* Page `0 of N` is shown for epubs with covers where the desktop client shows an empty page input instead.
* Moving through the outline or via page navigation gestures, from a page that is showing the proper page indicator, to another part that has an invalid outline path, doesn't clear the previous indicator.
* PDF outline is shown as current, only when scrolling pages or following links. Outline selection, page scrubber, or intra document navigation don't seem to change it.

michalrentka:
> * Page indicator in HTML/epub reader should be centered in the document, not the reader winow, because when the sidebar appears it doesn't shift.

Fixed 41dc62e76b4e9da69c7a6919330e6bcc6e20abbb

> * Page `0 of N` is shown for epubs with covers where the desktop client shows an empty page input instead.

I wanted to match PDF reader functionality where we don't have any page input. I'll hide the indicator if we're missing page info though. 62916ff3bf5b8750422f7687260ad888b8539872

> * Moving through the outline or via page navigation gestures, from a page that is showing the proper page indicator, to another part that has an invalid outline path, doesn't clear the previous indicator.

Fixed 62916ff3bf5b8750422f7687260ad888b8539872

> * PDF outline is shown as current, only when scrolling pages or following links. Outline selection, page scrubber, or intra document navigation don't seem to change it.

Fixed f7bc924c93efa225ffdfa7606c452945f7aaaf8f

mvasilak (approved):

PR #1297, opened 2026-06-07
Update reader

mvasilak:
* Updates HTMl/Epub reader document top position, so contets are not re-rendered when transitioning reader to and from full screen. Also, only moving annotation toolbar from or to top position, will cause a re-render.
* Fixes sepia appearance mode and improves appearance transitions and use in pop up view controllers.

### DIFF
--- a/Zotero/Models/Appearance.swift
+++ b/Zotero/Models/Appearance.swift
@@ -40,11 +40,34 @@ enum Appearance {
         case .dark:
             return "dark"
 
+        case .light, .sepia:
+            return "light"
+        }
+    }
+
+    var htmlEpubTheme: String? {
+        switch self {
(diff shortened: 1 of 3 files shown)


michalrentka (approved):
Looks good, but I'm not a fan of having big javascript junks in swift code (e. g. `colorSchemeUserScript`). Would it be possible to include those bigger js scripts in reader's `view.js` and only call direct functions? If it doesn't work for some reason or it's just too much work, feel free to merge.

mvasilak:
@michalrentka check https://github.com/zotero/zotero-ios/pull/1297/commits/cda3c64e509bf7f037183254dd6ae0593112e49a
The javascript changes have been integrated in the `reader` submodule.

PR #1306, opened 2026-07-09
Allow annotation editing in Html / Epub reader

michalrentka:
Enables editing of annotations in html/epub reader (https://github.com/zotero/reader/issues/212).

### DIFF
--- a/Zotero/Controllers/Database/Requests/CreateOrEditReaderAnnotationsDbRequest.swift
+++ b/Zotero/Controllers/Database/Requests/CreateOrEditReaderAnnotationsDbRequest.swift
@@ -1,5 +1,5 @@
 //
-//  CreateReaderAnnotationsDbRequest.swift
+//  CreateOrEditReaderAnnotationsDbRequest.swift
 //  Zotero
 //
 //  Created by Miltiadis Vasilakis on 20/9/24.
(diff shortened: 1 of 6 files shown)


Zotero/Controllers/Database/Requests/CreateOrEditReaderAnnotationsDbRequest.swift
```diff
@@ -37,19 +37,12 @@ class CreateReaderAnnotationsDbRequest<Annotation: ReaderAnnotation>: DbRequest
...
     func create(annotation: Annotation, parent: RItem, in database: Realm) {
-        let fromRestore: Bool
         let item: RItem
 
         if let _item = database.objects(RItem.self).uniqueObject(key: annotation.key, libraryId: libraryId) {
-            if !_item.deleted {
-                // If item exists and is not deleted locally, we can ignore this request
-                return
-            }
-
```
mvasilak, line 48: Is this edge case still covered otherwise, or it doesn't need to anymore?

mvasilak (approved):
The logged comments of the renamed requests should also be updated to use the new name.

Zotero/Controllers/Database/Requests/CreateOrEditReaderAnnotationsDbRequest.swift
michalrentka: I renamed the request to reflect that it also edits existing item. I checked where it's used and I didn't see a reason why it couldn't do that. It was used in places where items were being created, so it shouldn't really change anything, unless there are key conflicts. But then we'd have bigger problems anyway.

PR #1307, opened 2026-07-13
Fix tag picker

mvasilak:
Fixes issue reported in https://forums.zotero.org/discussion/132686/iphone-beta-tags-search-creates-duplicates


michalrentka (approved):
If you want to modernise a bit more, you could convert the table view data source / delegate to newer diffable datasource api.

michalrentka (approved):

PR #1314, opened 2026-08-14
Add pencil wheel

auscyber:
Changing colours is not particularly nice with an apple pencil and this copies the apple human interface guidelines so it produces a wheel to change colours in the pdf reader so that its actually nice. It's hover and squeeze, also i added a local usage thing so it can be signed as it doesnt work for me without being in the zotero organisation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -90,3 +90,7 @@ bundled/document_worker
 Dependencies/ONNXRuntime
 Zotero/Controllers/Citation/citation/utilities
 bundled/reader
+
+# Local build settings
+
+Local.xcconfig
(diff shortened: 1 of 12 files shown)


auscyber:
it also properly fits the pencilkit guidelines

auscyber:
@mvasilak

mvasilak:
@auscyber thank you for your effort, but it's better to request such a feature in the Zotero forums, so in case we decide to add it, it will be implemented in a manner that is consistent with the project. For instance, per the Apple human interface related [guidelines](https://developer.apple.com/design/human-interface-guidelines/apple-pencil-and-scribble#Squeeze), it's better if the squeeze action is discrete, rather than continuous, which for the task of tool options makes sense. That would also simplify a lot of the state workflow, as it wouldn't have some edge cases that are not handled ideally as is.

PR #1316, opened 2026-08-19
Html/Epub Support for Read Aloud

michalrentka:


Zotero/Controllers/SDTPack.swift
```diff
@@ -44,6 +44,17 @@ final class SDTPack {
     private let data: Data
     private let header: Header
     private let index: Index
+
+    /// The raw pack bytes, for handing the pack to the reader web view (`setSDTPack`).
+    var rawData: Data { data }
```
mvasilak: This just exposes `data` as they are needed by the reader. Why not just change the visibility of the existing constant?

Zotero/Controllers/SDTPack.swift
```diff
@@ -44,6 +44,17 @@ final class SDTPack {
...
+    /// Pack format version (the reader validates this against its own `SDT_PACK_VERSION`).
+    var packVersion: Int { Int(header.packVersion) }
+    /// Major component of the schema version (the reader validates this against its schema major version).
+    var schemaMajorVersion: Int { Int(header.schemaVersion.split(separator: ".").first.flatMap { Int($0) } ?? 0) }
```
mvasilak: Same here, since the `header` info is constant and just needs to be more visible, better do that, and move these computed properties, accoordingly named, to the `Header` struct itself. Btw, the external `Int()` in `schemaMajorVersion` seems redundant.

Zotero/Controllers/SDTPack.swift
```diff
@@ -44,6 +44,17 @@ final class SDTPack {
...
+
+    /// The document metadata dictionary (the `metadata` value of `materialize()`), decoded on its own without
+    /// materializing the whole content — cheap enough for e.g. reading the document language.
+    func metadataDictionary() throws -> [String: Any] { try metadataResult.get() }
```
mvasilak: Same here, just make `metadataResult` visible and use directly.

mvasilak (commented):
First comments:

It seems that a lot of the methods have parameters such as `rects: [CGRect], sdtStart: [Int]?, sdtEnd: [Int]?` to differentiate PDF and HTML/ePUB use cases. Maybe an enum parameter for these two cases could be passed instead?

When I create an annotation from read aloud, sometimes after I delete it from the read aloud annotation UI, it may still reappear. Per the agent:

> The read-aloud preview may enter the normal save path before it is marked as a preview. window._view.setReadAloudAnnotation() calls AnnotationManager.addAnnotation(), which can synchronously invoke onSaveAnnotations; however, readAloudPreviewIds.add(annotation.id) happens only after that call returns. When saving isn’t debounced, the preview is therefore forwarded to iOS and persisted before confirmation. A pending debounced save can also escape after cancellation removes the ID from the set. Could previews be created through a non-persisting annotation-manager path, with confirmation explicitly promoting them to normal annotations?

michalrentka:
> When I create an annotation from read aloud, sometimes after I delete it from the read aloud annotation UI, it may still reappear. Per the agent:
> 
> > The read-aloud preview may enter the normal save path before it is marked as a preview. window._view.setReadAloudAnnotation() calls AnnotationManager.addAnnotation(), which can synchronously invoke onSaveAnnotations; however, readAloudPreviewIds.add(annotation.id) happens only after that call returns. When saving isn’t debounced, the preview is therefore forwarded to iOS and persisted before confirmation. A pending debounced save can also escape after cancellation removes the ID from the set. Could previews be created through a non-persisting annotation-manager path, with confirmation explicitly promoting them to normal annotations?

I had to make a bit more changes than I anticipated at first. There were sync issues between javascript reader (which stored some read-aloud annotation state to postpone creation) and swift reader. Javascript internal state and some related debounces created unexpected issues. I decided to make the JS counterpart stateless to make it simpler for us and work around it in Swift code. Please check 0e2124ad5103daaec00f5319d536d3d4b98e1cbb and 3d09a420e79edbc806180006389310490c5ad0b9.

I'll make requested cleanups on Monday.

mvasilak (approved):
Found some edge cases. Other than those, it seems good.

* If I annotate with read aloud and exit the reader quickly, the annotation is not saved.
* Sometimes, if I navigate to another page, then start read aloud, and then try to annotate, I see selected the last sentence of the previous page.
* I start read aloud, open the read-aloud annotation, use an annotation tool to highlight another text directly, and then quickly delete the read-aloud annotation. If I exit the reader and re-open it, the manually created annotation hasn't persisted.

michalrentka:
> * If I annotate with read aloud and exit the reader quickly, the annotation is not saved.

Fixed in 446fe950cb2aeaae34088f5c08de573114c8c40e, those conditions didn't pass sometime.

> * Sometimes, if I navigate to another page, then start read aloud, and then try to annotate, I see selected the last sentence of the previous page.

That's by design, desktop does it too. If you start the annotation popup quickly, it selects a sentence before the currently read one, so if you start read aloud and tap the popup immediately, it selects the last sentence of previous page.

> * I start read aloud, open the read-aloud annotation, use an annotation tool to highlight another text directly, and then quickly delete the read-aloud annotation. If I exit the reader and re-open it, the manually created annotation hasn't persisted.

Fixed in 446fe950cb2aeaae34088f5c08de573114c8c40e, annotation popup and manual annotation shouldn't work side by side, so I just hide the popup if a tool is selected and everything should be confirmed.

michalrentka:
> It seems that a lot of the methods have parameters such as `rects: [CGRect], sdtStart: [Int]?, sdtEnd: [Int]?` to differentiate PDF and HTML/ePUB use cases. Maybe an enum parameter for these two cases could be passed instead?

Cleaned up in 7b17dead5d08b749bcccd5afc4c4d6ef47e7209f

Zotero/Controllers/SDTPack.swift
michalrentka: Updated in fc070a4b530b898c14062623b5e98ef30168a10f

Zotero/Controllers/SDTPack.swift
michalrentka: Updated in fc070a4b530b898c14062623b5e98ef30168a10f

Zotero/Controllers/SDTPack.swift
michalrentka: Updated in fc070a4b530b898c14062623b5e98ef30168a10f

mvasilak (approved):

PR #1318, opened 2026-08-28
Fixes sync issue where attachmentNeedsSync could be reset even if file is not uploaded

### LINKED ISSUE #1317: Sync issue: internal attachmentNeedsSync reset by sync
https://forums.zotero.org/discussion/133450/ipad-pdf-disappeared-ios-debug-log-d1219906611

michalrentka:
Fixes #1317

### DIFF
--- a/Zotero/Controllers/API/Responses/ItemResponse.swift
+++ b/Zotero/Controllers/API/Responses/ItemResponse.swift
@@ -306,6 +306,31 @@ struct ItemResponse {
         )
     }
 
+    /// `true` if this response describes an attachment which can store a file (`imported_file`, `imported_url`, `embedded_image`), but the remote copy of the file is missing.
+    /// The remote file is considered available if the response reports both `md5` and `mtime` (submitted by the client after a WebDAV upload, set by the backend for ZFS uploads),
+    /// or if it contains an `enclosure` link.
+    var isFileAttachmentWithMissingRemoteFile: Bool {
+        guard rawType == ItemTypes.attachment,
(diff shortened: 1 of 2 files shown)


mvasilak (commented):
This solution doesn't take into account the type of storage (Zotero or WebDAV).
E.g. an edge case not covered, is when the user changes storage. A synced file uploaded to the previous storage will have `isFileAttachmentWithMissingRemoteFile = false` even if it is not yet uploaded to Zotero. This defeats the purpose of `MarkAttachmentsNotUploadedDbRequest`.

Instead we could pass e.g. a specific `preserveAttachmentNeedsSync: Bool` parameter to `StoreItemsDbResponseRequest`, according to caller. We could still use `response.isFileAttachmentWithMissingRemoteFile` for logging.

Zotero/Controllers/API/Responses/ItemResponse.swift
```diff
@@ -306,6 +306,34 @@ struct ItemResponse {
...
+        else { return false }
+        switch linkMode {
+        case .importedFile, .importedUrl, .embeddedImage:
+            break
+
+        case .linkedFile, .linkedUrl:
+            // These attachments never store a file remotely, so there's nothing missing.
+            return false
+        }
+        guard !Defaults.shared.webDavEnabled else {
```
mvasilak: Shouldn't this be limited to my library only?
```suggestion
        if library.libraryId == .custom(.myLibrary), Defaults.shared.webDavEnabled {
```

mvasilak (changes requested):

Zotero/Controllers/API/Responses/ItemResponse.swift
michalrentka: Yes, you're right

mvasilak (approved):

PR #1319, opened 2026-08-28
Improve HTML/ePUB reader

mvasilak:
* Improves read aloud menu dismissal. Avoids accidental page change/full screen/annotation selection.
* Adds navigation bar button that exposes zoom actions (zoom in/zoom out/reset zoom).
* Improves read aloud button menu creation to avoid dismissing children menus when read aloud highlighting changes.
* Stores locally HTML/ePub reader scale per item.

### DIFF
--- a/Zotero/Scenes/General/Views/ReadAloudViewHandler.swift
+++ b/Zotero/Scenes/General/Views/ReadAloudViewHandler.swift
@@ -62,6 +62,8 @@ final class ReadAloudViewHandler<Delegate: SpeechManagerDelegate> {
     private weak var highlighterOverlay: ReadAloudHighlighterOverlayView?
     var isHighlighterOverlayVisible: Bool { highlighterOverlay != nil }
     weak var delegate: ReadAloudViewDelegate?
+    var menuWillPresent: (() -> Void)?
+    var menuDidDismiss: (() -> Void)?
     var isFormSheet: Bool {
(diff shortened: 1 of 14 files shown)


Zotero/Scenes/General/Views/ReadAloudViewHandler.swift
```diff
@@ -144,7 +146,9 @@ final class ReadAloudViewHandler<Delegate: SpeechManagerDelegate> {
     private func updateReadAloudButtonMenu(state: SpeechState) {
         guard let button = navbarHeadphonesButtonRef else { return }
         if (!state.isStopped || state.isOutOfCredits) && activeOverlay?.type != .bottomToolbar {
-            button.menu = createReadAloudMenu()
+            if button.menu == nil {
```
michalrentka: Doesn't this break menu updates? Like if you start running out of time, the menu wouldn't be re-created with the warning group of items. Or if some other state changes (voice name, play/pause/stop).

mvasilak: Check https://github.com/zotero/zotero-ios/pull/1319/commits/ad2181502942b930a72fd98d2a9ee2fe20c3d61e, it take into account to limit updates that would dismiss e.g. the speech rating submenu, just when the highlighted sentence changes.

michalrentka (approved):

PR #1321, opened 2026-09-04
Improve PDF reader annotations sidebar interaction

mvasilak:
Fixes some minor issues and edge cases regarding the interaction of PDF reader and annotations sidebar.

### DIFF
--- a/Zotero/Scenes/Detail/PDF/ViewModels/PDFAnnotationsActionHandler.swift
+++ b/Zotero/Scenes/Detail/PDF/ViewModels/PDFAnnotationsActionHandler.swift
@@ -42,6 +42,7 @@ final class PDFAnnotationsActionHandler: ViewModelActionHandler {
                 } else {
                     removeSelectionIfNeeded(state: &state, selectionFromDocument: selectionFromDocument)
                 }
+                updateEditingSelectionIfNeeded(state: &state)
 
                 // If sidebar editing is enabled and there are no results, disable it.
                 if state.sidebarEditingEnabled, (state.snapshotKeys ?? state.sortedKeys).isEmpty {
@@ -123,6 +124,7 @@ final class PDFAnnotationsActionHandler: ViewModelActionHandler {
                 applyFilter(to: &state)
                 state.changes = [.annotations, .filter]
                 removeSelectionIfNeeded(state: &state, selectionFromDocument: false)
+                updateEditingSelectionIfNeeded(state: &state)
             }
 
         case .setFilter(let filter):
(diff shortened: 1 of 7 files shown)


Zotero/Scenes/Detail/PDF/ViewModels/PDFAnnotationsActionHandler.swift
```diff
@@ -176,12 +163,12 @@ final class PDFAnnotationsActionHandler: ViewModelActionHandler {
 
     private func updateSortedKeys(in state: inout PDFAnnotationsState) {
         let sortedKeys = createSortedKeys(fromDatabaseAnnotations: state.databaseAnnotations, documentAnnotationKeys: state.documentAnnotationKeys)
-        if sortedKeys != state.sortedKeys {
+        if sortedKeys != state.snapshotKeys ?? state.sortedKeys {
```
michalrentka: I'd probably change this to `if sortedKeys != (state.snapshotKeys ?? state.sortedKeys) {`, just because `sortedKeys != state.snapshotKeys ?? state.sortedKeys` looks a bit misleading.

Zotero/Scenes/Detail/PDF/ViewModels/PDFAnnotationsActionHandler.swift
```diff
@@ -199,7 +186,7 @@ final class PDFAnnotationsActionHandler: ViewModelActionHandler {
                 }
                 return lhs.type == .database && rhs.type == .document
             })
-            return keys
+            return OrderedSet(keys)
```
michalrentka: Why don't you initialize `keys` as `OrderedSet` in the beginning rather than convert from `Array` to `OrdereSet` here? Both `append` and `sort` should work on `OrderedSet`, so no changes should be necessary.

michalrentka (approved):

Zotero/Scenes/Detail/PDF/ViewModels/PDFAnnotationsActionHandler.swift
mvasilak: Actually in this use case this is faster, as the ordered set hash table is created once, and doesn't have to be maintained with each append, or re-indexed after the sort. However, append is amortized O(1), so it really shouldn't make any noticeable difference in the end. Check https://github.com/zotero/zotero-ios/pull/1321/commits/90ff357b38c76fa6351bfdf9f53ff5ed5fad07e8

Zotero/Scenes/Detail/PDF/ViewModels/PDFAnnotationsActionHandler.swift
mvasilak: Check https://github.com/zotero/zotero-ios/pull/1321/commits/90ff357b38c76fa6351bfdf9f53ff5ed5fad07e8

PR #1324, opened 2026-09-09
Read Aloud Long Press Menus

michalrentka:
Updated PDF reader to have read aloud options in it's long press menus (on text or outside of text) and updated html/epub reader to support read aloud on text selection.


Zotero/Scenes/Detail/HTML:EPUB/Views/HtmlEpubDocumentViewController.swift
```diff
@@ -100,9 +101,21 @@ class HtmlEpubDocumentViewController: UIViewController {
                 self?.deselectText()
             }
 
+            var menuActions = [highlightAction, underlineAction]
+            if FeatureGates.enabled.contains(.speech) {
```
mvasilak, line 105: Rather than checking the feature gate directly, it should ask it's `parentDelegate.isReadAloudAvailable` like in the PDF document view controller case.

Zotero/Controllers/SpeechManager.swift
```diff
@@ -689,8 +715,31 @@ final class SpeechManager<Delegate: SpeechManagerDelegate>: NSObject, VoiceProce
             beginPlayback(page: page, startOffset: map(pageText(forPage: page)))
             return
         }
-        // Default play: begin near where the reader currently is. The in-view block index is read fresh at play time
-        // (async for HTML/EPUB; immediately nil for PDF, which starts at the top of the readable page).
+        // A text selection is mapped here, and not when the user picked the action, because the reader can only map a
+        // source position once it has the structured-document-text pack — which it was handed while the document loaded,
+        // just above.
+        if case .readerSelection(let source) = target {
```
mvasilak, line 721: All these cases handled with `if case` maybe are better in switch block.

Zotero/Controllers/SpeechManager.swift
```diff
@@ -689,8 +715,31 @@ final class SpeechManager<Delegate: SpeechManagerDelegate>: NSObject, VoiceProce
...
+                beginPlayback(page: page, startOffset: offset)
+            }
+            return
+        }
+        beginPlaybackAtReaderPosition(page: page)
+    }
+
+    /// Begins playback near where the reader currently is. The in-view block index is read fresh at play time (async for
+    /// HTML/EPUB; immediately nil for PDF, which starts at the top of the readable page).
+    private func beginPlaybackAtReaderPosition(page: Delegate.Index) {
```
mvasilak, line 738: Many new private methods that could just be nested with passed arguments.

Zotero/Scenes/Detail/HTML:EPUB/Views/HtmlEpubWebView.swift
```diff
@@ -34,6 +34,7 @@ class HtmlEpubWebView: WKWebView {
         super.buildMenu(with: builder)
         let newMenu = UIMenu(title: "", options: .displayInline, children: customMenuActions)
         builder.insertSibling(newMenu, afterMenu: .standardEdit)
+        builder.remove(menu: .speech)
```
mvasilak, line 37: This should be removed only when read aloud is not available.

mvasilak (approved):
Works well. It selects the closest sentence though, not start of paragraph as desktop does. That doesn't mean it should change, as this may make more sense for touch interfaces that don't present a button on hover to make visible that the start of paragraph is the point of entry.

Zotero/Scenes/Detail/HTML:EPUB/Views/HtmlEpubDocumentViewController.swift
michalrentka: f1793ee90cafdb2dd343a165c5447ac6dfd2eca0

Zotero/Controllers/SpeechManager.swift
michalrentka: Changed to one big switch fb255b0ae49596d96fc08b9f30b0ed3c01dfc4dc

Zotero/Controllers/SpeechManager.swift
michalrentka: Claude went through SpeechManager and did significant cleanups fb255b0ae49596d96fc08b9f30b0ed3c01dfc4dc

Zotero/Scenes/Detail/HTML:EPUB/Views/HtmlEpubWebView.swift
michalrentka: I wanted to remove it anyway because the feature is kind of meh, but I re-added it bd004c46148b3861ec37818109a6780e166ac163

Zotero/Scenes/Detail/HTML:EPUB/Views/HtmlEpubDocumentViewController.swift
```diff
@@ -100,9 +101,21 @@ class HtmlEpubDocumentViewController: UIViewController {
                 self?.deselectText()
             }
 
+            var menuActions = [highlightAction, underlineAction]
+            if parentDelegate?.isReadAloudAvailable == true {
```
mvasilak: This is always false now, because this runs before nested function `setupReadAloudIfNeeded` in `HtmlEpubReaderViewController`, as `setupViews` nested function accesses `HtmlEpubDocumentViewController.view`. It worked before because it used the feature gate, that I recommended to change.

Roll it back if it simplifies things, in the long run we can assume that read aloud will always be always available.

Alterntatively, the menu shouldn't be build when document view controller is loaded and hence it creates the web view, but perhaps the web view should have a delegate it asks during build, so it's not part of the initial setup, which makes sense if the menu in the future needs to be populated dynamically.

Zotero/Scenes/Detail/PDF/Views/PDFDocumentViewController.swift
```diff
@@ -862,14 +862,19 @@ extension PDFDocumentViewController: PDFViewControllerDelegate {
...
+        var children: [UIMenuElement] = [
             UIAction(title: L10n.Pdf.AnnotationToolbar.note, handler: { [weak viewModel] _ in
                 viewModel?.process(action: .createNote(pageIndex: pageView.pageIndex, origin: origin))
             }),
             UIAction(title: L10n.Pdf.AnnotationToolbar.image, handler: { [weak viewModel] _ in
                 viewModel?.process(action: .createImage(pageIndex: pageView.pageIndex, origin: origin))
             })
         ]
+        if parentDelegate?.isReadAloudAvailable == true {
+            children.append(UIAction(title: L10n.Speech.title, handler: { [weak self] _ in
```
mvasilak:
```suggestion
            children.append(UIAction(title: L10n.Speech.speak, handler: { [weak self] _ in
```

mvasilak (commented):
An issue that may pre-exist this PR, in PDF reader, selecting text to read aloud, it seems will always use the first displayed page, if more than one are displayed. Perhaps this `StartTarget` case needs an explicit page argument for which page to access.

Zotero/Scenes/Detail/HTML:EPUB/Views/HtmlEpubDocumentViewController.swift
michalrentka: My bad, I didn't notice the read aloud is missing, ae0e45eb52e053fcb29d07bb8dc2592231c03d75

michalrentka:
> An issue that may pre-exist this PR, in PDF reader, selecting text to read aloud, it seems will always use the first displayed page, if more than one are displayed. Perhaps this `StartTarget` case needs an explicit page argument for which page to access.

Fixed ea190d6ef2f7853fc3b4ee64c60e1d5745968c98

mvasilak (approved):

Issue #71, opened 2020-05-19
Allow filtering by a tag (or tags)


tnajdek:

michalrentka:
https://forums.zotero.org/discussion/101611/filter-by-custom-tags-on-ios#latest

PR #1326, opened 2026-09-14
Improve tag filter

mvasilak:
* Fixes lookup modal restoration in iPhone.
* Fixes tag filter when download toggle negates a selected tag.
* Fixes tag filer display all action visibility.
* Improves tag filter logic regarding size class.
* Fixes edge case where launching in iOS 18 the app collapsed in iPad might use a stale collapsed value.


Zotero/Scenes/Master/TagFiltering/Views/TagFilterViewController.swift
```diff
@@ -232,7 +220,7 @@ class TagFilterViewController: UIViewController {
             viewModel.process(action: .setShowAutomatic(!viewModel.state.showAutomatic))
         }
         var options: [UIAction] = [showAutomatic]
-        if traitCollection.horizontalSizeClass == .regular && UIDevice.current.userInterfaceIdiom == .pad {
+        if UIDevice.current.userInterfaceIdiom == .pad {
```
michalrentka: Do you know/remember why this was limited to iPad in "big" size? If we're allowing "small" iPad size, shouldn't it also be visible on iPhones?

Zotero/Scenes/Master/TagFiltering/Views/TagFilterViewController.swift
```diff
@@ -122,34 +129,14 @@ class TagFilterViewController: UIViewController {
...
-        if UIDevice.current.userInterfaceIdiom == .pad, previousTraitCollection?.horizontalSizeClass != traitCollection.horizontalSizeClass {
-            delegate?.tagOptionsDidChange()
-        }
-    }
-
     override func viewDidLayoutSubviews() {
         super.viewDidLayoutSubviews()
 
-        guard traitCollection.horizontalSizeClass == .regular && UIDevice.current.userInterfaceIdiom == .pad && !didAppear else { return }
+        guard context == .masterBottomSheet && !didSetInitialContentOffset else { return }
```
michalrentka: You can just move this whole thing to `viewIsAppearing` now. Should work the same and no need for `didSetInitialContentOffset`.

michalrentka (approved):

Zotero/Scenes/Master/TagFiltering/Views/TagFilterViewController.swift
mvasilak: I see that the limitation comes from #71, but I don't see any reference there why it is as such. I agree though, there is no reason to limit it. Check https://github.com/zotero/zotero-ios/pull/1326/commits/206f7de7ed3bdd68ea474f8966d765c374289d3b

Zotero/Scenes/Master/TagFiltering/Views/TagFilterViewController.swift
mvasilak: Good point, check https://github.com/zotero/zotero-ios/pull/1326/commits/5b3c3e9440a4eae3c4ef47bc704d81139af50d99

